PR #2, opened 2023-06-14
add RFC instructions to readme

khieta:
*Issue #, if available:*

*Description of changes:* Added instructions for the RFC process to the README.


john-h-kastner-aws (approved):

README.md
```diff
@@ -1,11 +1,77 @@
...
+
+It's often helpful to get feedback on your concept before diving into the design details required for an RFC. You may open an issue with a `Question` label on this repo to start a high-level discussion, with the goal of eventually formulating an RFC pull request with the specific implementation design.
+
+## What is the process?
+
+In short, to get a major feature added to Cedar, you must first get the RFC merged into the RFC repo as a markdown file. At that point the RFC is 'active' and may be implemented with the goal of eventual inclusion into Cedar.
+
+* Work on your proposal in a markdown file based on the template (<https://github.com/cedar-policy/rfcs/blob/main/0000-template.md>). Put care into the details: RFCs that do not present convincing motivation, demonstrate understanding of the impact of the design, or are disingenuous about the drawbacks or alternatives tend to be poorly-received. Copy your markdown file in `text/my-feature.md`, where my-feature is descriptive.
```
john-h-kastner-aws, line 46: In the rust rfc repo the draft RFC is in `0000-my-feature.md`

khieta: I believe it's still called `0000-template.md` in the Rust repo. But based on you comment, I came to the conclusion that we should delete the `0000` from the filename, since we're not using the filename to track the RFC number like the Rust repo is.

cdisselkoen: Should we have RFC numbers?  It seems to me that numbering the RFCs would be helpful

khieta: I see the value of numbering them for easier reference to the original PR (since I imagine we'll want the RFC number to correspond to the PR number). But they probably don't serve any value beyond that. I'm fine adding them back 🤷‍♀️

Anyone have strong opinions?

khieta: (What I mean by "adding them back" is adding the `0000-` back to the template name + adding instructions for how to number your RFC, which we didn't have before.)

khieta: Ok, I'll take that as consensus to add a number scheme. Are we ok with making the RFC number correspond to the PR number (note: this is what rust-lang/rfcs does)? This makes the original PR easier to find, but will lead to non-consecutive RFC numbers. For example, since this is PR #2, this means that our first RFC will be numbered 0003. 

Alternatively, we could maintain our own counter for RFCs, only assigning a number before the RFC PR is merged (note: this is what vuejs/rfcs does).

I think it'll be a little easier to maintain the former, but I also think the latter is prettier.

john-h-kastner-aws: pr numbers has my vote.  It should avoid needing to update the file name with the correct number immediately before merge.

khieta: done in the latest commit!

cdisselkoen (approved):

PR #5, opened 2023-06-16
Add `is` operator

khieta:
This RFC was migrated from [cedar-policy/cedar#94](https://github.com/cedar-policy/cedar/issues/94)

[Rendered](https://github.com/cedar-policy/rfcs/blob/feature/khieta/is-operator/text/0005-is-operator.md)


text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+- `Namespace::User::"alice" is User` is false
+- `Namespace::User::"alice" is Namespace::User` is true
+- `User::"alice" is Namespace::User` is false
+
+## Changes required to validator
+
+`e is et` will be typed as `True` if the validator can guarantee that expression `e` is an entity of type `et`, or `False` if the validator can guarantee that `e` is a non-entity type or an entity not of type `et`.
+Otherwise, the validator gives `e is et` type `Bool`.
+
+To improve typing precision, we can extend the current notion of "effects" to track information about entity types.
```
john-h-kastner-aws, line 115: This sounds doable. It would be a necessary enhancement to make template slots useful if we accept RFC #3.

text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+However, this requires users to manually add an entity's type as an attribute, and leads to the possibility that an entity might be created with an attribute that doesn't actually match its type.
+
+3. The behavior of `is` can also be simulated using groups.
+For each entity type (e.g., `File`), a user could create a group (e.g., `Type::"File"`) and add all entities of that type to the group.
+Then to check whether an entity has a particular type, users could use `in` (e.g., `resource in Type::"File"`).
+Like alternative (2), this requires some extra work on the part of the user (setting up the memberOf relation), and does not prevent entities from being added as members of the wrong group.
+
+# Unresolved questions
+
+1. Should we allow `is` to apply to action entities? Action entities should _always_ have type `Action`, although they may have different namespaces. So `is` could still be used to distinguish between `Namespace1::Action` and `Namespace2::Action`.
```
john-h-kastner-aws, line 174: I wouldn't. IMO `action is` probably shouldn't be written, but it isn't actively harmful. It would also be a pain to distinguish between `Action` entities and `AnyEntities` which could be actions, but `is` really should apply to `AnyEntity`.

There's precedent for different allowed conditions in the policy scope, so, if others want to restrict `is` on action I would be in favor of restricting `action` constraint in the scope only.

text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+
+2. Should users be able to indicate that an entity may be of multiple types, e.g., `principal is [User, Group]`? Can this be combined with `in` (e.g., `principal is [User, Group] in Group::"orgA"`, or even `principal is [User, Group] in [Group::"orgA", Group::"orgB"]`)?
```
john-h-kastner-aws, line 176: Users will ask for this :). (not necessarily in favor of doing it for the initial impl though)

text/0005-is-operator.md
aaronjeline: Agreeing with @john-h-kastner-aws. We already have special rules in the scope, but in the condition clauses actions are first class entities and should remain as such.

text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+
+```
+permit(principal, action == Action::"view", resource in Folder::"public");
+```
+
+Assuming that they pass in inputs where the principal is of type `User` and the resource is of type `File`, this will work as expected.
+However, it could also return "allow" for other principal types like `Folder` or resource types like `User`.
+This is a bug on the user's end: they shouldn’t make a query with the action "view" when the principal is not a `User` and the resource is not a `File`.
+They are breaking the contract they stated in their schema.
+Nevertheless, it might be confusing that such a query does not result in a "deny".
```
mwhicks1, line 42: Do we have an example of using `is` that is not defending against erroneous use of the schema? Maybe we can rework the example to do that? I'd feel more persuaded by that then using redundant information to defend users against themselves.

text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+};
+```
+
+# Detailed design
+
+## Changes required to parser/AST
+
+- Update the grammar and AST to support expressions of the form `e is et` where `et` is a `Name` (e.g., `Namespace::EntityType`).
+- Update the grammar and AST to support expressions of the form `e1 is et in e2` (but not `e1 in e2 is et`, because this is more difficult to mentally parse).
+- Update the grammar to allow expressions like `e is et` and `e1 is et in e2` in the policy scope.
```
mwhicks1, line 95: Do we need to include `is` in the policy scope, and do we need to include it as part of `in` expression? We don't allow things like `resource has public` in the policy scope, nor do we allow expressions like `resource has public in Folder::"public"`. What necessitates putting `is` there? I would argue that it doesn't make sense to do that because `is` expressions will not be used for policy indexing.

text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+>     permit(
+>       principal == User::"admin",
+>       action == Action::"viewPhoto",
+>       resource
+>     );
+>
+> In your schema, you indicate that the "viewPhoto" action should always be accompanied by a `Photo` for the resource, so the policy doesn't need to say that explicitly (and doesn't need to perform a runtime check).
+
+If we support an `is` operator, then customers may be encouraged to use a single "view" action for multiple resource types, rather than a new "viewX" actions for each resource type X.
+
```
mwhicks1, line 157: Our own motivating example violates this tenet. I think we should change the example so that we satisfy the tenet, but still show that adding `is` makes sense.

text/0005-is-operator.md
```diff
@@ -0,0 +1,176 @@
...
+# Alternatives
+
+1. For the most part, users can currently solve the problems that `is` solves by specifying in the schema what types of principals and resources an action applies to, or by using `has` checks, per the examples in the [Motivation](#motivation) above.
+However, in some cases, `is` may result in policies that are easier to read.
```
mwhicks1, line 161: Which cases? Let's see those in the motivation, if the ones there now are just as easily handled with existing mechanisms.

text/0005-is-operator.md
mwhicks1: If we don't allow `is` in the policy scope at all (same rule as `has`), then @john-h-kastner-aws 's suggestion is vacuously satisfied.

cdisselkoen: I'll take the other opinion, that we should allow `is` to apply to action entities.  Outside of the schema, Cedar doesn't require that action entities always have type `Action`, and I think `action is` might be useful for schema-less users.  (And perhaps, in the future, we might relax the restriction for `Action` types even for schema users, in a back-compatible way; I'm not suggesting that here, but we could at least leave the door open.)

text/0005-is-operator.md
cdisselkoen: > because `is` expressions will not be used for policy indexing.

Why wouldn't they be?  It seems to me like some consumer of Cedar could easily choose to index on `is` expressions.  In fact this might be a good idea from a performance perspective; when you see an authz query for a resource of type `XYZ`, you can use indexing to avoid pulling policies that use `resource is` with any type other than `XYZ`.  Seems to me that we should leave this door easily open to Cedar users.

mwhicks1: You could do policy indexing on attributes too. Should we allow attributes in the scope?

cdisselkoen: Maybe we should!  Write the RFC :)

cdisselkoen: More seriously, we should have a story for why we include certain expressions and not others that are legal in the policy scope.  And then apply that general principle here.

text/0005-is-operator.md
cdisselkoen: Re-reading, I see that actually I agree with @john-h-kastner-aws and @aaronjeline, and everyone is proposing allowing `action is` (at least outside the policy scope).

text/0005-is-operator.md
mwhicks1: The general principle revolves around indexing based on "edges" between _entities_ in the policy store that are on the RHS of either `==` or `in`, e.g., `User::"Alice"` or `"Album::"Photos"`. Since a type like `Photo` is not an entity, I'm scratching my head a little to see how it would fit in this scheme. Even if it did, what would it mean to allow `resource is Photo in Album::"Photos"`?  You'd potentially have a hyper-edge.

Since we don't have a good reason right now to include `is` in indexing, it seems simpler and more conceptually pure to leave `is` in the `when` clause, at least for now, and then revisit with experience. Two-way door.

cdisselkoen: I concede your point about indexing.

I did some more thinking about why it "feels" to me like we should allow `is` in the policy scope.  I came up with two arguments; maybe neither of them are any good :) and I acknowledge your point that we can always disallow it for now and later decide to add it back-compatibly.  But here goes.

First, I think users will want and expect that they can use `is` in the policy scope.  Features isomorphic to `is` have been requested many times, and in most (all?) of those cases, the motivating example presents `is` in the policy scope.  It seems to be the intuitive place for it, to our users.  If we don't allow it, users will certainly ask for it.

Second, an `is` restriction on `principal` or `resource` in a policy feels like a fundamental part of the policy you'd want up front.  I would argue that the policy scope "feels" in some way different than the `when` clause, in that the scope "defines which principals/resources we're even talking about" while the `when` clause performs some checks on those principals/resources.  Using `is` to restrict a policy to only apply to `User` principals or `File` resources feels like a fundamental part of the policy, feels like defining which principals/resources we're even talking about, rather than feeling like a check being performed on a principal/resource we've already selected.  To me, at least.  And I think putting those things up top makes intuitive sense, much like how the type signature of a function is put on top, before its body.

Again I recognize that neither of these arguments are very strong :) The argument that we shouldn't allow `is` in the policy scope certainly has more principled support, namely that it can be grounded in an argument about indexing, and it's the two-way-door option.  But just putting my brainstorming on paper here.

khieta: I also think `is` makes sense in the policy scope (for readability, at least). Similar to `in` expressions, `is` expressions restrict the "scope" of what the policy applies to, and set the context for any other restrictions in the policy condition. I feel like `is` is different from `has` since it directly references an entity type defined in the schema, which will restrict the types of entities that it can be related to by membership.

Re indexing: I think I need more context/definitions here to understand the problem. Do we have an example written up somewhere?

----

I appreciate the point that deciding not to allow `is` in the policy scope is something that we can easily reverse later, but I think it's best to discuss/decide now rather than kicking the can down the road. Like @cdisselkoen said, every time a user has asked for this, their example policy used `is` in the scope, so we should expect this request.

text/0005-is-operator.md
khieta: One think I was wondering while writing up this RFC: why do we have this tenet? What is "bad" about using a generic action type with multiple principal/resource types?

text/0005-is-operator.md
khieta: This second sentence was referring to "Use case 2"

text/0005-is-operator.md
khieta: Initial draft of request validation RFC: #11

text/0005-is-operator.md
D-McAdams: Answering the question about what's "bad" about generic action type...  This is a lesson learned through experience. There are two key points:

- The input context can vary when acting on different resource types. For a real-life example from AWS, see the [documentation for S3](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html). Make your screen really big so you can see the column in the table named "Condition keys". That's the equivalent of what Cedar calls "context". Notice how the context values can be different when operating on different resource types. If generic actions were used, then the context would be the union of all possible inputs for all resource types, most of them labeled as optional, and it becomes challenging to provide useful validation or documentation for the action.
- It's easy to accidentally overgrant permissions. For example, consider an action like "read" that applies to 10 different resource types. It's relatively easy to slip up and write a policy that accidentally overgrants to too many types of resources. By looking at the policy alone, it's not entirely clear what it gives access to. Even if the `is` operator existed, somebody still needs to remember to use it. In contrast, resource-specific action names are constrained by default, and self-documenting.
- And, if desired, you can still get the same effect of generic action names by decorating each action with attributes like `type=[read|write|list|view|delete]`, and writing policies like `permit(...) when {action.type == "read"}`

Looking at the major cloud providers as a point of reference, AWS uses resource-specific action names. Azure and GCP appear to as well, although structured a bit differently. For example, rather than an RPC-style action name like `getClustersInContainer`, it will be `container.clusters.get`. But, similar effect.

mwhicks1:
I move that we accept this RFC but with usage of `is` limited to conditions, and as its own expression not combined with `in`. This is a two-way door. We have consensus on this aspect, it presents no risk to scope-based policy indexing, and as we get more experience we can choose to allow `is` in policy scopes and/or combination with `in` as usage becomes more compelling.

text/0005-is-operator.md
mwhicks1: > it's best to discuss/decide now rather than kicking the can down the road.

I'm not sure I agree. Amazon culture favors quick decisionmaking for a reason. We have limited information now, so more talking doesn't necessarily help. Making a two-way door decision allows customers to start using what we add, and from that usage we can get more feedback on the parts we don't.

text/0005-is-operator.md
mwhicks1: This looks confusing to me. `foo in [X,Y]` looks like set inclusion, i.e., `foo` is either in `X` or `Y`. `foo is [X,Y]` seems like it might desugar to `foo is X` _and_ `foo is Y`. Analogy to `has`: If I wrote `foo has [X,Y]` would you think I'm saying it has both attributes, or just one of them?

So whether they might ask or not, I am seeing some downsides in terms of interpreting the syntax. Maybe better not to go there.

khieta:
Thanks for the feedback everyone! I am working on revising the initial draft -- hopefully I'll have it posted tomorrow. Here's a sneak peak of what I plan on changing:
* do not allow `is` in the policy scope
* do not allow combined syntax like `.. is .. in ..`
* remove "example 1", which can be solved more naturally by #11 
* spend some more time thinking about how `is` interacts with the best practice of "typed actions"

text/0005-is-operator.md
cdisselkoen: I disagree a little.  You say

> `foo is [X,Y]`seems like it might desugar to `foo is X` _and_ `foo is Y`

but the precedent you cite (`foo in [X,Y]`) has the other behavior (implicit _or_), which is exactly the right interpretation here as well.  What's inconsistent?

This is not a statement either way of my position on the actual proposal to allow `is [X,Y]`, which I don't have a formed opinion on yet

jeffsec-aws:
One example I was working on is ability to define a policy to a specific entity like:
```
permit (
    principal is User, // here we want to mean any entity of type User while some Admin and Partner entity types might also exist
    action == Action::"read",
    resource == Document::"ToS"
);
```
I agree that this can be implemented in the condition clause instead to prevent change of other operations like listing policies. It would have the representation:
```
permit (
    principal,
    action == Action::"read",
    resource == Document::"ToS"
) when {
   principal  is User, // here we want to mean any entity of type User while some Admin and Partner entity types might also exist
};
```
While some mitigations would also solve the question at stake, `is` operator existence is more scalable and would prevent to

*  have to make the schema larger by creating specific purpose actions ():
    ```
        "globalRead": {
            "attributes": {},
            "appliesTo": {
                "principalTypes": [ "User"],
                "resourceTypes": [ "Document" ],
            }
        }
    ```
* have to go the Group relationship way:
    ```
    permit (
        principal in Group::"All users" // This has to be maintained and pass as an entity component.... which can be large, therefore the caveat vs being able to use == User or == User::"*"
        action == Action::"read",
        resource == Document::"ToS"
    );
    ```

khieta:
Revision (finally) posted! Based on in-person discussion, I decided to leave the proposal to include `is` and `is ... in` in the scope unchanged, but I did make several revisions to the motivating example.

text/0005-is-operator.md
khieta: In the latest revision posted, `action is` is allowed anywhere outside the scope.

text/0005-is-operator.md
khieta: Latest revision includes a note saying that we will not support this syntax. (But will reconsider if/when someone asks for it.)

text/0005-is-operator.md
```diff
@@ -0,0 +1,163 @@
...
+- Update the grammar and AST to support expressions of the form `e is et` where `et` is a `Name` (e.g., `Namespace::EntityType`).
+- Update the grammar and AST to support expressions of the form `e1 is et in e2` (but not `e1 in e2 is et`, because this is more difficult to mentally parse).
+- Update the grammar to allow expressions like `e is et` and `e1 is et in e2` in the policy scope for principals and resources.
+The action part of the scope will only allow expressions like `action == Action::"foo"` and `action in [Action::"foo", Action::"bar"]`, as before.
+
+### Changes required to evaluator
+
+`e is et` will evaluate to true iff `et` is a `Name` (e.g., `Namespace::EntityType`) and `e` is an expression that is an entity of type `et`.
+Otherwise it will evaluate to false.
+To maintain the property that "conditions in the scope never error," `e is et` will never return an error; it will simply return false if any error occurs.
```
mwhicks1, line 84: What possible errors are there? The only one I can think of is that `e` evaluates to an entity reference that does not exist in the store. If that's the only error, let's call it out specifically -- it's the same semantics as for `e1 in e2` when `e1` doesn't exist, right?

mwhicks1 (approved):

text/0005-is-operator.md
cdisselkoen: Evaluating `e is et` shouldn't even require an entity store lookup

jeffsec-aws: I think @mwhicks1 insinuated a look up upon the schema definition, not the "entity store"

cdisselkoen: Evaluation never looks at the schema, and I presume no one here is proposing to change that?

cdisselkoen: Evaluating `e is et` only requires looking at the literal typename component of `e`'s UID

cdisselkoen (approved):

text/0005-is-operator.md
khieta: In the policy scope, we only allow `principal is ...` and `action is ...`, so failure will never occur. But you could still have ill-typed expression in the policy condition. Looking again at the implementation of evaluation for `in`, it looks like we do want to report type errors here rather than just evaluating to `false`.

So, proposed revised text:
```
`e is et` will evaluate to true iff `et` is a `Name` (e.g., `Namespace::EntityType`) and `e` is an expression that is an entity of type `et`.
If `e` is an entity-typed expression, but does not have type `et`, then `e is et` will evaluate to false.
If `e` is a non-entity-typed expression, the evaluator will produce a type error.

Note that this design maintain the property that "conditions in the scope never error" because `principal` and `resource` are guaranteed to be entities, so the expression `(principal|resource) is et` will never error.
```

cdisselkoen: like `2 is number`, where we'd somehow have to distinguish the builtin type `number` from the entity type `number`

mwhicks1: > Evaluating e is et can only result in an error if e is a non-entity-typed expression (e.g., "foo" is User). 

Of course! Not sure what I was thinking ...

jeffsec-aws:
The md file for this RFC has a Typo.

The text spells out

```
In many cases, this can also be done by adding restrictions to the schema. For example, the appliesTo field for each action in the schema allows users to specify which entity types can be used as principals and resources for that action. In the example [above](https://github.com/cedar-policy/rfcs/blob/feature/khieta/is-operator/text/0005-is-operator.md#basic-example), the schema could specify that applyFile should only apply to principals of type User and resources of type File. However, this doesn't help if the user is not using a schema, the action is unconstrained, or the policy refers to an action that may act on multiple principal/resource types.
```

While the example above is:

```
permit(
  principal is User,
  action == Action::"viewFile",
  resource is File in Folder::"public"
);
```

Therefore the action should be `viewFile` and `applyFile`

khieta:
Fixed the typo. Thanks @jeffsec-aws

aaronjeline (approved):

shaobo-he-aws (approved):

text/0005-is-operator.md
Swolebrain: One potentially irrelevant nitpick is that the fact that we can't index on type membership seems to me like it's more of a quirk of dynamodb, our data model, our indexing strategy, and our security practices (hashing everything to protect customer data from being visible) than it is a limitation of indexing itself. I could be wrong about this but i believe one could design an indexing strategy similar to R-trees that expressed entity type membership just like R trees express containments of points in space. 

It may not work in practice at all because EntityUIDs might not be subdivisible enough, I don't know. If that didn't work my next idea would be a reverse index akin to those used in full text search (also not supported by ddb). But there may be some other ways to cleverly index them that I haven't thought about. Point is that even if AVP can't do something right now because of our db schema, this doesn't mean we won't ever be able to do it, and it also doesn't mean customers can't do it.

john-h-kastner-aws (approved):

anwarmamat (approved):

khieta:
The final comment period (FCP) for this RFC is starting now, with intent to accept. The FCP will end 2023-07-28 at noon PT / 3pm ET / 7pm UTC.  Please add comments, and especially any objections, if you have any.  For more on the RFC process, see https://github.com/cedar-policy/rfcs.

andrewmwells-amazon (approved):

aaronjeline (approved):

PR #6, opened 2023-06-16
minor updates to template

khieta:
A couple quick updates to the RFC template:
* Add a "Status" so that people can easily check if an RFC is Active/Landed/Rejected (using the terminology in the README).
* Remove the "Adoption strategy" section because this information is already covered by other sections (e.g., "Drawbacks").

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -1,6 +1,7 @@
 - Start Date: (fill me in with today's date, YYYY-MM-DD)
 - Target Major Version: (2.x / 3.x / X.x)
 - Reference Issues: (fill in existing related issues, if any)
+- Status: Pending (change to Active if the RFC is accepted and merged)
 - Implementation PR: (leave this empty)
 
 # Summary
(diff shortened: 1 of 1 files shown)


0000-template.md
```diff
@@ -1,6 +1,7 @@
 - Start Date: (fill me in with today's date, YYYY-MM-DD)
 - Target Major Version: (2.x / 3.x / X.x)
 - Reference Issues: (fill in existing related issues, if any)
+- Status: Pending (change to Active if the RFC is accepted and merged)
```
cdisselkoen, line 4: Should we have more of a "timeline" section, which lists the start date, date the FCP started, date it became Active, etc?

For that matter, should FCP be a "status"?  Should we have a separate tracker for RFC statuses, maybe a table in the README or a separate top-level file?

khieta: I like the timeline idea. I'll add that in a new commit shortly.

Another possibility is to use GitHub labels. For example, we could label PRs with `pending`, `final-comment-period`, `active`, `landed`, or `rejected`. My only hesitation is that labels are easy to (intentionally or unintentionally) change whereas changing content in a file requires an explicit commit.

khieta: Updated the PR, see what you think.

@cdisselkoen I will be AFK from now until Monday, so feel free to take over this PR in the meantime. Otherwise I'll get back to it in a couple days.

cdisselkoen (approved):

shaobo-he-aws (approved):

PR #7, opened 2023-06-16
template groups RFC

mwhicks1:
RFC for template groups, which fixes https://github.com/cedar-policy/cedar/issues/106

[Rendered](https://github.com/cedar-policy/rfcs/blob/template-groups/text/0007-template-groups.md)

### DIFF
--- a/text/0004-template-groups.md
+++ b/text/0004-template-groups.md
@@ -1,0 +1,105 @@
+- Start Date: 2023-06-16
+- Target Major Version: 2.x
+- Reference Issues: https://github.com/cedar-policy/cedar/issues/106
+- Implementation PR: 
+
+# Summary
+
+Allow template policies to be grouped together, so they can be linked all at once. Doing so ensures that the application writer cannot mistakenly link one policy in the group separately from all the policies.
+
+# Basic example
+
+Consider this pair of template policies:
+```
+@id("policy1")
+permit(
+  principal == ?principal,
+  action, 
+  resource in ?resource);
+
+@id("policy2")
+permit(
+  principal == ?principal,
+  action == Action::"viewDoc",
+  resource in Directory::"DivisionDocs");
+```
+This pair represents a _role_: the linked principal can access any resource in the linked group (policy1), or it can view any document in a particular collection. A grouping mechanism would help connect them. For example:
+```
+@templategroup("role1")
+@id("policy1")
+permit(
+  principal == ?principal,
+  action, 
+  resource in ?resource);
+
+@templategroup("role1")
+@id("policy2")
+permit(
+  principal == ?principal,
+  action == Action::"viewDoc",
+  resource in Directory::"DivisionDocs");
+```
+Here we have labeled both policies with the same `@templategroup`; we link them at the same time using an API that references the group (`role1`), rather than individual policy (`policy1` and `policy2`).
+
+# Motivation
+
+We want to make sure that related template policies are always linked, together. Doing so ensures that permissions-level invariants are properly maintained.
+
+Grouping is also useful when reading and analyzing policies, since it makes evident that template groups represent a uniform whole that should be reasoned about together. A UX for policy management could display grouped policies together, in a different color, etc. Automated analysis must consider all possible linkages for templates, and the grouping adds useful constraints, e.g., that a linkage principal _p_ for `policy1` implies a linkage of the same _p_ for `policy2` in the same template group.
+
+# Detailed design
+
+We update the APIs to add semantics to the `@templategroup` annotation, as shown in the [Basic example](#basic-example). In particular, we will need to update the APIs for linking template policies to consider the `@templategroup` annotation, and not just the policy ID. A linkage for group `role1` should link any template having that group. A linkage for ID `policy1` should _fail_ if `policy1` also has a `@templategroup` annotation -- this is important for forbidding linkages to individual group policies.
+
+The updated APIs are for the CLI and core Cedar.
+
+# Drawbacks
+
+The main reason not to do this is that it adds extra code that customers _could_ implement themselves, using the existing annotation mechanism. However, doing so would lose the benefits of making the mechanism known to UIs, analysis tools, etc. and it would impose extra work on customers.
+
+
+# Alternatives
+
+We could add new syntax to Cedar that supports policy groups. For example
+```
+@id("role1")
+{ 
+    @id("policy1")
+    permit(
+      principal == ?principal,
+      action, 
+      resource in ?resource);
+
+    @id("policy2")
+    permit(
+      principal == ?principal,
+      action == Action::"viewDoc",
+      resource in Directory::"DivisionDocs");
+}
+```
+Here we allow any statements to by syntactically grouped within braces, and add an annotation on the group. This is potentially more visually clear to readers, and evokes [AWS IAM policies, which can consist of multiple statements grouped together](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_statement.html). But it would require extending the Cedar grammar.
+
+We could opt to do nothing at the level of the language, and rely on users of Cedar (such as Amazon Verified Permissions) to implement grouping using metadata of their choice, even as suggested here. We could also add the grouping syntax above, but not more than that, again relying on users to interpret the metadata annotations themselves. However, if grouping policies is likely to be generally useful, it seems useful to provide support for it.
+
+Users can also achieve the desired grouping effect by combining both templates into a single one:
+```
+@id("bothpolicies")
+permit(
+  principal == ?principal,
+  action, 
+  resource)
+when {
+   resource in ?resource ||
+   (action == Action::"viewDoc" && resource in Directory::"DivisionDocs")
+};
+```
+This has the drawback that the combined policy is more difficult to understand. (It also requires the template slot `?resource` to appear in the `when` condition, rather than the scope, which [RFC-0003](https://github.com/cedar-policy/rfcs/blob/main/text/0003-placeholders-in-conditions.md
+) aims to address.) Finally, this policy will not index very well in Verified Permissions because constraints on the `action` and `resource` do not appear in the policy scope (and could not, since there is more than one).
+
+# Adoption strategy
+
+Right now the APIs are hardcoded to assume that policies and templates have IDs, and linkage occurs by ID. These APIs either need to be generalized or a new API needs to be added to link according to group. Probably the easiest thing is to reinterpret "ID" in the API as "ID or Group", and to check groups first, IDs second, where linking by ID is forbidden if any group ID is present for the policy.
+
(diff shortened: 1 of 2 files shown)


text/0004-template-groups.md
```diff
@@ -0,0 +1,105 @@
+- Start Date: 2023-06-16
```
cdisselkoen, line 1: Filename should be 0007-template-groups.md following our convention of numbering the RFC after its PR number in this repo

text/0004-template-groups.md
```diff
@@ -0,0 +1,105 @@
...
+permit(
+  principal == ?principal,
+  action, 
+  resource)
+when {
+   resource in ?resource ||
+   (action == Action::"viewDoc" && resource in Directory::"DivisionDocs")
+};
+```
+This has the drawback that the combined policy is more difficult to understand. (It also requires the template slot `?resource` to appear in the `when` condition, rather than the scope, which [RFC-0003](https://github.com/cedar-policy/rfcs/blob/main/text/0003-placeholders-in-conditions.md
```
cdisselkoen, line 96: Link doesn't work (and won't until such time as RFC 3 gets merged); instead, you could consider using the link https://github.com/cedar-policy/rfcs/pull/3

cdisselkoen (changes requested):
I like the direction and motivation of this RFC; I think it would be great to have a mechanism to group and instantiate templates together, for all the reasons described in the Motivation section.  I do have one important concern with the proposed solution.

Currently, no attributes are "special" in Cedar, they are all completely up to the user to interpret.  This RFC would change that, and that's a large change in our philosophy, which is not currently addressed in the text of the RFC (IMO it's worth a mention in "Drawbacks".)  For this reason, I would lean toward preferring the first option under "Alternatives", namely, adding a dedicated syntax for this feature.  The syntax currently suggested under "Alternatives" might bear unfortunate resemblance to policy annotations, while actually being nothing of the kind.  So instead I propose this new syntax for discussion:
```
templategroup "role1" {
    @id("policy1")
    permit(...)

    @id("policy2")
    permit(...)
}
```
In the future, we could even extend this to support attributes on template groups themselves:
```
@foo("bar")
templategroup "role1" {
    ...
}
```
but we don't have to do that, and I certainly am not proposing that allowing attributes on template groups be included in this RFC.  Just keeping the door open.

text/0004-template-groups.md
```diff
@@ -0,0 +1,105 @@
...
+We want to make sure that related template policies are always linked, together. Doing so ensures that permissions-level invariants are properly maintained.
+
+Grouping is also useful when reading and analyzing policies, since it makes evident that template groups represent a uniform whole that should be reasoned about together. A UX for policy management could display grouped policies together, in a different color, etc. Automated analysis must consider all possible linkages for templates, and the grouping adds useful constraints, e.g., that a linkage principal _p_ for `policy1` implies a linkage of the same _p_ for `policy2` in the same template group.
+
+# Detailed design
+
+We update the APIs to add semantics to the `@templategroup` annotation, as shown in the [Basic example](#basic-example). In particular, we will need to update the APIs for linking template policies to consider the `@templategroup` annotation, and not just the policy ID. A linkage for group `role1` should link any template having that group. A linkage for ID `policy1` should _fail_ if `policy1` also has a `@templategroup` annotation -- this is important for forbidding linkages to individual group policies.
```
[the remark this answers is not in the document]
mwhicks1: Adjusted to make clear that we shouldn't treat these annotations specially.

text/0004-template-groups.md
mwhicks1: Fixed.

mwhicks1:
I have mixed feelings about baking in named template groups. It feels incongruous with the explicit _non_ naming of individual policies. That's why I wanted a similar mechanism. But I like the built-in syntax change.

cdisselkoen:
If the explicit name is a turn-off, we could still have the syntax
```
templategroup {
    @id("policy1")
    permit(...)

    @id("policy2")
    permit(...)
}
```
In this case, to indicate the name of the template group in the CLI, we might actually need to support annotations on template groups.

mwhicks1:
>  should the new group IDs be part of the same ID-space as policy/template IDs?

Good question -- they should be part of the same namespace so that there's no ambiguity when you are linking whether you mean to link by Policy ID or by Group ID.

mwhicks1:
> If the explicit name is a turn-off, we could still have the syntax

That's a good idea. What do others sink?

emina:
I vote for making the grouping mechanism general enough to allow template groups to _share_ templates.  For example, suppose that you define 3 templates T1, T2, and T3. Then, you want to users to have permissions T1 and T2, and others to have permissions T2 and T3.  In this case, we'd like to create two template groups that reuse these templates, so { T1, T2 } and { T2, T3 }.  This argues against inlining template definitions into the group definitions, but rather having the groups reference the names of the included templates.  

I've kept the description of the above use case abstract, but it comes from a real application.

shaobo-he-aws:
> I vote for making the grouping mechanism general enough to allow template groups to _share_ templates. For example, suppose that you define 3 templates T1, T2, and T3. Then, you want some users to have permissions T1 and T2, and others to have permissions T2 and T3. In this case, we'd like to create two template groups that reuse these templates, so { T1, T2 } and { T2, T3 }. This argues against inlining template definitions into the group definitions, and instead having the groups reference the names of the included templates.
> 
> I've kept the description of the above use case abstract, but it comes from a real application.

Would this use case motivate the approach where the users implement this functionality using existing APIs (e.g., filter policy set by template group annotation)?

emina:
> Would this use case motivate the approach where the users implement this functionality using existing APIs (e.g., filter policy set by template group annotations)?

If I’m understanding the suggestion correctly, users would manage template instantiations in this case manually, by searching for policies with specific annotations?  That’s certainly possible, but then template groups as a syntactic construct aren’t as useful as they could be.  Also, to achieve the desired effect through annotations, the user would need to employ some clumsy workarounds because we can’t bind more than one value to an annotation key.  For example, they’d pack the names of all template groups to which a template belongs into a string, something like `@templategroup(“G1, G2”)` and then parse the string.

cdisselkoen:
@emina's use case seems to want a syntax such as

```
@id("policy1")
permit(...)

@id("policy2")
permit(...)

@id("policy3")
permit(...)

templategroup "role1" { "policy1", "policy2" };
templategroup "role2" { "policy2", "policy3" };
```

Unfortunately this still runs afoul of the concern farther up this thread that suddenly particular attributes (`id`) would be interpreted specially by Cedar.  Makes me think that not having explicit policy names was a mistake.  Imagine if we could have

```
template "policy1":
permit(...)

template "policy2":
permit(...)

template "policy3":
permit(...)

policy "some static policy unrelated to this example":
permit(...)

templategroup "role1" { "policy1", "policy2" };
templategroup "role2" { "policy2", "policy3" };
```

I know Verified Permissions needs to assign its own UUIDs to each policy, but there's no reason they couldn't identify policies by their own UUIDs even if Cedar had syntax for explicit names.

emina:
Yes, I agree that it’s undesirable to have attributes with special meaning to Cedar.  

Perhaps sharing templates is a rare enough use case that it’s okay not to support it syntactically.  Though it’s also unsatisfying for the syntactic solution to not cover this use case too. 

I wonder if a cleaner alternative is to not add extra syntax, but provide dedicated APIs that make instantiation of template groups easy?  Another feature that could help with this is to extend the annotation mechanism to take not only strings as keys but also sets of strings (to avoid parsing).

D-McAdams:
The RFC should explain the customer input or end-goal that led to this proposal. What use case is this enabling that can't be achieved today? More specifically, imagine we were writing the user guide for this feature. How would we explain the situations in which to use it? Something along the lines of _"You may want to deliver a permissions management experience like X. You could do so by doing Y, but it has drawback Z. These drawbacks can be avoided by using template groups in the following manner..."_ (In fairness, I recall Mike explained it to me in person, but I've already forgotten the details.)

mwhicks1:
> The RFC should explain the customer input or end-goal that led to this proposal. What use case is this enabling that can't be achieved today?

The motivation for this feature comes form https://github.com/cedar-policy/cedar/issues/81. There, @WanderingStar gave two templates that together describe a single conceptual policy rule (which requires using a feature proposed in RFC https://github.com/cedar-policy/rfcs/pull/3)
```
@id("AdminAdmin")
permit(
  principal == ?principal,
  action, 
  resource in ?resource);     // take any action in the scope that you're admin on;

@id("AdminNavigate")
permit(
  principal == ?principal,
  action == Action::"Navigate", 
  ?resource in resource); // allow access up the tree for navigation
```
They said that they'd prefer a single template because with two "it would be possible to get into strange states where someone had `AdminAdmin` applied, but not `AdminNavigate`". The single template is more complex:
```
@id("Admin")
permit(
  principal == ?principal,
  action, 
  resource)
when {
    // take any action in the scope that you're admin on
    resource in ?resource

    // allow access up the tree for navigation
    || (action == Action::"Navigate" && ?resource in resource)
};
```
The proposal of this RFC aims for the best of both worlds: easier-to-read templates, but grouped together to ensure they are always linked together, and can be analyzed together.

cdisselkoen:
We could have (might want?) Cedar API support for template groups even if we don't provide Cedar syntax indicating which policies are grouped together.  For instance (and note I haven't thought about this proposal very hard yet)

```
impl PolicySet {
    pub fn mark_as_template_group(&self, group_id: PolicyId, ids: impl IntoIterator<Item = PolicyId>) -> Result<(), some_error_type>;
}
```

and then Cedar could at least allow instantiating the template group at once, using the specified `group_id` (see `link_group` below); and maybe even, prohibit instantiating the individual policies that were incorporated into the template group. 

```
impl PolicySet {
    pub fn link_group(&mut self, group_id: PolicyId, new_id: PolicyId, vals: HashMap<SlotId, EntityUid>) -> Result<(), PolicySetError>;
}
```

mwhicks1:
> We could have (might want?) Cedar API support for template groups even if we don't provide Cedar syntax

Indeed. While this might be useful, I think a library-level extension to `PolicySet` that implements would work, and would just layer on top of the functionality already there. People who want groups can just lift that code from the CLI into their own code, no? 

I worry about fixing our notion of a group to `PolicyID`, and having to fix that particular policies have group IDs, in the API. Doing it all in the CLI means the whole concept of group, and that you can be member of just one (say), is not baked into Cedar.

cdisselkoen:
Good point.  This addresses my concern.

We could also do the implementation as its own crate in the `cedar` repo, and have the CLI depend on that.  This might be more reusable than pointing people to copy code from the CLI repo.  But that would be more work, and at any rate we can discuss that on a PR and not on this RFC.

mwhicks1:
I'm canceling this RFC and reopening the original issue but applied to just to the CLI rather than the Cedar language and APIs (and hence it no longer needs to be an RFC).

PR #8, opened 2023-06-16
Partial schema validation RFC

john-h-kastner-aws:


mwhicks1 (commented):
(Got delayed, so pushing a couple of comments but not a complete review.)

max2me:
I just want to point out that ability to declare `additionalAttributes` on any Record type (as this RFC proposed) provides a fine-grained control over validation and makes it possible to accomodate scenarios that AWS Verified Access has. There service knows all top level keys of the context but might not have information about all attributes of nested Records.

text/0008-partial-schema-validation.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+We can now typecheck the policy under the assumption that `access_level` is a `String`,
+so the partial schema validator will report an error -
+this time for the left `>` operand.
+Including `"additionalAttributes": true` in the schema informs the validator that the listed attributes are still incomplete,
+so an expression `principal.role == "Admin"` would still be accepted by the validator.
+
+Alternatively, we might have complete information about the attributes present for `User` entities.
+When we know that `access_level` does not exist, the `shape` for the principal entity type can include `"additionalAttributes": false` to specify that there are no undeclared attributes for this entity type.
```
khieta, line 125: Note that `"additionalAttributes": false` is the default, so it is unnecessary to say this explicitly.

text/0008-partial-schema-validation.md
```diff
@@ -0,0 +1,207 @@
...
+With this information, the validator will assume that the principal may be in any group,
+so any `in` expression with a `User` entity as the left operand will have type `Boolean` instead of `False`.
+This will cause the validator to accept the policy so long as the rest of the policy typechecks.
+
+## Schema Format
+
+We extend the schema format to support the above example schema.
+Specifically, we add the following properties to the JSON:
+
+* Any record type, including the record types for entity attributes and action context attributes, may include `"additionalAttributes": true` to specify that values having that record type may have attribute other than those declared.
```
khieta, line 174: may have attribute -> may have attributes

text/0008-partial-schema-validation.md
khieta: I see now that it's noted in a later section, but still good to include here to head off confusion.

text/0008-partial-schema-validation.md
```diff
@@ -0,0 +1,207 @@
...
+* An action definition may include `“additionalMemberOf”: true` having the same purpose as `additionalMemberOfTypes` for entity type, but applying to the `memberOf` list.
+
+All properties added to the schema default to `false`,
+so an entity declaration in a schema read by the partial schema validator will be interpreted the same as when read by the standard validator.
+
+
+## Effect on the standard validator
+
+This change does not necessarily require any change to the behavior of the standard validator,
+but the additions to the schema format could be soundly supported.
```
khieta, line 185: Will the implementation make these changes, or is this noting that the changes are possible?

text/0008-partial-schema-validation.md
```diff
@@ -0,0 +1,207 @@
...
+while still reporting an error for any undeclared entity types appearing in policies.
+
+# Drawbacks
+
+The primary drawback to this approach to partial schema validation is that it does not provide any soundness guarantees which could be provided by an inference based approach to partial schema validation.
+When partial schema validation as proposed accepts a policy, it gives no guarantees about what errors could occur when evaluating the policy.
+
+By adding this form of partial schema validation, we may also make it more difficult to add schema type inference later.
+A sound, type inference based, approach to partial schema validation would report _more_ type error than this proposal,
+so it could not be used as a drop-in replacement.
```
khieta, line 199: type error -> type errors

Also: why would an inference-based approach result in more errors?

text/0008-partial-schema-validation.md
```diff
@@ -0,0 +1,207 @@
...
+We would instead need to offer it as a separate validation mode,
+but supporting multiple partial schema validation options could be confusing for users.
+
+# Alternatives
+
+* Policy validation is not required to use Cedar,
+  so we could continue to support validation only for users who are able to provide a complete schema.
+* We could design and implement a type inference algorithm for Cedar.
```
khieta, line 207: It would be useful to have more details about the schema inference option to determine whether it is a viable alternative.

luxas:
Hi! 👋 What is the status on this RFC? I see that the `partial-validate` feature is at least partially implemented. Did the design change as part of implementation? I just noticed the `additional_attributes` bool on the `RecordType`, which led me here. 

One use-case I have (which is related, but slightly different) is to be able to say for some struct type that "any property that is not known, should have this type"; e.g. string, just like OpenAPI v3's `additionalProperties`. This would allow serializing a string-string map with arbitrary values into Cedar as-is, with something like `additionalProperties { ty: TypeVariant::String }`. You can do this today with a has check for a type that it empty, e.g.
```
type LabelsMap = {}
type Metadata = {
   labels: LabelsMap,
}
entity Resource = {
  metadata: Metadata,
}

permit(principal, action, resource == Resource) when {
  resource.metadata.labels has foolabel && resource.metadata.labels.foolabel == "bla"
};
```
The VS Code schema validator (Cedar 4.4.0) allows this, even in the default validation mode, although the type of `resource.metadata.labels.foolabel` is not known (which could result in a schema typing error not being caught and thus a runtime error). However, `cedar validate` in strict mode correctly tells this is (as per the schema) an impossible policy. At runtime, using `cedar authorize`, this policy works well, but indeed requires the `has` guard.

If `additionalProperties` was part of the schema, the `has` guard maybe wouldn't be needed, and it would be possible to typecheck the value. However, I realize this would probably get dangerously close to maps, which are not part of Cedar today (except in `tags`), and thus might make this harder to analyze.

I'm happy to open a dedicated issue on this topic, as it's only tangientally related to `additional_attributes` as used in this RFC, but wanted to mention this here first.

luxas:
Thanks for the response! 

I think that for my case, I'll work around it by either:

a) "bottom-up": dynamically updating the schema of the "open" Cedar record (which in JSON/OpenAPI is an unstructured string-string map) to include the record attributes that become well-known by actually referencing them in a policy (e.g. by using #74), or
b) "top-down": walking objects in the database to discover which labels (record attribute keys) "exist" in the system, and thus populating the schema, or
c) (what I have now) model the map as a `Set<{key: String, value: String}>` which works well, but is not as expressive and intuitive to use.

So I don't think I need this feature per se, if I do any of the above, as I anyways need the soundness guarantees that the normal schema validator provides.

luxas:
> a) Sounds tricky since you'll need to infer the type of the attributes based on how they're used. Doing it soundly in the general case will be difficult.

In this case, I know that `key` is always a string (as it's modelled from a JSON object), and I know the Cedar type of `value` too, thanks to the OpenAPI schema (and in 90%+ of cases just a string). So what I infer from policies is just the attribute names/keys, basically, which I think should work.

> A good workaround if the limited expressiveness isn't an issue, but I'm curious why you have this instead of using entity tags. Entity tags were largely motivated by supporting this pattern while allowing direct access to the values by hasTag and getTag operations.

The reason was (until I read your message and realized how to do it, thanks!) that I didn't realize that I don't have to put tags on the "parent" object, but can declare an intermediate type. So the simplified practical example:

```
namespace before {
    entity Pod = {
        metadata: ObjectMeta,
    };
    type ObjectMeta = {
        labels: Set<StringKeyValue>,
        annotations: Set<StringKeyValue>,
    };
    type StringKeyValue = {
        key: String,
        value: String
    };
    action update appliesTo {
        principal: User,
        resource: Pod
    };
    entity User;
}
```

This is the existing case for https://github.com/awslabs/cedar-access-control-for-k8s/

When I was thinking about tags, I somehow thought about putting the tags on ObjectMeta (the "parent"), which didn't really fit nicely, especially as there were two of these open records, `labels` and `annotations`.

With this discussion, I realized that nothing stops me from declaring an intermediate entity, not sure why I didn't realize this before, when reading the tags RFC.

```
namespace after {
    entity Pod = {
        metadata: ObjectMeta,
    };
    type ObjectMeta = {
        labels: StringMap,
        annotations: StringMap
    };
    entity StringMap = {
        
    } tags String;
    action update appliesTo {
        principal: User,
        resource: Pod
    };
    entity User;
}
```

Now I can write:

```
permit(
    principal,
    action,
    resource is after::Pod
) when {
    resource.metadata.labels.hasTag("owner") && resource.metadata.labels.getTag("owner") == "lucas" 
};
```

And to support the case of figuring out all of the available labels (when not caring about the values), to make sure there aren't other tags/labels I don't want to be there, but also don't know the name of, I can add:

```
entity StringMap = {
    keys: Set<String>
} tags String;
```

such that I can make sure the object doesn't contain other, unauthorized labels, only the two allowlisted ones here (owner, other-ok-label):

```
permit(
    principal,
    action,
    resource is after::Pod
) when {
    resource.metadata.labels.hasTag("owner") &&
    resource.metadata.labels.getTag("owner") == "lucas" &&
    ["owner", "other-ok-label"].containsAll(resource.metadata.labels.keys) 
};
```

For posterity, this is how it would have looked like if we had open records, not very different:

```
permit(
    principal,
    action,
    resource is after::Pod
) when {
    resource.metadata.labels has "owner" &&
    resource.metadata.labels.owner == "lucas" &&
    ["owner", "other-ok-label"].containsAll(resource.metadata.labelKeys) 
};
```

Thanks for helping with the thinking! 
It might be good to document this somewhere in the docs as "so, you think you need an open record" 😄

PR #9, opened 2023-06-19
Disallow embedded whitespace in EntityUID RFC

D-McAdams:
[rendered](https://github.com/D-McAdams/rfcs/blob/main/text/0009-disallow-whitespace-in-entityuid.md)


shaobo-he-aws:
IMHO, the issue that this RFC attempts to address derives from the assumption that users are not aware of this particular piece of Cedar syntax. In other words, a well-informed user should avoid to use namespace strings directly. Instead, they should use a normalized form such as Cedar ASTs or printed ASTs.

text/0005-disallow-whitespace-in-entityuid.md
```diff
@@ -0,0 +1,94 @@
...
+3. Customers are anticipated to build meta-permissions layers that restrict callers to manipulating policy store contents for only certain namespaces. This may lead to policies such as `forbid(...) when {context.namespace = "ExampleCo::Photoflash"};`. There is a risk that an unauthorized actor could bypass this restriction by using a namespace with embedded spaces.
```
shaobo-he-aws, line 61: If we stick to the tenet where we should avoid using direct string representations or use normalized ones, we'd recommend `namespace` attribute to be a record or a string containing normalized namespaces.

text/0005-disallow-whitespace-in-entityuid.md
```diff
@@ -0,0 +1,94 @@
...
+  resource
+);
+```
+
+This breaking change is only being considered due to the relative newness of Cedar and the probability of breaking a real production scenario being low, and it can only be done if the community agrees that altering this behavior will pay long term dividends that outweight the risk of near-term impacts.
+
+## Alternatives
+
+We could change the behavior of `EntityTypeName::from_str()` without changing the syntax used in Cedar policies themselves. In `EntityTypeName::from_str()`, whitespace would be disallowed throughout -- either before, after, or in the middle of the typename. In Cedar policies, whitespace would still be allowed in every position it’s allowed today.  This would avoid a breaking change to the Cedar language and avoid diverging from the behavior of most mainstream programming languages, which do allow whitespace around namespaces and namespace separators.  This alternative would fully address the Examples 1, 2, and 3 in this PDF, and while it would not address the malicious policy example given, that should be weighed against the impact of a breaking change to the Cedar language and diverging from other commonly-used languages.
```
shaobo-he-aws, line 93: The alternative seems incompatible with the proposal summary where it says we should disallow whitespaces and comments in `EntityUID`, which consists of an entity type and a uid.

D-McAdams:
The RFC has been updated to align with the feedback received so far. What was originally the "Alternative Approach" is now the recommended approach.

text/0005-disallow-whitespace-in-entityuid.md
mwhicks1: @shaobo-he-aws to be clear, you are suggesting that Rust codewriters use the [EntityTypeName](https://github.com/cedar-policy/cedar/blob/main/cedar-policy/src/api.rs#L938) or [EntityNameSpace](https://github.com/cedar-policy/cedar/blob/main/cedar-policy/src/api.rs#L959) structs directly? I.e., when parsing a JSON schema, the parser should invoke the `from_str` methods of these structs, and then when using them as keys in a database, they should invoke `fmt` to output to a normalized string?

That seems doable to me, but I can understand the concern raised in the RFC that developers may not think to do this. Indeed, I think the third example is something that application writers could easily miss.

text/0005-disallow-whitespace-in-entityuid.md
```diff
@@ -0,0 +1,98 @@
...
+3. Customers are anticipated to build meta-permissions layers that restrict callers to manipulating policy store contents for only certain namespaces. This may lead to policies such as `forbid(...) when {context.namespace = "ExampleCo::Photoflash"};`. There is a risk that an unauthorized actor could bypass this restriction by using a namespace with embedded spaces. 
+
+While it is technically possible for applications to mitigate this risk by diligently using Cedar tooling to normalize the values, the little-known nature of this Cedar behavior implies that few will know they *should* normalize the value. As a point of reference, application developers who have worked with Cedar extensively for over a year were bitten by this bug in production. Hence, this is likely to result in bugs in many other Cedar-based implementation with similar logic, risking a perception that Cedar is fragile or unsafe.
+
+## Detailed design
+In `EntityTypeName::from_str()`, whitespace would be disallowed throughout -- either before, after, or in the middle of the typename.
+
+## Drawbacks
```
mwhicks1, line 62: Another drawback that I don't think is called out here is the conceptual incongruity of this change. 

A namespaced type that is acceptable in a policy, e.g., `ExampleCo  ::  Photoflash  ::  User`, will be rejected when passed as a string argument to `EntityTypeName::from_str()`. This could be confusing and potentially problematic; e.g., if a schema namespace parameter is `"ExampleCo  ::  Photoflash  ::  User"` then the schema will be rejected, and thus policies will not be validated against it. 

The mitigation is to have a clear error message about what's going on. I think that should be doable because we'll have to write a custom parser for this more restricted namespaced type anyway.

cdisselkoen (approved):
Thanks for addressing the comments.  I'm happy with this proposal and agree that the benefits outweigh the risk of making a breaking change to the behavior of `EntityTypeName::from_str()`.

0xjjoyy:
Does `AuthorizationQuery` call `EntityTypeName::from_str` ?

text/0005-disallow-whitespace-in-entityuid.md
shaobo-he-aws: Yes, that's my idea.

text/0005-disallow-whitespace-in-entityuid.md
```diff
@@ -0,0 +1,98 @@
...
+3. Customers are anticipated to build meta-permissions layers that restrict callers to manipulating policy store contents for only certain namespaces. This may lead to policies such as `forbid(...) when {context.namespace = "ExampleCo::Photoflash"};`. There is a risk that an unauthorized actor could bypass this restriction by using a namespace with embedded spaces. 
+
+While it is technically possible for applications to mitigate this risk by diligently using Cedar tooling to normalize the values, the little-known nature of this Cedar behavior implies that few will know they *should* normalize the value. As a point of reference, application developers who have worked with Cedar extensively for over a year were bitten by this bug in production. Hence, this is likely to result in bugs in many other Cedar-based implementation with similar logic, risking a perception that Cedar is fragile or unsafe.
+
+## Detailed design
+In `EntityTypeName::from_str()`, whitespace would be disallowed throughout -- either before, after, or in the middle of the typename.
```
shaobo-he-aws, line 60: Just to double check. Tokens (*whitespace* and *comments*) that are ignored during parsing should not appear in valid entity type names, right? In other words, composition of `EntityTypeName::from_str` and `EntityTypeName::to_string` over a valid entity type name should produce the same string.

text/0005-disallow-whitespace-in-entityuid.md
```diff
@@ -0,0 +1,98 @@
...
+2. Policy Stores can implement logic that relies on string comparisons against the EntityTypeName. In a real issue, an application using Cedar sought to preclude callers from passing Actions in the inline slice of entity definitions. It did so by checking if an EntityTypeName matched `.*::Action`. It presumed that `:: Action` was invalid syntax and would be rejected by the Cedar evalator, the same as any other syntatically invalid input. This resulted in a bug, as it allowed callers to bypass the extra validation that the application sought to enforce.
+3. Customers are anticipated to build meta-permissions layers that restrict callers to manipulating policy store contents for only certain namespaces. This may lead to policies such as `forbid(...) when {context.namespace = "ExampleCo::Photoflash"};`. There is a risk that an unauthorized actor could bypass this restriction by using a namespace with embedded spaces.
```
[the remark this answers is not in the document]
mwhicks1: The example doesn't rely on Cedar being able to extract the namespace from an entity. I think this is where Darin is coming from:

The `context.namespace` would be populated by the application submitting the Cedar request, and the action (not shown) would relate to adding a policy to the store (say). The concern is that `context.namespace` is populated in part by user data. The application would want to make sure the user data is a legit namespace, so it would run it through the Cedar parser. What if the user data was `"ExampleCo::  Photoflash"`? The parser would _accept_ it as of right now. That checked, the application needs to _normalize_ this data, so that it stores `"ExampleCo::Photoflash"` in `context.namespace`, not `"ExampleCo::  Photoflash"`. This way, it matches the `when` condition as expected. With the change proposed in this RFC, the normalization step is not required, which means the application code will be less error prone.

mwhicks1:
I'm happy now. I think we should make this change.

cdisselkoen:
@0xjjoyy 

> Does `AuthorizationQuery` call `EntityTypeName::from_str` ?

There is no `AuthorizationQuery` in Cedar.  Do you mean [Request](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Request.html)?  If so, constructing the `EntityUid`s to pass to `Request::new()` will require going through `EntityTypeName::from_str`, either directly or indirectly, so it would be covered by this change.  More generally, it's internally an invariant of Cedar that any value of type `EntityUid` has effectively already been validated and normalized, with whitespace removed.  This was already true and will remain so.  The proposal in this RFC is just that `EntityTypeName::from_str` will not perform normalization, instead it will expect an already-normalized string and fail otherwise.

cdisselkoen:
Crossed wires, sorry.  I was referring to the main Cedar repo, which is implemented in Rust (github.com/cedar-policy/cedar).  The Java bindings depend on the Rust library in that repo.  We should align the terminology better between Java and Rust -- please feel free to open an issue on the Java bindings to update the terminology

text/0005-disallow-whitespace-in-entityuid.md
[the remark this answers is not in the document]
aaronjeline: Gotcha, I understand now.

andrewmwells-amazon (approved):

shaobo-he-aws (approved):

D-McAdams:
The final comment period (FCP) for this RFC is starting now, with intent to accept this RFC.  Due to the potential security concerns of this RFC, we would like to shorten its FCP, and have the FCP end Friday 2023-06-23 at noon PT / 3pm ET / 7pm UTC.  If no new unresolved major objections are voiced by then, this RFC will be merged.
For more on the RFC process, see https://github.com/cedar-policy/rfcs.

anwarmamat (approved):

shaobo-he-aws (approved):

PR #10, opened 2023-06-22
Clarify the template instructions for RFC title

D-McAdams:
Updates the template instructions to clarify that `# RFC Template` can be fully replaced with the submitter's title


cdisselkoen (changes requested):
I love this change but let's not include the text of RFC 9 in the PR :)

D-McAdams:
I just saw that and was looking at how to amend. Blah...

PR #11, opened 2023-06-23
Validate requests

khieta:
[rendered](https://github.com/cedar-policy/rfcs/blob/rfc/khieta/validate-queries/text/0011-validate-requests.md)

### DIFF
--- a/text/0011-validate-requests.md
+++ b/text/0011-validate-requests.md
@@ -14,25 +14,27 @@
 
 ## Summary
 
-Given a schema, _validation_ checks that policies conform to the schema, and _schema-based parsing_ checks that entities conform to the schema.
-This RFC proposes to add support for checking that a query conforms to a schema.
+Given a schema, _validation_ checks that policies conform to the schema, and _schema-based parsing_ checks that entities and context conform to the schema.
+This RFC proposes to add support for checking that a request conforms to a schema.
 
 ## Basic example
 
 Say the schema specifies that an action `readFile` should only be used with principals of type `User` and resources of type `File`.
 This RFC proposes to change the signature of the `Request` constructor to add an optional schema argument that can be used to check  argument types when constructing a `Request`.
 
-Examples (w/ invalid syntax for EUIDs):
+Examples (using informal syntax for EUIDs):
 
 ```rust
 Request::new(Some(principal), Some(action), Some(resource), context, None); // no schema; works like before
+
 Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(File::"secret_file.txt"), context, Some(schema)); // ok
-Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(User::"bob"), context, Some(schema)); // error, invalid resource
+
+Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(Folder::"some_folder"), context, Some(schema)); // error, invalid resource
 ```
 
 ## Motivation
 
-(Example migrated from #5)
+### Example 1: Unexpected authorization decisions
 
 Say that a user wants to allow any `User` to view files in the folder `Folder::"public"`.
 Currently, to achieve the desired behavior, the user should define a schema with the action entity `readFile` that can apply to principals of type `User` and resources of type `File`, and then write the following policy.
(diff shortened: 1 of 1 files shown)


text/0011-validate-requests.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Summary
+
+Given a schema, _validation_ checks that policies conform to the schema, and _schema-based parsing_ checks that entities conform to the schema.
+This RFC proposes to add support for checking that a query conforms to a schema.
+
+## Basic example
+
+Say the schema specifies that an action `readFile` should only be used with principals of type `User` and resources of type `File`.
+This RFC proposes to change the signature of the `Request` constructor to add an optional schema argument that can be used to check  argument types when constructing a `Request`.
```
mwhicks1, line 23: Do we have to change the signature of the constructor? Can we add a new constructor instead, with the schema parameter added (whereas the old one stays as is) ?

text/0011-validate-requests.md
```diff
@@ -0,0 +1,72 @@
...
+
+Say that a user wants to allow any `User` to view files in the folder `Folder::"public"`.
+Currently, to achieve the desired behavior, the user should define a schema with the action entity `readFile` that can apply to principals of type `User` and resources of type `File`, and then write the following policy.
+
+```
+permit(principal, action == Action::"readFile", resource in Folder::"public");
+```
+
+Assuming that they pass in inputs where the principal is of type `User` and the resource is of type `File`, this will work as expected.
+However, it could also return "allow" for other principal types like `Folder` or resource types like `User`.
```
mwhicks1, line 45: This feels contrived. Why would I have a `Folder` principal type when using the `readFile` action? Moreover, why does `User` as a resource type matter in this case? Presumably no `User` entities exist in `Folder::"public"`. Is there a way to tweak your "for example" to make the chances of a bug occurring seem more plausible?

text/0011-validate-requests.md
```diff
@@ -0,0 +1,72 @@
...
+This is a bug on the user's end: they shouldn’t make a query with the action `readFile` when the principal is not a `User` and the resource is not a `File`.
+They are breaking the contract they stated in their schema.
+Nevertheless, it might be confusing that such a query does not result in a "deny".
+
+This RFC proposes to add a check, when constructing the `Request` type, that the query is valid according to the schema.
+
+## Detailed design
+
+Validation and schema-based parsing are both "opt-in" in the sense that they are not required, so we'll the same for query validation.
```
mwhicks1, line 54: Typo ("we'll the same for")

text/0011-validate-requests.md
```diff
@@ -0,0 +1,72 @@
...
+Validation is done by a separate API (`Validator::validate`), while schema-based parsing is done by passing a (optional) schema argument into various parsing APIs, e.g., `Entities::from_json_str` and `Context::from_json_str`.
+If the schema argument is `None`, then parsing does no validation.
+
+This RFC proposes to modify the constructor for the `Request` type to add an optional schema type, similar to schema-based parsing.
+
+## Drawbacks
+
+- This is a breaking change that will affect any users that are manually constructing `Request`s (which is likely all users).
```
mwhicks1, line 62: Does it have to be? For example, we have a `new_with_unknowns` constructor (https://github.com/cedar-policy/cedar/blob/f233431efb82a1323e35cba2023b34613b8c99e5/cedar-policy-core/src/ast/request.rs#L91); could we have a `new_with_validation` constructor? 

For that matter -- should this RFC apply to partial evaluation-ready `Request`s too?

text/0011-validate-requests.md
khieta: I guess a more natural mistake in this setup would for the resource to be a `Folder`. In this case, the result of the policy could be "allow" (assuming that folders can be nested). Is this a more convincing use case?

I'll do some thinking to come up with a better example. Open to suggestions in the meantime.

text/0011-validate-requests.md
khieta: To answer your second question: yes, I think we'll want validation supported for partial `Request`s too, once those are exposed.

mwhicks1: > The downside is that this will break existing users; the benefit is that this will force them to be aware of the new feature and start validating their requests.

Good characterization of the tradeoff. IMO, I'd rather go with two APIs because I don't want to break existing users, and because I expect that the non-validation path is the common one. People will largely use the validation path for debugging.

How would you feel about having two `new` methods, one with 4 parameters and one with 5? (I'm pretty sure Rust supports overloading.) The five-parameter one would call validation, and then invoke the four-parameter one.

text/0011-validate-requests.md
mwhicks1: I like the `Folder` use-case. Then you'd have to change your example, though, to use `is` for the resource, rather than the principal, right?

text/0011-validate-requests.md
shaobo-he-aws: IIRC, Rust doesn't support function overloading.

text/0011-validate-requests.md
```diff
@@ -0,0 +1,72 @@
...
+
+```rust
+Request::new(Some(principal), Some(action), Some(resource), context, None); // no schema; works like before
+Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(File::"secret_file.txt"), context, Some(schema)); // ok
+Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(User::"bob"), context, Some(schema)); // error, invalid resource
+```
+
+## Motivation
+
+(Example migrated from #5)
```
shaobo-he-aws, line 35: It seems #5 doesn't have a hyperlink when rendered.

text/0011-validate-requests.md
cdisselkoen: Rust does not support function overloading.  You can use traits to give the same function different definitions on different types, but the signature has to match with the exception of the type `T` that's implementing the trait.

khieta:
I've revised the RFC and am looking for new reviews. Summary of changes:
1. Revised the User/File/Folder example per Mike's suggestion.
2. Added an additional example to the motivation section to show why validating requests is important in conjunction with standard validation.
3. Added an "unresolved question" about changing the `Request::new` API vs. adding a new `Request::new_with_validation` API.

Please weigh in on point (3) if you haven't already.

text/0011-validate-requests.md
john-h-kastner-aws: I somewhat in favor of making the breaking change.

cdisselkoen: Same.  It seems like a worthwhile change to force people to adapt to when updating to Cedar 3.x (whenever 3.x happens).

cdisselkoen: I'm imagining that the actual performance overhead of request validation would be negligible, noting that the issue of validating the `context` is orthogonal (users can already do so or not as they construct the `Context` object which is passed to `Request::new()`), so the only performance overhead would be checking the `principal`, `action`, and `resource` are of compatible types. (Right?)

mwhicks1: To clarify: Users will validate the context if they do schema-based parsing to construct. They aren't required to do that.

cdisselkoen: Correct.  It's orthogonal to whether they validate the request (which would only happen if the supply the schema to `Request::new()`).  They could do either one alone, or both, or neither.

khieta:
Revised! Notable changes:
* Per in-person discussion yesterday, instead of modifying the `Request::new` constructor, we will add a new `Request::new_with_validation` constructor.
* Added more concrete details about implementation (copied below for reference).

> The new API will have the following signature:
> 
> ```rust
> pub fn new_with_validation(
>         principal: Option<EntityUid>,
>         action: Option<EntityUid>,
>         resource: Option<EntityUid>,
>         context: Context,
>         schema: &Schema
>     ) -> Result<Request> {
>         ...
>     }
> ```
> 
> (This is the same as the signature for `Request::new`, aside from the schema argument and return type.)
> 
> If the action is `None` (indicating that it is "unspecified"), then the function performs no additional checks. If the action is `Some`, then the function checks that the specified action is present in the schema, and that the principal and resource are consistent with the action's `appliesTo` lists in the schema.

text/0011-validate-requests.md
```diff
@@ -20,16 +20,16 @@ This RFC proposes to add support for checking that a request conforms to a schem
...
 
 ```rust
-Request::new(Some(principal), Some(action), Some(resource), context, None); // no schema; works like before
+Request::new(Some(principal), Some(action), Some(resource), context); // original API; unchanged by this RFC
 
-Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(File::"secret_file.txt"), context, Some(schema)); // ok
+Request::new_with_validation(Some(User::"alice"), Some(Action::"readFile"), Some(File::"secret_file.txt"), context, Some(schema)); // returns a Request
 
-Request::new(Some(User::"alice"), Some(Action::"readFile"), Some(Folder::"some_folder"), context, Some(schema)); // error, invalid resource
+Request::new_with_validation(Some(User::"alice"), Some(Action::"readFile"), Some(Folder::"some_folder"), context, Some(schema)); // returns an error (invalid resource)
```
mwhicks1: Drop the `Some(...)` part here, since the schema is required for this new constructor.

mwhicks1 (approved):

aaronjeline (approved):

john-h-kastner-aws (approved):

cdisselkoen (approved):

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

khieta:
Based on discussion, we generally support the addition of the `Request::new_with_validation` constructor. However, I've decided that this change does not merit the bar for an RFC, so rather than moving into FCP, I will be closing the issue. The core content of the RFC will be moved to an issue in the `cedar` repository.

khieta:
Moved to `cedar` [issue #191](https://github.com/cedar-policy/cedar/issues/191).

PR #12, opened 2023-06-26
Add two more affected functions and accepted date

andrewmwells-amazon:
*Description of changes:*
Add two more affected functions (`EntityNamespace::from_str()` and `EntityUid::from_str()`) and accepted date (today).

### DIFF
--- a/text/0009-disallow-whitespace-in-entityuid.md
+++ b/text/0009-disallow-whitespace-in-entityuid.md
@@ -9,7 +9,7 @@
 
 - Start Date: 2023-06-19 
 - Date Entered FCP: 2023-06-21
-- Date Accepted:
+- Date Accepted: 2023-06-26
 - Date Landed:
 
 ## Summary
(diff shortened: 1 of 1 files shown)


cdisselkoen:
@andrewmwells-amazon can you also update the link for "Implementation PR" to be cedar-policy/cedar#133? Thanks

shaobo-he-aws:
I tend to object to these changes. I tend to disagree. IIRC, the motivation of the RFC is based on the usage pattern where an entity id is referred by two components (typename + uid) instead of a unified struct (i.e., `EntityUid`). And this usage pattern kind of justifies the discrepancy from the Cedar grammar. I think the rule of thumb is to avoid such discrepancy as much as possible unless we have a compelling use case.

cdisselkoen:
I think we should make the change @andrewmwells-amazon is currently proposing.  The RFC motivation speaks to the case where two strings are treated as different by JSON (eg in the Cedar schema) but the same by Cedar.  This trap occurs not just with entity types, but also in namespaces and full UIDs.  Consider two schemas, one declared with namespace `"Foo "` and the other with `"Foo"` -- Cedar will see these as the same namespace, which could cause customer confusion.  

More generally, people don't expect whitespace and comments to be valid inside JSON strings, which is what the status quo has.  Adjusting all three functions should cover all of these cases where whitespace and comments are currently valid inside JSON strings.

shaobo-he-aws:
Got it. I didn't expect namespaces to appear in the schemas other than being entity type names.

shaobo-he-aws (approved):

PR #13, opened 2023-06-30
Add enums to schema

aaronjeline:
[Rendered](https://github.com/cedar-policy/rfcs/blob/enums/text/0013-schema-enums.md)


text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- integration of this feature with other existing and planned features
+- cost of migrating existing Cedar applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
```
khieta, line 70: Still needs to be filled out.

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+## Timeline
+
+- Start Date: 2023-06-30
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+Add support for enumerations of `bool`, `string`, and `long` types in schemas, and use this information when doing policy validation and schema-based-parsing.
```
khieta, line 17: I imagine enumerations will be most useful for strings, so I think we should consider restricting the scope of this RFC to only `string`s.
* Enumerations could possibly be useful for `long`s too (e.g. for error codes), but that case can also be encoded using `string`s. The `long` type doesn't necessarily give an advantage because I doubt you would want to do `long`-specific operations on the values in the enum (e.g., addition, less-than comparison). 
* I don't think enums will be useful for `bool` (what is the point of a field that always has to be `true`?).

If I'm wrong, and you had specific use cases in mind, then please add them to the RFC.

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+```
+In this case, we don't specify the `Enum` is a subset of `String`, as that's implicit.
+
+## Unresolved questions
+
+* Do we have any mechanism for enums of extensions types?
+* Do we want any way to support exhaustive case statements?
+* Any reason to support empty enums?
```
khieta, line 97: > Any reason to support empty enums?

I don't think so. The meaning of that would just be a field that has no valid value, right?

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+}
+```
+In this case, we don't specify the `Enum` is a subset of `String`, as that's implicit.
+
+## Unresolved questions
+
+* Do we have any mechanism for enums of extensions types?
+* Do we want any way to support exhaustive case statements?
```
khieta, line 96: > Do we want any way to support exhaustive case statements?
 
That sounds like an RFC in its own right. Maybe flesh out some of the details and proposed syntax here?

text/0013-schema-enums.md
aaronjeline: Mike had a use case for bools, I'll add that. Longs just seems intuitive, but you're right you could model it with strinsg as well

text/0013-schema-enums.md
aaronjeline: correct

text/0013-schema-enums.md
aaronjeline: Agree it should be it's own RFC, and it's a two way door to not have it in this one

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+Add support for enumerations of `bool`, `string`, and `long` types in schemas, and use this information when doing policy validation and schema-based-parsing.
+
+## Basic example
+
+Maintaining backward compatibility with JSON schema as much as possible, we can match its enum feature. For example:
```
cdisselkoen, line 21: "backward compatibility" might be a strange way to word this, perhaps just "compatibility" or "inspiration from"?

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+      "enum" : ["red", "blue", "green"]
+    }
+}
+```
+
+The enum part limits the color attribute to one of three possibilities. We would permit enumerations of this flavor for numbers and booleans too.
+
+## Motivation
+
+Enumerations are a common need, and this change supports them in Cedar. It avoids the mistake of expecting a string (say) to match only a few values, but then it inadvertently matches too many.
```
cdisselkoen, line 37:
```suggestion
Enumerations are a common need, and this change supports them in Cedar. It avoids the mistake of expecting a string (say) to match only a few values, but then encountering a different value at runtime.
```

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+
+Enumerations are also useful for schema-based parsing, and reject entities/contexts that don't properly pick a value from the enumeration.
+
+
+## Detailed design
+
+Schema specification:
+* The types `String`, `Bool`, and `Long` MAY have a field `enum` in their specification. 
+* If the field `enum` is not present the semantics is unchanged.
+* If the field `enum` is present, it's value MUST be a non-empty list.
```
cdisselkoen, line 49:
```suggestion
* If the field `enum` is present, its value MUST be a non-empty list. 
```

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+  "type" : "Record",
+  "attributes" : {
+    "color" : {
+      "type" : "Enum",
+      "values" : ["red", "blue", "green"]
+    }
+  }
+}
+```
+In this case, we don't specify the `Enum` is a subset of `String`, as that's implicit.
```
cdisselkoen, line 91: Having the "parent type" implicit is an interesting idea.  One could imagine even supporting enums of heterogeneous types, or of types like entity-literal. 
```
"values": ["red", 2, User::"alice"]
```
That's probably a bad idea, but with the syntax here, we'd be leaving the door open more easily to extend enums to other types.

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Unresolved questions
+
+* Do we have any mechanism for enums of extensions types?
```
cdisselkoen, line 95: This would be much easier with the syntax in Alternative 2

text/0013-schema-enums.md
cdisselkoen: The only reason I could see this would be if you are generating a schema programmatically from some other representation of the data, and where empty-enum (no valid value) would be a natural thing to generate for some edge case

text/0013-schema-enums.md
cdisselkoen: I'm curious about the use case for bools.  I don't see the harm in supporting longs and bools in addition to strings.

cdisselkoen (approved):
I'm in favor of this change as stated.  I'm not sure whether I prefer the syntax in the main proposal or the Alternative 2 syntax; I added some comments for discussion

text/0013-schema-enums.md
```diff
@@ -0,0 +1,97 @@
...
+- cost of migrating existing Cedar applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+## Alternatives
+
+### Alternative 1: Don't support it
+
+
+### Alternative 2: Different syntax
```
mwhicks1, line 77: Unless we have a good reason to not use JSON schema, e.g., it's inadequate for some important use-case, I don't think there's a good reason for an alternative.

mwhicks1 (approved):
I approve so long as we keep JSON Schema-based syntax (i.e., with `enum` rather than the `values` alternative). I wouldn't go for an alternative syntax without demonstration that JSON Schema-based syntax is inadequate for an important use-case.

text/0013-schema-enums.md
aaronjeline: Added use cases

shaobo-he-aws (approved):

text/0013-schema-enums.md
```diff
@@ -0,0 +1,145 @@
...
+The presence of the `enum` field has the validator interpret the specification as a union between singleton types for everything in the `enum` list.
+
+We can extend the validator to support singleton types for strings and numbers, and to allow union types for strings, booleans, and numbers. We have singleton types for booleans already, where bool is equivalent to the union true|false. This feature would generalize the same approach to numbers and strings.
+
+For the future addition of numeric ranges to schemas, you could technically have both enumerations and ranges, but the enumeration can be checked against the range at validation time, and then the range can be ignored.
+
+Like other types, enums are usable in the "common types" mechanism.
+
+
+### Detailed Use Cases
```
john-h-kastner-aws, line 63: Add an example for improved error detection in polices:

```
permit(principal, action, resource) when {
  principal.department == "ales"
};
```

We can type the equality as `False`, which would float up as a validation error. If we're clever we can raise an error sooner if we notice that equality that it will never be true. (Harder than it might seem because that could be `false` for only some query environments.

text/0013-schema-enums.md
```diff
@@ -0,0 +1,145 @@
...
+## Timeline
+
+- Start Date: 2023-06-30
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+Add support for enumerations of `bool`, `string`, and `long` types in schemas, and use this information when doing policy validation and schema-based-parsing.
```
john-h-kastner-aws, line 17: I'm not fully convinced we should add long enums at the same time as we have long ranges PR is progress. We should think a bit more carefully about how they behave when in the language together.

john-h-kastner-aws (approved):

text/0013-schema-enums.md
aaronjeline: I agree this warrants more conversation.

khieta (approved):

text/0013-schema-enums.md
```diff
@@ -0,0 +1,154 @@
...
+## Basic example
+
+Maintaining backward compatibility with JSON schema as much as possible, we can match its enum feature. For example:
+
+```json
+"type": "Record",
+  "attributes": {
+    "color": {
+      "type": "String",
+      "enum" : ["red", "blue", "green"]
```
shaobo-he-aws, line 28: I'd prefer `choices` to `enum` as the field name.

cdisselkoen: That sounds sensible to me.  Or `options`

mwhicks1: We want to follow what JSON schema does. That's `enum`. https://json-schema.org/understanding-json-schema/reference/enum

cdisselkoen: Oh, in that case yes let's do that

emina:
The feature proposed in this RFC will have a negative impact on the analyzability of Cedar. 

The RFC lets users specify attribute types that are ad-hoc subtypes of existing (primitive) types.  There is no effective way to encode these ad-hoc constraints into the multi-sorted first-order logic supported by SMT solvers.  Subtypes aren't supported natively in the logic, and there is no efficient way to express them---specifically, we'd need to use quantifiers to do so.  

If we use quantifiers to enforce subtyping constraints, the resulting encoding falls out of the decidable fragment, and the analysis is likely to become brittle, suffering from timeouts and increased sensitivity to small changes in the encoding.  

If we omit subtyping constraints, then the analysis becomes incomplete:  it can produce spurious counterexamples that violate the subtyping constraints.  

There is a variant of this proposal that is encodable.  The variant would involve creating explicit enumerated types that are distinct from all other types, similarly to how this would be done in language like Rust.  For example, the schema could declare `enum color { red, green, blue }` and we'd have to add syntax that lets policies refer to literals of type `color`, e.g., `color#red`.  

This alternative design is unsatisfying because it won't have some of the nice properties that enums have in statically typed languages (e.g., an implicit ordering on the literals).  This is because Cedar is dynamically typed, and the enum definitions wouldn't be available at runtime. In fact, it would be perfectly fine to write `color#yellow` in a policy, and this wouldn't result in a runtime error.

It is unclear that this alternative design brings sufficient benefit to justify the cost of implementing and supporting it.  

I suggest that we reject this RFC for the time being, and revisit if there is substantial demand for it in the future.

cdisselkoen:
The resolution implied by @emina's message above leaves me unsatisfied.  This RFC represents real user problems -- we've been asked about the possibility for this feature by multiple folks -- and I don't think we should reject it without brainstorming real alternative solutions for those users.

khieta:
Update: we've decided to archive this RFC in favor of other options like #53

aaronjeline:
Closed as being unworkable from a typing perspective. A different enums proposal was accepted: https://github.com/cedar-policy/rfcs/blob/main/text/0053-enum-entities.md

PR #14, opened 2023-07-11
rfc: create a batch version of is_authorized

glb:
*Issue #, if available:* None

*Description of changes:*

Initial experiments with language bindings show that while calling `is_authorized` for single requests is adequately performant, doing a full round-trip many times to satisfy a single higher-level operation incurs significant overhead ("consistently 2 orders of magnitude greater [than the cost of calling the function in Rust]").

As a result, we propose a batch version of `is_authorized` that can accept a list of authorization requests and provides a matching list of responses, eliminating the repeated round-trip overhead.

### DIFF
--- a/text/0014-is-authorized-batch.md
+++ b/text/0014-is-authorized-batch.md
@@ -202,7 +202,7 @@ representation.
 
 2. If this proposal is implemented too conservatively by simply duplicating the
    `is_authorized` function and dropping in some code to iterate over the input
-   requests, having both `is_authorized` and an `is_authorized_batch` APIs could
+   requests, having both `is_authorized` and `is_authorized_batch` APIs could
    result in double the amount of code to maintain.
 
 3. If this proposal is implemented too aggressively by making `is_authorized`
--- a/text/0014-is-authorized-batch.md
+++ b/text/0014-is-authorized-batch.md
@@ -1,0 +1,247 @@
+# Create a batch version of is_authorized
+
+## Related issues and PRs
+
+- Reference Issues:
+  [k9securityio/cedar-py#13](https://github.com/k9securityio/cedar-py/issues/13)
+- Implementation PR(s): <!-- (leave this empty) -->
+
+## Timeline
+
+- Start Date: 2023-07-11
+- Date Entered FCP: <!-- (leave this empty, update when the PR enters FCP) -->
+- Date Accepted: <!-- (leave this empty, update when the PR is merged) -->
+- Date Landed:
+  <!-- (leave this empty, update when the implementation is in a stable
+  release) -->
+
+## Summary
+
+Initial experiments with language bindings show that while calling
+`is_authorized` for single requests is adequately performant, doing a full
+round-trip many times to satisfy a single higher-level operation incurs
+significant overhead ("consistently 2 orders of magnitude greater \[than the
+cost of calling the function in Rust]").
+
+As a result, we propose a batch version of `is_authorized` that can accept a
+list of authorization requests and provides a matching list of responses,
+eliminating the repeated round-trip overhead.
+
+## Basic examples
+
+### All or nothing
+
+In this example, all of the requests in the batch must be allowed in order for
+the action to be taken. If any request is denied, the action is rejected.
+
+A concrete example of this would be a higher-level action that is composed of
+multiple actions, potentially on multiple resources. Here we show a
+`score_photo` composite operation that requires the principal to have both the
+`view` and `score` permissions on a photo in order for it to be successful.
+
+```rust
+const ACTION_VIEW: EntityUid = ...;
+const ACTION_SCORE: EntityUid = ...;
+
+fn score_photo(principal: EntityUid, photo: EntityUid, context: Context, score: u8) {
+    let view_request = Request::new(Some(principal), Some(ACTION_VIEW), Some(photo), context);
+    let score_request = Request::new(Some(principal), Some(ACTION_SCORE), Some(photo), context);
+    let requests = vec![view_request, score_request];
+
+    // not shown: obtaining policy set and entities
+
+    let authorizer = Authorizer::new();
+
+    let responses = authorizer.is_authorized(&requests, policy_set, entities);
+
+    // all requests must be allowed to take action; if any
+    // are denied then reject the action
+    if responses.iter().any(|&x| x.decision() == Decision::Deny) {
+        // reject
+        return
+    }
+
+    // apply score
+}
+```
+
+### Filtering on authorization result
+
+In this example, `list_photos` is tasked with returning a list of photos that
+the principal has access to.
+
+```rust
+const ACTION_LIST: EntityUid = ...;
+
+fn list_photos(principal: EntityUid, context: Context) -> Vec<Photo> {
+    // not shown: obtaining list of photo IDs
+    let photo_ids: Vec<EntityUid> = ...;
+    let requests = vec![];
+
+    for id in photo_ids {
+        requests.push(Request::new(Some(principal), Some(ACTION_LIST), Some(id), context))
+    }
+
+    // not shown: obtaining policy set and entities
+
+    let authorizer = Authorizer::new();
+
+    let responses = authorizer.is_authorized(&requests, policy_set, entities);
+
+    let photos = vec![];
+
+    // we should have a request for each photo ID
+    // (in order) and a response for each request,
+    // so transitively we have a response for each
+    // photo ID.
+    for (photo_id, response) in photo_ids.iter().zip(responses.iter_mut()) {
+        match response.decision() {
+            Decision::Allow => {
+                let photo = ...;
+                photos.push(photo)
+            },
+            Decision::Deny => {
+                // skip
+            }
+        }
+    };
+
+    photos
+}
+```
+
+## Motivation
+
+At time of writing at least three language bindings exist
+([ref](https://github.com/cedar-policy/cedar-awesome#language-and-platform-integrations))
+that provide mechanisms to call the Cedar Authorizer's
+[`is_authorized` method](https://github.com/cedar-policy/cedar/blob/d9201161d8ce45e015c8bcb1cd4191c11039709a/cedar-policy/src/api.rs#L300).
+
+While the majority of requests likely operate on a single resource and can be
+handled efficiently by the current implementation, some requests (especially
+`list` requests) may need to check authorization against multiple resources, and
+it is inefficient to do a full round-trip for every request. Similarly, a
+higher-level operation may involve multiple lower-level actions on multiple
+resources.
+
+In naive comparison testing with the
+[k9securityio/cedar-py](https://github.com/k9securityio/cedar-py)
+`is_authorized` implementation against a
+[very naive `is_authorized_batch` implementation](https://github.com/glb/cedar-py/commit/98bf3e1fa2e90bb6f7326cbd846a3293fdbd960b),
+I observed speedups of at least 5x for even simple cases, as documented in
+[k9securityio/cedar-py#13](https://github.com/k9securityio/cedar-py/issues/13).
+
+@skuenzli also
+[reported](https://github.com/k9securityio/cedar-py/issues/13#issuecomment-1629790294)
+that:
+
+> The time observed in Python is consistently 2 orders of magnitude greater than
+> the authorizer.is_authorized call. These particular requests use a simple
+> policy, a simple schema, and ~170 entities.
+
+The initial thought was to simply create an `is_authorized_batch` API for the
+`cedar-py` Python binding, but in discussion in the Cedar Policy Language Slack
+there was support from @aaronjeline and @andrewmwells-amazon for exposing this
+API in Cedar core so that it could be picked up more easily by other language
+bindings, as they expect those bindings to have similar needs and performance
+concerns.
+
+## Detailed design
+
+### Accept a batch of requests with the current shape
+
+This approach creates a new `is_authorized_batch` function, similar to the
+`is_authorized` function, that accepts a list of `Request` structs
+([ref](https://github.com/cedar-policy/cedar/blob/d9201161d8ce45e015c8bcb1cd4191c11039709a/cedar-policy/src/api.rs#L2006))
+and returns a list of `Response` structs
+([ref](https://github.com/cedar-policy/cedar/blob/d9201161d8ce45e015c8bcb1cd4191c11039709a/cedar-policy/src/api.rs#L338)),
+which might have an implementation something like this:
+
+```rust
+impl Authorizer {
+    /// Returns a vector of authorization responses for the input vector of requests
+    /// with respect to the given `PolicySet` and `Entities`.
+    ///
+    /// The order of responses matches the order of requests.
+    pub fn is_authorized_batch(
+        &self,
+        requests: &Vec<Request>,
+        policy_set: &PolicySet,
+        entities: &Entities,
+    ) -> Vec<Response> {
+        let mut responses: Vec<Response> = vec![];
+
+        for request in requests {
+            let response = self.is_authorized(request, policy_set, entities);
+            responses.push(response);
+        }
+
+        responses
+    }
+}
+```
+
+In the special case where the input `requests` list is empty, the function
+should be a no-op and return an empty list.
+
+An advantage of this alternative is that it gives full flexibility to the API
+consumer in how they structure and batch requests.
+
+A disadvantage of this alternative is that for the initial motivating case of a
+`list` operation on a set of resources the principal, action, and context will
+be duplicated across the list of requests. However, this is not necessarily a
+fatal disadvantage, as a language binding author could write optimized versions
+of the binding that translates the special cases on the Rust side of the binding
+and generates the list of `Request` objects from a more-optimized intermediate
+representation.
+
+## Drawbacks
+
+1. Expanding the API surface can be confusing to consumers, as they may not know
+   which of the `is_authorized_*` methods best meets their needs.
+
+2. If this proposal is implemented too conservatively by simply duplicating the
+   `is_authorized` function and dropping in some code to iterate over the input
+   requests, having both `is_authorized` and an `is_authorized_batch` APIs could
+   result in double the amount of code to maintain.
+
+3. If this proposal is implemented too aggressively by making `is_authorized`
+   accept a batch of requests and not maintaining the existing function
+   signature, existing Cedar applications would need to be updated.
+
+## Alternatives
+
+### Do nothing
+
+It is entirely acceptable to do nothing in the Cedar core to accommodate this
+request. The published `is_authorized` API is sufficient, and language binding
+authors can independently create batch implementations to optimize for their
+needs as they discover them.
+
+A disadvantage of encouraging independent evolution is that bindings may
+diverge, causing barriers for developers who may want or need to use multiple
+language bindings in their work.
+
+### Accept a modified request that has multiple resources but uses the same principal, action, and context
+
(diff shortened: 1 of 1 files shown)


text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+concerns.
+
+## Detailed design
+
+### Accept a batch of requests with the current shape
+
+This approach creates a new `is_authorized_batch` function, similar to the
+`is_authorized` function, that accepts a list of `Request` structs
+([ref](https://github.com/cedar-policy/cedar/blob/d9201161d8ce45e015c8bcb1cd4191c11039709a/cedar-policy/src/api.rs#L2006))
+and returns a list of `Response` structs
```
aaronjeline, line 156: We'd want some way to associate which `Response` corresponds to which request right? We don't need that for the first example, but for listing queries we do.

glb: My original version of the example had

```rust
    for (request, response) in requests.iter().zip(responses.iter_mut()) {
```

but I realized that actually the requests weren't really what I wanted to correlate.

text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+> The time observed in Python is consistently 2 orders of magnitude greater than
+> the authorizer.is_authorized call. These particular requests use a simple
+> policy, a simple schema, and ~170 entities.
+
+The initial thought was to simply create an `is_authorized_batch` API for the
+`cedar-py` Python binding, but in discussion in the Cedar Policy Language Slack
+there was support from @aaronjeline and @andrewmwells-amazon for exposing this
+API in Cedar core so that it could be picked up more easily by other language
+bindings, as they expect those bindings to have similar needs and performance
+concerns.
```
andrewmwells-amazon, line 147: If you don't mind me expanding your RFC, I can also include some details about the inefficiencies of constructing an `Evaluator` per `is_authorized` call.

text/0014-is-authorized-batch.md
aaronjeline: +1 to andrew, I think we need some correlation but pairs seem good enough

glb: OK; not familiar with the syntax for this in Rust yet, will see what I can learn from `zip`'s implementation. I assume it would still be possible to zip with another iterable so that a person could do what I've shown in the list_photos example with correlating `photo_id` and `response`?

text/0014-is-authorized-batch.md
andrewmwells-amazon: >Is there a mechanism that a person could use to lazy-load entities as they're needed, or is that just a whole lot of headaches that should be avoided?

Not currently, but maybe something like https://github.com/cedar-policy/rfcs/pull/16 would work?

text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+In the special case where the input `requests` list is empty, the function
+should be a no-op and return an empty list.
+
+An advantage of this alternative is that it gives full flexibility to the API
+consumer in how they structure and batch requests.
+
+A disadvantage of this alternative is that for the initial motivating case of a
+`list` operation on a set of resources the principal, action, and context will
```
Thor-Bjorgvinsson, line 191: What is the rational of having `context` part of the request instead of having a single `context` in batch_is_authorized?

text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+significant overhead ("consistently 2 orders of magnitude greater \[than the
+cost of calling the function in Rust]").
+
+As a result, we propose a batch version of `is_authorized` that can accept a
+list of authorization requests and provides a matching list of responses,
+eliminating the repeated round-trip overhead.
+
+## Basic examples
+
+### All or nothing
```
cdisselkoen, line 32: This use case (composite operations) seems different enough from the others (batch or list APIs) that I think it should probably be a different API.  This one returns a single authz result, while the others are asking for N authz results.

text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+2. If this proposal is implemented too conservatively by simply duplicating the
+   `is_authorized` function and dropping in some code to iterate over the input
+   requests, having both `is_authorized` and `is_authorized_batch` APIs could
+   result in double the amount of code to maintain.
+
+3. If this proposal is implemented too aggressively by making `is_authorized`
+   accept a batch of requests and not maintaining the existing function
+   signature, existing Cedar applications would need to be updated.
+
+## Alternatives
```
cdisselkoen, line 212: I'd like to propose an additional alternative for discussion.

One drawback with the RFC's proposal as written is that it has to make a single choice as to which parts of the batch request remain constant across all requests, and which parts vary per request.
- Only the resource varies, and principal/action/context/entities/policies are constant?  This seems likely for the filtering-photos use case
- Principal-resource pairs, with action/context/entities/policies constant?
- Principal-action-resource-context tuples, with entities/policies constant?  This might be the most natural interpretation of "batch API"
- Etc

The batch API would be inherently inflexible; it has to basically pick one of these choices, and be suboptimal for the others.

Instead, the various language bindings could expose what I'd call a multi-step interface for `is_authorized()`, similar to what's possible today for Rust users of `is_authorized()`.  One FFI call would construct a Rust `PolicySet` object and return some kind of opaque reference to it.  Another call would construct a Rust `Entities` object and return an opaque reference.  Then the bindings' `is_authorized()` would take these preconstructed opaque references representing policies and entities, rather than re-parsing policies and entities with every call.

The penalty for one-off `is_authorized()` calls is small -- just two extra trips across the FFI boundary, which I imagine will be dwarfed by the actual parsing time for policies and entities in most cases.  The benefit we get is that callers can reuse parsed policies and entities across arbitrary calls.  This API doesn't have to choose between the above options but can easily accommodate them all, as callers choose which arguments remain constant and which vary.  It also avoids the concerns in one of the above threads about the right format to return 100 API results, as in this alternative, each call just returns one result.

text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+It is entirely acceptable to do nothing in the Cedar core to accommodate this
+request. The published `is_authorized` API is sufficient, and language binding
+authors can independently create batch implementations to optimize for their
+needs as they discover them.
+
+A disadvantage of encouraging independent evolution is that bindings may
+diverge, causing barriers for developers who may want or need to use multiple
+language bindings in their work.
+
+### Accept a modified request that has multiple resources but uses the same principal, action, and context
```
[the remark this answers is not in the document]
glb: Thanks @andrewmwells-amazon ; it feels like this comment should trigger a note in the `Detailed design` section highlighting that the trivial example implementation given is likely not the most performant option and that further improvements could be made. Make sense?

text/0014-is-authorized-batch.md
glb: @cdisselkoen thanks for the comment. I wrote this example based on my interpretation of a scenario that @skuenzli sketched out for taking multiple requests that was more general than simply multiple resources which I may have misinterpreted. Hoping he can comment with either a better example or more clarity.

Personally I wouldn't necessarily feel a need for a separate core API when the implementation is effectively a `.any` as shown here, but I wouldn't object to one either.

text/0014-is-authorized-batch.md
glb: @Thor-Bjorgvinsson great question; I was trying to accommodate the general use case and didn't gather enough information about the motivation. My original proposal was the one shown in the alternatives where everything except the resources was static across all requests, but initial feedback was that it would be better to support a full request, and accepting a `Vec<Request>` was right there and relatively straightforward as a parallel construction to `is_authorized`'s signature. I should have asked more questions about what needed to vary to support their use case.

I expect that `principal` and `context` are likely to be constant across the set of requests in any given batch, but I don't have data or adequate input. I'll try to get some.

text/0014-is-authorized-batch.md
glb: @cdisselkoen This is an interesting alternative. I'd be curious about lifecycle management for these opaque references; how would the Rust objects that are referenced be freed once the references are no longer used on the other side of the boundary? I'm going to guess that there's a standard pattern for this?

There's a task to gather more metrics to help us understand where time is being spent in the Python binding (https://github.com/k9securityio/cedar-py/issues/14); hopefully this data will help guide us in resolving your comment "The penalty for one-off is_authorized() calls is small -- just two extra trips across the FFI boundary, which I imagine will be dwarfed by the actual parsing time for policies and entities in most cases."

Does this comment match up with @andrewmwells-amazon 's [comment about only constructing the Evaluator once](https://github.com/cedar-policy/rfcs/pull/14#discussion_r1259943823) or does it expand on it?

cdisselkoen: I'd say the two comments are orthogonal.  Constructing the evaluator is different from parsing policies or entities.

text/0014-is-authorized-batch.md
skuenzli: I'm interested in an API that accepts:

* multiple requests
* single policy set, schema, and entities

and returns a result for each request.

We populate the context of the request with information relevant to the principal and resource that we write policies against, e.g. `context.principal.name` and `context.resource.name`.

so perhaps something like:

`is_authorized(&self, reqs: Vec<&Request>, p: &PolicySet, e: &Entities) -> Vec<Response>`

It would be the responsibility of the caller to scope the policy set, schema, and entities appropriately to cover the requests.

For my own use cases, I'm interested in batching:
1. multiple actions with a single principal and resource
2. (perhaps) multiple actions and multiple resources with a single principal

The multi-action+multi-resource batching would be beneficial for a common, expensive use case we have, though I haven't done the work to determine the marginal benefit between 1 & 2.  And there are definitely tradeoffs.  In particular, error handling and communication back to the caller may become significantly more complicated if there is a fan-in of hundreds/thousands/more principal-action-resource combinations (true for 1 and more so for 2).  

Everything has limits.

So if Cedar adds a batch-oriented API, I think it's acceptable and wise to enforce some sort of limit, e.g. 1000 p-a-r combos.

text/0014-is-authorized-batch.md
skuenzli: I agree that we'd need to correlate request and response.  I think it should be the caller's responsibility to construct a single policy set, schema, and entities to support the requests.

re correlation options:

* correlating pairs works, but makes me nervous without some direct correlation mechanism even with JSON guaranteeing preservation of Array order, which is relevant for many/most binding libs
* perhaps request should support an optional `correlationId` property that is returned in the authz result
* I'd also be ok providing a map of requests keyed by a correlation id; I'm guessing I'll end up building that structure anyway so I can unpack the batch of results because authz result does not contain a copy of the request, only decision and diagnostics

text/0014-is-authorized-batch.md
glb: Thanks @skuenzli!

> We populate the context of the request with information relevant to the principal and resource that we write policies against, e.g. context.principal.name and context.resource.name.

It sounds like it's critical for your use case to have context vary per request, is that correct? There's a [comment](https://github.com/cedar-policy/rfcs/pull/14#discussion_r1261254029) that asks why context isn't constant; seems like this is a reason to keep it as part of each request in the batch.

> So if Cedar adds a batch-oriented API, I think it's acceptable and wise to enforce some sort of limit, e.g. 1000 p-a-r combos.

Could you say more about why you think the batch size limit should be applied at the Cedar level? It seems like the caller could set its own limits on batch size depending on the performance characteristics and risks that it's willing to accept, and Cedar could stay out of it the same way it doesn't put limits on number of entities or policy size.

skuenzli: Yes, our context varies by request with both principal and (usually) resource data for that request.

re limits:
Yeah, it might not make sense for the Cedar Authorizer to enforce limits directly.  Perhaps that should be the Cedar user / end service owner's responsibility.

But I'm sure Cedar is designed and built with some constraints and usage patterns in mind.  It's useful to guide users how to use it effectively even if it's just in the library API docs.

text/0014-is-authorized-batch.md
glb: From [this comment](https://github.com/cedar-policy/rfcs/pull/14#discussion_r1261827416) it looks like `context` could vary across the request set. I like the idea that an authorization request is the same shape whether it is being used in a single `is_authorized` call or it's being sent in a batch to `is_authorized_batch`.

text/0014-is-authorized-batch.md
glb: What I'm hearing is that folks are looking at:

```rust
pub fn is_authorized_batch(
    &self,
    requests: &Vec<Request>,
    policy_set: &PolicySet,
    entities: &Entities,
) -> Vec<Response> { ... }   
```

and thinking "it would be helpful to explicitly return the request with the response it created", something more like:

```rust
    pub fn is_authorized_batch(
        &self,
        requests: &Vec<Request>,
        policy_set: &PolicySet,
        entities: &Entities,
    ) -> Vec<(Request, Response)> { ... }
```

is that accurate? I'm also hearing an alternative suggest to extend `Request` to include a `correlationId` that would be copied to & returned back in the `Response`; not certain if that is seen as sufficient or whether there's a desire to have both the `correlationId` and the `Vec<(Request, Response)>` return in the final solution.

I'm contemplating the transport back over to the other side of the FFI boundary, which I know `cedar-py` uses JSON for, and the implied re-marshalling of the requests when sending them back alongside the responses. We don't have instrumentation results yet, but from the commentary and gut feel it seems like the JSON marshalling may be a material contributor to the latency and we might want to minimize the amount of data we're marshalling and passing across the boundary.

I'm not sure I understand the rationale for copying the request into the return value, as `responses[n]` is defined as the response for `requests[n]`, the caller will have `requests` already, and if they want to explicitly iterate through them together they can do so with `for (request, response) in requests.iter().zip(responses.iter_mut()) { ... }` as I outlined above and in the listing example.

Now that folks have had a couple of days for the comments to percolate, any new thoughts? Have I misunderstood the comments?

text/0014-is-authorized-batch.md
```diff
@@ -0,0 +1,247 @@
...
+
+        for request in requests {
+            let response = self.is_authorized(request, policy_set, entities);
+            responses.push(response);
+        }
+
+        responses
+    }
+}
+```
```
andrewmwells-amazon, line 182: Copying conversation from https://github.com/cedar-policy/rfcs/pull/14#discussion_r1261468425 here.
I suggest we implement the API you propose such that the Evaluator is only constructed once (which can be substantially more efficient if there are many complex entities). I think we'll have to do something like that inevitably, so I don't find `drawback 1` very concerning.

glb: Thanks @andrewmwells-amazon ... I've done a bit of poking at the `Evaluator` and it looks like the majority of the work done in constructing an `Evaluator` is in constructing the `EntityAttrValues`:

https://github.com/cedar-policy/cedar/blob/55c786a9a9f48c309e3373d4ec253f566fcc9f52/cedar-policy-core/src/evaluator.rs#L197-L224

(the `RestrictedEvaluator` created in line 214 seems to just store a reference to the extensions)

https://github.com/cedar-policy/cedar/blob/55c786a9a9f48c309e3373d4ec253f566fcc9f52/cedar-policy-core/src/evaluator.rs#L79-L81

but I'm not sure how much work is done by the call to `partial_interpret` on the context.

I'm wondering if it would make sense to extract the `EntityAttrValues::new` call a level (or two) so that the `EntityAttrValues` is re-used by `is_authorized_batch` but otherwise a new `Evaluator` is still constructed for each request.

Thoughts?

text/0014-is-authorized-batch.md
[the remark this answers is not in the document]
glb: Conversation has moved [here](https://github.com/cedar-policy/rfcs/pull/14/files/c8dbb14300e610f486f595b333741af30b0e141b#r1264245725).

text/0014-is-authorized-batch.md
andrewmwells-amazon: Yeah, I think that makes sense. If `is_authorized_batch` takes in an Evaluator or an `EntityAttrValues` then a user with a large entity store and a stream of non-batched requests could call `is_authorized_batch` with a batch size of 1 and not have to pay for repeatedly constructing an `Evaluator` (where building the `EntityAttrValues` is the expensive part).

glb: @andrewmwells-amazon streaming is an interesting take. I was thinking that `is_authorized_batch` would construct the `EntityAttrValues` once and then iterate over the requests in the batch and assuming that the provided `Entities` was either a subset of the full store or a dynamic store like in #16 . To me the batch version fits pretty well with the motivating use cases that I had in mind. Your take seems to fit well with @cdisselkoen 's [proposed alternative](https://github.com/cedar-policy/rfcs/pull/14#discussion_r1261277984), and brings back the question I had about lifetimes: who would own the `EntityAttrValues` and make sure that it was cleaned up when it was no longer needed? (recalling that another of the motivating factors here was improving performance for batches of requests passed from a binding written in another language). How does that sort of thing usually get handled in Rust?

text/0014-is-authorized-batch.md
cdisselkoen: For inefficiencies of constructing an `Evaluator` per `is_authorized` call, see https://github.com/cedar-policy/cedar/issues/227, which is a change we're considering independent of this RFC (and might reduce the motivation for this RFC a little)

text/0014-is-authorized-batch.md
cdisselkoen: I agree we want to evaluate the attr values once and reuse across many calls to `is_authorized`.  See https://github.com/cedar-policy/cedar/issues/227.  I believe the solution we'll go for there is that the `Entities` object that comes into the main `is_authorized` API will under-the-hood carry evaluated attr values already -- this means that the work of evaluating the values moves from `is_authorized` to the construction of the `Entities` itself, and thus can be amortized across many `is_authorized` calls.  Given that, the question merely becomes about the lifetime of the `Entities` object.

I'm not sure how lifetimes are normally handled across FFI boundaries, but one option could be to use `Rc` (ref-counted pointer) on the Rust side, increment the refcount when returning the object to the foreign language, and provide a new API call that the foreign language could use to dispose of the object (on the Rust side, decrementing its refcount).  Concretely, in the Python case, the "opaque reference" returned to Python would be something like an `Rc<Entities>` where we manually increase the refcount by one as part of the FFI interface; and there would be a Python function `dispose()` or something, where calling `dispose()` on that opaque reference would call back to Rust where we'd manually decrement the refcount.  But again, I don't know what the typical patterns are for lifetimes across FFI calls.

glb: Thanks @cdisselkoen . It seems to me that changing the behaviour so that the work [currently being done to construct](https://github.com/cedar-policy/cedar/blob/55c786a9a9f48c309e3373d4ec253f566fcc9f52/cedar-policy-core/src/evaluator.rs#L203) `EntityAttrValues` could be done earlier (or lazily) would indeed reduce the motivation for an `is_authorized_batch` in the Cedar API level. Language bindings could still choose between the opaque reference solution and a batch operation depending on the behaviour that their users needed. For example, the use cases that I'm contemplating would not particularly benefit from a long-lived reference, but others might definitely want such a thing.

aaronjeline:
Seems like we diverged in the solutions we wanted here. Ok to close this? We'll re-submit a new CR with our solution.

PR #18, opened 2023-07-11
Generalized `in` operator

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/generalized-in/text/0018-generalized-in.md)

### DIFF
--- a/text/0018-generalized-in.md
+++ b/text/0018-generalized-in.md
@@ -1,0 +1,184 @@
+# Generalized `in` operator
+
+## Related issues and PRs
+
+- Reference Issues:
+- Implementation PR(s):
+
+## Timeline
+
+- Start Date: 2023-07-11
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+Allow `in` to be used for set membership, not just hierarchy membership. In particular, allow it to be used with non-entity-typed values on the LHS. The detailed design of this proposal (see below) ensures that back-compatibility is maintained for all existing valid Cedar policies.
+
+## Basic examples
+
+### Example 1
+Today's Cedar:
+```
+permit(principal, action == Action::"view", resource == Document::"mine.doc")
+when {
+    ["sales", "legal"].contains(principal.department)
+};
+```
+
+Proposed:
+```
+permit(principal, action == Action::"view", resource == Document::"mine.doc")
+when {
+    principal.department in ["sales", "legal"]
+};
+```
+
+### Example 2
+Today's Cedar:
+```
+permit(
+    principal == Principal::"long-uuid-here",
+    action == Action::"view",
+    resource in Database::"prod"
+) when {
+    resource has tags
+        && resource.tags has tagA
+        && ["value1", "value2"].contains(resource.tags.tagA)
+};
+```
+
+Proposed:
+```
+permit(
+    principal == Principal::"long-uuid-here",
+    action == Action::"view",
+    resource in Database::"prod"
+) when {
+    resource has tags
+        && resource.tags has tagA
+        && resource.tags.tagA in ["value1", "value2"]
+};
+```
+
+## Motivation
+
+In today's Cedar, `.contains()` is the only way to check set membership. There are two problems with this:
+1. This is simply awkward in some use cases, such as the two examples above. Subjectively, the `.contains()` syntax is hard to read in these cases, because the set is required to be on the LHS and the value you're actually interested in is required to be on the RHS. In these examples, having the set on the RHS instead of the LHS reads more intuitively.
+2. Anecdotally, many users intuitively expect that `in` can be used for set membership and not just hierarchy membership, and are confused when this results in Cedar type errors.
+
+## Detailed design
+
+Cedar's existing `in` keyword is extended to also work for set membership, in a way that is back-compatible with the current semantics. Recall the current semantics of `in`:
+
+* `principal.manager in Group::"A"`
+
+This is true if `principal.manager` is a member of `Group::"A"`.
+
+* `principal.manager in [Group::"A", Group::"B"]`
+
+This is true if `principal.manager` is a member of either group. Equivalent to `principal.manager in Group::"A" || principal.manager in Group::"B"`.
+
+This has been referred to as the "deep" semantics of `in` -- the `x in [A, B]` form doesn't just check if `x` is equal to `A` or `B`, but (recursively) if `x` is a member of `A` or `B` in the hierarchy.
+
+In contrast, set membership is inherently shallow. If `in` was naively adapted to support set membership, the above example would be ambiguous: did the policy author mean set membership (in which case they're intending `principal.manager == Group::"A" || principal.manager == Group::"B"`), or today's Cedar's deep hierarchy membership (with the semantics above)?
+
+This RFC proposes to combine these two semantics in a back-compatible way.
+Informally, `in` is "deep" for entity elements, and shallow for non-entity elements.
+More formally, the semantics of `a in B` is:
+* if `a` and `B` are both entity references, then `true` if `a` is either equal to `B`, or a descendant of `B` in the hierarchy, else `false`
+    * this is not changed from today's Cedar
+* if `B` is a Cedar set and `a` is not an entity reference, then it's ordinary shallow set membership, i.e., equivalent to `B.contains(a)`
+    * this case is a type error in today's Cedar
+* if `B` is a Cedar set and `a` is an entity reference, then WLOG let `B` be `[B1, B2, B3]`. The semantics of `a in [B1, B2, B3]` is `a op B1 || a op B2 || a op B3` where for each `Bx`, if `Bx` is an entity reference then `op` is `in`, otherwise `op` is `==`
+    * in the case where `B` only contains entity references, this behavior matches today's Cedar
+    * in the case where `B` contains other values, this would be a type error in today's Cedar
+* in all other cases, type error
+    * this is not changed from today's Cedar
+
+Here's one additional example (we'll call it Example 3):
+```
+resource.owner in ["foo", "bar", User::"Group"]
+```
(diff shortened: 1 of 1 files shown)


text/0018-generalized-in.md
```diff
@@ -0,0 +1,184 @@
...
+    - Response: Is there a real-world use case that needs shallow set membership with entity references?
+- The new `in` is nearly redundant with `.contains()`. For ordinary set-membership with non-entities (as in both of the basic examples at the top), users would now have two different ways to do the same thing -- something we have previously said we want to generally avoid if possible.
+    - Response: It seems valuable to have two ways to do set-membership, one with the set on the LHS and the other with the set on the RHS. Cedar already supports "flipped" versions of some operators, which are redundant except for switching their operands -- consider `<` and `>`.
+- A final objection is simply that one of the Alternatives below might be better. See discussion below.
+
+## Alternatives
+
+### Alternative 1: Do nothing (status quo)
+
+The Motivation section above explains why this is undesirable.
```
khieta, line 124: I think doing nothing is the wrong choice, but we could still do something simpler that what's proposed by the RFC to reduce the pain. In particular: we could add a special error message for when someone tries to use `in` in place of `contains`. In the [slack thread](https://cedar-policy.slack.com/archives/C0547KH7R19/p1689253188448809) linked above, the error message reported was:
```
Errors: while evaluating policy policy0, encountered the following error: type error: expected (entity of type any_entity_type), got string
```
It would have been presumably easier to diagnose & fix the issue if we instead reported something to the effect of "type error: `in` expects an entity on the LHS. Use `.contains()` for set membership."

khieta (approved):
Overall, I like this proposal and would support accepting as-is, although I have a slight preference for some of the alternatives. I left a couple comments to start off the discussion.

text/0018-generalized-in.md
cdisselkoen: +100 to this.  I'll open an issue right now, we can at least do this now while the RFC continues to be under discussion

text/0018-generalized-in.md
```diff
@@ -0,0 +1,185 @@
...
+Under this proposal, this expression is `true` if `resource.owner` is the string `"foo"`, the string `"bar"`, the entity reference `User::"Group"`, or any entity reference which is `in` `User::"Group"`.
+
+Arguably, this is the "intuitive" semantics that users want for `in` -- it blends "deep" `in` for entity references while also providing shallow set-membership for sets.
+
+One final note -- this RFC does not propose any changes to what syntax is legal in the policy scope. And of course, it doesn't change the behavior of any syntax that's currently legal in the policy scope.
+
+## Drawbacks
+
+- The new semantics is more complicated than Cedar's current semantics for `in`. This could lead to additional confusion.
+    - Response: The current semantics also leads to confusion, which this proposal at least partially addresses.
```
mwhicks1, line 113: The current semantics leads to confusion that is easily remedied (once you fix the error message): You really wanted `.contains` but you used `in` instead. Oops. Change your expression then you are done.

text/0018-generalized-in.md
```diff
@@ -0,0 +1,185 @@
...
+        && resource.tags has tagA
+        && resource.tags.tagA in ["value1", "value2"]
+};
+```
+
+## Motivation
+
+In today's Cedar, `.contains()` is the only way to check set membership. There are two problems with this:
+1. This is simply awkward in some use cases, such as the two examples above. Subjectively, the `.contains()` syntax is hard to read in these cases, because the set is required to be on the LHS and the value you're actually interested in is required to be on the RHS. In these examples, having the set on the RHS instead of the LHS reads more intuitively.
+2. Anecdotally, many users intuitively expect that `in` can be used for set membership and not just hierarchy membership, and are confused when this results in Cedar type errors. See, for example, [this thread](https://cedar-policy.slack.com/archives/C0547KH7R19/p1689253188448809) on the Cedar Slack.
```
mwhicks1, line 69: Do we have any examples of customers who want the semantics we are actually proposing? The slack thread you mention is confusion about which operator to use: The user wanted exactly what `.contains()` is providing, semantically, but mistakenly used operator `in` instead. A better error message could direct them to the right construct.

I'm dubious of making a serious language change to make it "easier" for people to write code. This sort of choice was made with many scripting languages (I'm looking at you, Javascript), and the results are not always positive. 

Do we have examples where a customer wants `principal.foo in ["a", "b", User::"foo"]` to be `==` for the first two, and deep `in` for the last? If yes, that's more support for this proposal. If no, what would customers think should happen? If they mostly guess right, that's more support. If they don't, then you are possibly introducing a subtle bomb in the semantics.

text/0018-generalized-in.md
```diff
@@ -0,0 +1,185 @@
...
+    * this is not changed from today's Cedar
+
+Here's one additional example (we'll call it Example 3):
+```
+resource.owner in ["foo", "bar", User::"Group"]
+```
+Under this proposal, this expression is `true` if `resource.owner` is the string `"foo"`, the string `"bar"`, the entity reference `User::"Group"`, or any entity reference which is `in` `User::"Group"`.
+
+Arguably, this is the "intuitive" semantics that users want for `in` -- it blends "deep" `in` for entity references while also providing shallow set-membership for sets.
+
```
mwhicks1, line 107: Is "intuitiveness" something you can test? What if we did a Quiz on the Cedar Slack to ask people's expectations for various examples?

text/0018-generalized-in.md
```diff
@@ -0,0 +1,185 @@
...
+};
+```
+
+## Motivation
+
+In today's Cedar, `.contains()` is the only way to check set membership. There are two problems with this:
+1. This is simply awkward in some use cases, such as the two examples above. Subjectively, the `.contains()` syntax is hard to read in these cases, because the set is required to be on the LHS and the value you're actually interested in is required to be on the RHS. In these examples, having the set on the RHS instead of the LHS reads more intuitively.
+2. Anecdotally, many users intuitively expect that `in` can be used for set membership and not just hierarchy membership, and are confused when this results in Cedar type errors. See, for example, [this thread](https://cedar-policy.slack.com/archives/C0547KH7R19/p1689253188448809) on the Cedar Slack.
+
+## Detailed design
```
mwhicks1, line 71: What is the impact of this change on validation?

text/0018-generalized-in.md
cdisselkoen: The reason for this proposal instead of a simpler semantics like Alternative 5, is that it can be done backwards-compatibly, while Alternative 5 is a breaking change.  If we think the breaking change is acceptable for Cedar 3, we can consider Alternative 5 and, as @khieta also says on a different thread here, the semantics would be simpler and more intuitive.  However we haven't so far greenlit any breaking changes to the Cedar policy syntax itself (at least to my memory), and I'm hesitant to do that, even in a "new major version" like Cedar 3.  This particular breaking change would be particularly insidious because it wouldn't be that previously-valid policies now produce parse errors, evaluation errors, or even validation errors, but rather that the behavior simply silently changes out from under you when you upgrade to Cedar 3.

emina: I think Alternative 5 is too invasive at this point, even with a major version change.  This is a fundamental change to the semantics of a core operator.

I think we should make the error message better / more informative, and then wait to see if a change to semantics is needed, and if so, which alternative meets most users’ expectations.

cdisselkoen: re: breaking changes in Cedar policy syntax for Cedar 3 -- I've now (somewhat hypocritically) proposed one in [RFC 20](https://github.com/cedar-policy/rfcs/pull/20), but I believe there are significant mitigating factors in that case that are not present with Alternative 5.  In RFC 20's case, the break is that previously valid Cedar policies would now produce errors, while in Alternative 5, previously valid Cedar policies would now silently have a different non-error behavior, which in my opinion is worse.  Also, I suspect the number of users who have Cedar policies in the class that RFC 20 would break, is significantly smaller than the number of users who have Cedar policies in the class that Alternative 5 would break.

text/0018-generalized-in.md
cdisselkoen: Good point.  And for those tracking, the error message fix is cedar-policy/cedar#177

cdisselkoen: Updated the text here (and under Alternative 1) in light of cedar#177

text/0018-generalized-in.md
```diff
@@ -0,0 +1,184 @@
...
+
+In this alternative, users wouldn't have any way to get the current "deep" semantics. Some ways to address that:
+* We could add a new keyword that has the current "deep" semantics -- say, `inAny`
+* If we eventually introduce a more general feature to map operators over sets, users could get the "deep" semantics by mapping `in` over a set
+* If the RHS is a literal, users could just write out `a in G1 || a in G2 || ...`. This doesn't help for the case where the RHS is not a literal, e.g. today's `User::"Henry" in resource.allowedGroups`
+
+Reasons not to take this alternative:
+* It would be a breaking change for some existing Cedar policies
+* If we add a new keyword like `inAny`, the distinction between `in` and `inAny` could cause confusion
+* At least one user has opined that it's unintuitive that `John in [SalesTeam::"California", SalesTeam::"Oregon"]` would have different semantics than `John in SalesTeam::"WestCoast"` (assuming the natural group memberships there)
```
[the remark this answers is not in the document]
andrewmwells-amazon: IMO this change is a bad idea without alternative 5. Alternative 5 is a large enough breaking change that I oppose this RFC.

text/0018-generalized-in.md
cdisselkoen: I've updated the discussion of Alternative 5 to note the concerns about backward compatibility mentioned in this thread.  Since this thread (5 months ago), we have also significantly improved the error message, and I've adjusted the RFC text to note this as well.  As a result, I will resolve this thread.

cdisselkoen:
I've updated the RFC to note its current status, which we're terming "archived" -- it doesn't currently have much support, but also we're open to resuming the discussion on it at any time. Feel free to leave a comment and resume the discussion. But in absence of any activity or popular demand, the Cedar team does not plan to take any action on this RFC in the near future.

PR #19, opened 2023-07-11
Stricter, but more explainable, validation

mwhicks1:
[Rendered](https://github.com/cedar-policy/rfcs/blob/stricter-validation/text/0019-stricter-validation.md)

This is a proposal to amend strict mode validation to be simpler, and easier to explain.


text/0019-stricter-validation.md
```diff
@@ -0,0 +1,77 @@
...
+2. When checking in strict mode, we construct a new AST that performs the needed transformations, as in step 2 today. 
+3. Either way, we can add type annotations. For strict mode, they could be added during the transformation; for permissive mode, no transformation is needed. No checking is done after annotations are added, since the checking is always done, regardless of mode, prior to the transformation.
+
+Permissive mode validation considers different `Action`s to have distinct types, e.g., `Action::"view"` and `Action::"update"` don't have the same type. This is in anticipation supporting distinct actions having distinct attribute sets. Today, the least upper bound of any two actions is the type _AnyEntity_, which is a general supertype of all entities, used internally. Such a type supports expressions that are found in the policy scope such as `action in [Action::"view", Action::"edit"]`. This expression typechecks because the least upper bound of `Action::"view"` and `Action::"edit"` is _AnyEntity_ so the expression is type `Set<`_AnyEntity_`>`. Therefore in our amended strict mode we either need to:
```
john-h-kastner-aws, line 61: The `action in [Action::"a", Action::"b"]` case discussed here does show up as a few failing test cases. Treating `AnyEntity` as part of "depth subtyping" seems wrong - it definitely applies width subtyping. I'm in favor of forcing all actions to share one entity type. We might even be able to make this work by allowing depth subtyping between the type of concrete attributes for a particular action and declared attribute types. This way we would still give `Action::"view".is_read_only` type `False` (if it is defined `false`) while being able to support a LUB between `Action::"view"` and `Action::"read"` where their LUB would be `Action` in which `is_read_only : Bool`.

text/0019-stricter-validation.md
```diff
@@ -0,0 +1,77 @@
+# Standalone strict validation
+
```
john-h-kastner-aws, line 2: After hashing out these changes in the Rust I'm convinced that these changes make sense.

There was one cases I encountered where the new definition was stricter in a way I didn't anticipate. If you have an expression like `<strict_error> || true`, the old strict typechecker would see that `|| true` makes the whole expression true, so it got rewritten to `true` without strict typechecking, but now strict typechecking is applied to the LHS before we know RHS is `true`, so strict type errors are reported.

This is another instance where the new behavior is not bad. In this case it's because we used to have permissive type errors reported regardless of the `|| true` (it would be unsound to do otherwise), but strict type error reporting did depend on the right operand. The new behavior is more consistent.

john-h-kastner-aws (approved):

text/0019-stricter-validation.md
mwhicks1: To be clear: You are saying that anticipated future support for attributes of `Action` entities would require (1) defining an `Action` type in the `entityTypes` part of the schema that tabulates the possible attributes and their types, and (2) having individual actions in the `actions` part define values of those attributes, which are checked against the `Action` type. Your comment about depth subtyping is that you might infer a more specific type for individual actions that is a subtype of the general type. Is this correct?

If so, I'm good with this, too.

john-h-kastner-aws: Yes. The goal is that boolean, and later string/long enum, typed action attributes can have a more specific type inferred while still requiring that all the same attributes are defined.

text/0019-stricter-validation.md
```diff
@@ -0,0 +1,77 @@
...
+2. When checking in strict mode, we construct a new AST that performs the needed transformations, as in step 2 today. 
+3. Either way, we can add type annotations. For strict mode, they could be added during the transformation; for permissive mode, no transformation is needed. No checking is done after annotations are added, since the checking is always done, regardless of mode, prior to the transformation.
+
+Permissive mode validation considers different `Action`s to have distinct types, e.g., `Action::"view"` and `Action::"update"` don't have the same type. This is in anticipation supporting distinct actions having distinct attribute sets. Today, the least upper bound of any two actions is the type _AnyEntity_, which is a general supertype of all entities, used internally. Such a type supports expressions that are found in the policy scope such as `action in [Action::"view", Action::"edit"]`. This expression typechecks because the least upper bound of `Action::"view"` and `Action::"edit"` is _AnyEntity_ so the expression is type `Set<`_AnyEntity_`>`. Therefore in our amended strict mode we either need to:
+1. Support _AnyEntity_ as part of depth subtyping.
```
john-h-kastner-aws, line 62: Regarding `AnyEntity`: It's also used to support typechecking template slots. That won't work with this definition of strict typechecking (it doesn't work with the old definition either, but we added special case to allow it). How do we want to handle them now? In a prior RFC we seemed OK with only doing template validation after linking.

mwhicks1: We use the `principalTypes` and `resourceTypes` information from the cross-product. In particular:

We know for a particular action what the principal and resource type must be. The slot is either on the RHS of a `==` or a `in`. If the former, you just type the slot with the principal/resource type. If the latter, you look up the possible `memberOfTypes` and use these, systematically.

emina: +1 for using the cross-product approach to typecheck the templates statically, independently of linking.  This is needed in order to analyze template behavior wrt all possible instantiations.

mwhicks1: Not based on scope conditions, based on the schema. If your policy has 
```
principal == ?principal
```
in the scope, then you can give `?principal` the same type as `principal` used for the cross product. So if the current `principal` type is `User`, typecheck the policy with `?principal:User` too. If instead we have
```
principal in ?principal
```
in the scope, then you can further extend the cross product to type `?principal` with each of the `principal` type's `memberOfTypes`. For example, if `principal` is being typed as `User` and the `memberOfTypes` for `User` in the schema are `Team` and `Region`, then you typecheck the policy first with `principal:User` and `?principal:Team` and then `principal:User` and `?principal:Region`. 

Do likewise for `?resource`. Make sense?

john-h-kastner-aws: The procedure you described seems to be based on the scope conditions.

john-h-kastner-aws: If we let template slots appear in the body of a policy, then this starts to make sense.

emina: This makes sense if you think of actions as (overloaded) functions with signatures.  The suggestion here is to typecheck the template against all possible action signatures.  This isn't circular---it depends just on the schema.

A naive implementation is to iterate through every single action in the schema, and typecheck the template against all of its overloaded signatures.

We can make this more efficient by looking at scope constraints, but this is an optimization, not something inherent to the algorithm that @mwhicks1 proposed.

mwhicks1: @emina described it well. Focusing on the shape of the current policy is an optimization. You could ignore it, but you'd end up with repeated effort.

mwhicks1: Proof is in the pudding of course: We should be able to prove validation soundness of this approach!

andrewmwells-amazon (approved):

emina (approved):

text/0019-stricter-validation.md
john-h-kastner-aws: decisions on this: They must have the same action type. For the purpose of Dafny, this is an action type without any attributes since it doesn't know about action attributes

john-h-kastner-aws:
For an unspecific entity type in an applies to list we will also interpret this as the action applying to all possible entity types.

text/0019-stricter-validation.md
```diff
@@ -0,0 +1,143 @@
...
+It instead applies to the "unspecified" principal or resource, which allows a users to make authorization requests using that action without providing the principal or resource,
+or when providing an entity with any of the defined entity types as the principal or resource.
+This is implemented by assigning the `AnyEntity` type to a variable when it is unspecified,
+but this depends on width subtyping between entity types which we want to eliminate from strict validation.
+
+It is not possible to extend the set of type environments to handle this case in a way that does not use `AnyEntity`.
+Constructing one type environment for each declared entity type does not soundly handle the case where an entity is _not_ provided in the query.
+If some attribute exists for every declared entity type, then an access to that attribute would be permitted, but that attribute cannot be safely accessed if an entity is not used in the query.
+We would need to further extend set of type environments to include an entity representing this case,
+but, other than `AnyEntity`, we do not have a type that can represent the type of an unspecified entity.
```
emina, line 118: What is special about `AnyEntity` type here?  Or, put differently, what is preventing us from constructing a synthetic entity type with no attributes, say `ArbitraryEntity`, and using that instead of `AnyEntity`?  Or, maybe this text is saying that such a solution is equivalent to using `AnyEntity` without subtyping, and we're simply reusing the name `AnyEntity` from permissive validation?

text/0019-stricter-validation.md
```diff
@@ -0,0 +1,143 @@
...
+The `AnyEntity` type is only a problem when it allows for construction of a union type representing multiple entity types.
+We want to typecheck the policy with a distinct type environment for every entity type and for an unspecified entity.
+For each query environment, we can over-approximate by instead using `AnyEntity` as the type of the query variable.
+In each query environment, `AnyEntity` represents exactly one entity type,
+and will always represent exactly one entity type as long as the least upper bound of `AnyEntity` with every type is not defined.
+
+If we approximate every type environment with `AnyEntity`, the different type environments are now identical, so we only need to typecheck once with this approximate type environment.
+Using the `AnyEntity` types will cause the strict-mode validator to reject some expressions which might otherwise be well typed.
+For example, `principal == User::"alice"` is well typed when `principal` is `User`, and for any other entity type for `principal`, including the unspecified entity type, the expression could have type `False`, but we will not accept it.
```
emina, line 130: Why would we not accept an equality that always evaluate to `False`?  The text above suggests that we do allow this and then replace such equalities with the constant expression `False` in the transformation phase of the strict validator.

text/0019-stricter-validation.md
john-h-kastner-aws: `AnyEntity` isn't special. The point is that if strict mode doesn't allow for any subtyping with `AnyEntity`, then `AnyEntity` becomes this arbitrary entity type.

text/0019-stricter-validation.md
john-h-kastner-aws: We could accept it, but this is the limitation of the proposed approach. It lets us use the current typechecker for strict validation without any additional effort, but misses this case.

text/0019-stricter-validation.md
emina: Got it, so we're reusing the name `AnyEntity` but not any of the associated properties from permissive validation.  In that case, maybe we want to give it a different name in the context of strict validation to avoid confusion.  Maybe something like `ArbitraryEntity`.

text/0019-stricter-validation.md
emina: I don't understand why this case is any different than other cases though.  Suppose that the type of `principal` was `Bot`.  Wouldn't we get that `principal == User::"alice"` is `False` in that case?   And if so, what's different when `principal` is `AnyEntity` instead?

john-h-kastner-aws: If we view `AnyEntity` as `AbitraryEntity` as you suggest, then I think this can't get type `False`. The arbitrary entity type could just as easily be `User` as any other entity type.

emina: Is there any disadvantage to the `ArbitraryEntity` approach?  To me, it seems like the simplest solution (both to explain and in terms of the precision it gives).  But I haven't thought through all the consequences :)

emina: Ah, sorry, I misread your "can't get type `False`" as "_can_ get type `False`". 

But why not?  

We have two totally different entity types, and we're comparing instances of those different entity types for equality.  Why is that not `False`?  The fact that we call one of them `ArbitraryEntity` shouldn't matter.

emina: To clarify, an instance of the `ArbitraryEntity` type _cannot_ be equal to an entity of any other type.  It's a distinct type that we call `ArbitraryEntity` that has no attributes.  All instances of this type will be necessarily different from all instances of other types, such as `User`.  Only two instances of `AribtraryEntity` can return  a non-False result when compared for equality.

emina (approved):

andrewmwells-amazon (approved):

aaronjeline (approved):

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

andrewmwells-amazon (approved):

PR #20, opened 2023-07-14
Disallow duplicate keys in Cedar records

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/unique-record-keys/text/0020-unique-record-keys.md)

### DIFF
--- a/text/0020-unique-record-keys.md
+++ b/text/0020-unique-record-keys.md
@@ -57,7 +57,9 @@ The proofs are required to reason about "well-formed" records as a pre- and post
 The status quo also requires reasoning about the order keys appear in a record literal, meaning that the keys can't be naively sorted without preserving information about their original order.
 If duplicate keys were prohibited, the order of keys wouldn't matter (but see "Unresolved questions" below).
 
-A tertiary motivation for this proposal is that it would bring the Cedar language in line with our desired behavior on duplicate keys in other Cedar API surfaces. See [cedar#159](https://github.com/cedar-policy/cedar/issues/159).
+Two secondary, more minor motivations for this proposal are:
+- It would bring the Cedar language in line with our desired behavior on duplicate keys in other Cedar API surfaces. See [cedar#159](https://github.com/cedar-policy/cedar/issues/159).
+- Cedar records resemble other languages' struct types more than they resemble map types. The current behavior would be unexpected for struct types, and the proposed behavior would be intuitive.
 
 ## Detailed design
 
@@ -65,7 +67,11 @@ We'd do something very like [cedar#169](https://github.com/cedar-policy/cedar/pu
 
 In the case of record literals, the error could be caught in the CST->AST step, resulting in the user seeing a parse error rather than an evaluation error.
 In the case of `context`, the error would need to be caught when constructing the `Context` object.
-In other cases, e.g. if evaluating a Cedar subexpression produces a record with duplicate keys, the user would see an evaluation error. (But see "Unresolved questions" below.)
+In the case of records in entity attributes, the error could be caught when constructing an `Entity` object.
+There are no other cases, because currently, Cedar has no other way to construct a record value:
+there are only record literals, and records preexisting in `context` or entity attribute values.
+
+All of these cases would be caught prior to `is_authorized()` in Core, so we would not need a new category of evaluation error. No new evaluation errors would happen due to this proposal.
 
 I don't believe this change would have a significant effect on validation or width/depth subtyping of records, but please correct me if I'm wrong.
 
@@ -75,6 +81,7 @@ This is a breaking change, not just to the Cedar API surface, but to the Cedar l
 That's a significant drawback not to be taken lightly.
 In this case, it's mitigated somewhat by the fact that we're turning previously-valid policies into errors, not changing them into a silently different behavior.
 It's also mitigated by the fact that most Cedar users are probably (citation needed) not writing policies in the class that will have a behavior change -- i.e., policies with duplicate keys in record literals.
+And finally, it's mitigated by the fact that Cedar's current behavior on duplicate keys was not documented in the public docs (e.g., in [this section](https://docs.cedarpolicy.com/syntax-datatypes.html#record)), so if any policies are relying on this behavior, it's probably by accident.
 
 ## Alternatives
 
@@ -86,14 +93,9 @@ I believe this would require a breaking change to the EST format, and record lit
 
 ## Unresolved questions
 
-### Evaluation errors for duplicate keys
-Is it actually possible in today's Cedar, to have duplicate keys in a record value other than a record literal or `context`? I.e., to have duplicate keys in a record value produced by evaluating some subexpression (other than a record literal or the `context` variable)?
-Cedar has no way to augment or union record values, and does not allow computed keys in record literals, so off the top of my head the only way to do this would be via `GetAttr`, which would imply duplicate keys in the entity store, and we can/should disallow that as well.
-If I'm not missing something and there actually is no other way, then we can catch all duplicate-keys errors prior to calling `is_authorized()`, and we don't need a new category of evaluation error.
-
 ### Order-independence for record keys
 At first glance, one would hope that this change would be sufficient to ensure that the order of keys in a record literal (or more generally, any record value) does not matter.
-However, I don't think that's quite true -- if keys are mapped to expressions with errors, the order does matter, because only the first error encountered will be returned.
+However, that's not quite true -- if keys are mapped to expressions with errors, the order does matter, because only the first error encountered will be returned.
 While we're making this change, should we also make a change so that we have true order-independence for Cedar record keys?
 I could think of at least three ways to do that:
 1. Always evaluate all values in a record, and return potentially multiple evaluation errors rather than just the first
--- a/text/0020-unique-record-keys.md
+++ b/text/0020-unique-record-keys.md
@@ -8,7 +8,7 @@
 ## Timeline
 
 - Start Date: 2023-07-14
-- Date Entered FCP:
+- Date Entered FCP: 2023-07-28
 - Date Accepted:
 - Date Landed:
 
@@ -96,6 +96,9 @@ I believe this would require a breaking change to the EST format, and record lit
 ## Unresolved questions
 
 ### Order-independence for record keys
+
+[Resolved: we won't make any further changes as part of this RFC, but may address some of this in a different RFC]
+
 At first glance, one would hope that this change would be sufficient to ensure that the order of keys in a record literal (or more generally, any record value) does not matter.
 However, that's not quite true -- if keys are mapped to expressions with errors, the order does matter, because only the first error encountered will be returned.
 While we're making this change, should we also make a change so that we have true order-independence for Cedar record keys?
--- a/text/0020-unique-record-keys.md
+++ b/text/0020-unique-record-keys.md
@@ -9,7 +9,7 @@
 
 - Start Date: 2023-07-14
 - Date Entered FCP: 2023-07-28
-- Date Accepted:
+- Date Accepted: 2023-08-04
 - Date Landed:
 
 ## Summary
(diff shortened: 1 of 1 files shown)


text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,103 @@
...
+
+We'd do something very like [cedar#169](https://github.com/cedar-policy/cedar/pull/169), except that instead of implicitly dropping all but the last value for each key (in `.collect()`), we'd be careful to emit an error when constructing an `Expr` if the record has duplicate keys.
+
+In the case of record literals, the error could be caught in the CST->AST step, resulting in the user seeing a parse error rather than an evaluation error.
+In the case of `context`, the error would need to be caught when constructing the `Context` object.
+In other cases, e.g. if evaluating a Cedar subexpression produces a record with duplicate keys, the user would see an evaluation error. (But see "Unresolved questions" below.)
+
+I don't believe this change would have a significant effect on validation or width/depth subtyping of records, but please correct me if I'm wrong.
+
+## Drawbacks
```
emina, line 72: I'm supportive of this change as written.  Another mitigating factor is that the current behavior with respect to duplicate keys doesn't seem to be documented.  For example:

* https://docs.cedarpolicy.com/syntax-datatypes.html#record

So, if any policies are relying on this behavior, it's purely by accident.  

Another justification for this change is that records are more closely related to structs than maps in other languages.  With that in mind, allowing duplicate keys during literal construction is counter-intuitive.

text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,103 @@
...
+The main alternative is doing nothing, which of course has the reverse pros/cons of the proposal as explained in "Motivation" and "Drawbacks".
+
+Another alternative, which would address the first motivation and not the second, would be to change the EST format so that it supports record literals with duplicate keys.
+I believe this would require a breaking change to the EST format, and record literals would not be expressible as JSON maps, which is unfortunate because I believe that's the most natural/ergonomic.
+(Devil's advocate: we could maybe find a way to support JSON maps in the common case with some escape syntax for expressing record literals that have duplicate keys. If we want to take this alternative seriously, I can investigate this.)
+
+## Unresolved questions
+
+### Evaluation errors for duplicate keys
+Is it actually possible in today's Cedar, to have duplicate keys in a record value other than a record literal or `context`? I.e., to have duplicate keys in a record value produced by evaluating some subexpression (other than a record literal or the `context` variable)?
```
emina, line 90: This is not possible today because record literals are the only way to construct a record value. All values that are not literals are coming from the input (context or entities).

text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,103 @@
...
+Cedar has no way to augment or union record values, and does not allow computed keys in record literals, so off the top of my head the only way to do this would be via `GetAttr`, which would imply duplicate keys in the entity store, and we can/should disallow that as well.
+If I'm not missing something and there actually is no other way, then we can catch all duplicate-keys errors prior to calling `is_authorized()`, and we don't need a new category of evaluation error.
+
+### Order-independence for record keys
+At first glance, one would hope that this change would be sufficient to ensure that the order of keys in a record literal (or more generally, any record value) does not matter.
+However, I don't think that's quite true -- if keys are mapped to expressions with errors, the order does matter, because only the first error encountered will be returned.
```
emina, line 96: The order of evaluation still matters, even without duplicates.  This is true for both set and record literals, and it's indeed due to error behavior.

text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,103 @@
...
+While we're making this change, should we also make a change so that we have true order-independence for Cedar record keys?
```
emina, line 97: I would lean against making further changes.  The current error behavior is not surprising, and it matches what we do for `&&` and `||`:  expressions are evaluated in order, from left to right, and that's the order in which you get errors.  Also, note that if we change this for record literals, we should also do so for set literals, for consistency.

emina (approved):

text/0020-unique-record-keys.md
shaobo-he-aws: I couldn't think of any significant effects off the top of my head. That being said, it would make the Dafny type checker (maybe also the evaluator?) a little bit slower since iterating a map in Dafny is not trivial (see: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-dafny/def/util.dfy#L64).

text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,103 @@
...
+```
+does not evaluate to `{ foo: 3 }` as you might expect based on the previous example, but rather, it results in an evaluation error.
+
+This RFC proposes that all of the above expressions and policies would result in an evaluation error due to duplicate keys in a record.
+
+## Motivation
+
+There are two main motivations for this change.
+
+First, in the status quo, Cedar policy ASTs actually cannot losslessly roundtrip to EST and back.
```
shaobo-he-aws, line 48: Is it possible to let EST to contain lists instead of maps?

cdisselkoen: Yes; I discuss this under Alternatives

shaobo-he-aws: My bad. Thank you.

text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Detailed design
+
+We'd do something very like [cedar#169](https://github.com/cedar-policy/cedar/pull/169), except that instead of implicitly dropping all but the last value for each key (in `.collect()`), we'd be careful to emit an error when constructing an `Expr` if the record has duplicate keys.
+
+In the case of record literals, the error could be caught in the CST->AST step, resulting in the user seeing a parse error rather than an evaluation error.
+In the case of `context`, the error would need to be caught when constructing the `Context` object.
+In other cases, e.g. if evaluating a Cedar subexpression produces a record with duplicate keys, the user would see an evaluation error. (But see "Unresolved questions" below.)
+
+I don't believe this change would have a significant effect on validation or width/depth subtyping of records, but please correct me if I'm wrong.
```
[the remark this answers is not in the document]
emina: FWIW, for models and proofs, there is no advantage for checking for duplicates in the parser.  To take advantage of this check, the model evaluator would need to carry around complex preconditions that state record literals are free of duplicates.  This is more trouble than checking for duplicates in the evaluator and throwing an error if some are found.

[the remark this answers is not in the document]
cdisselkoen: @emina's comment above implies that actually we should be able to avoid evaluation errors for duplicate keys, which makes this moot.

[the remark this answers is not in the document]
emina: Sorry, I should have been more clear:  it doesn't help to do this check in the parser, at least not for proofs.  It would be harder to take advantage of that check in the evaluator than to just do the check in the evaluator.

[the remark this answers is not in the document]
emina: Thinking about this again:  the RFC needs to specify (a) the evaluation order, and (b) where duplicate key error is thrown and in what order wrt other errors.

To see why, consider this record:  `{ “one”: 1, “one”:  1 - “three” }`  

What error should this throw?

If the duplicate check is done in the parser, then it should be “Duplicate key error.”

If the duplicate check is done in the evaluator, then there are two options: duplicate error or type error.  Which is thrown depends on _when_ we check for duplicates.  Is it before or after evaluating the RHS of a key-value expression?

text/0020-unique-record-keys.md
cdisselkoen: your comment specifically mentions a case (entities) that's not covered here.  Worth investigating duplicate keys in records in entity attributes

text/0020-unique-record-keys.md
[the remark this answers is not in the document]
cdisselkoen: Updated

text/0020-unique-record-keys.md
cdisselkoen: Revised RFC text per this discussion

text/0020-unique-record-keys.md
cdisselkoen: @emina, I added text to the RFC re: your comments.

emina (approved):
I support making this change.

text/0020-unique-record-keys.md
emina: Looks good to me!  Approved.

text/0020-unique-record-keys.md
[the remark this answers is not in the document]
cdisselkoen: > it doesn't help to do this check in the parser ... just perform the check in the evaluator.

(1) Performing the check in the parser is better in the sense that it catches the error earlier, eg, when the user constructs a Policy object, long before they get around to calling `is_authorized()`.
(2) Performing the check in the evaluator fails to address the Motivation 1 for this RFC.
(3) Performing the check in the parser amortizes the check across multiple evaluations of the policy.  Performing the check in the evaluator would have to be repeatedly and redundantly performed every time the policy is evaluated.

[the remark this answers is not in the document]
cdisselkoen: Updated the RFC text to describe error precedence (in Detailed Design)

text/0020-unique-record-keys.md
aaronjeline: This also reflects the semantics we decided on for PE. You're not necessarily guaranteed to get the _same_ error, but you are guaranteed to get _an_ error regardless of ordering.

aaronjeline (approved):

shaobo-he-aws (approved):

text/0020-unique-record-keys.md
```diff
@@ -0,0 +1,107 @@
...
+2. Provide no guarantee about which error the user will receive when multiple record keys have errors (nondeterministic error behavior) -- I think the formal spec currently guarantees the first error?
+3. Guarantee that record values will be evaluated in lexicographic order of their keys, regardless of the order the keys appear in the policy text
+
+This issue is particularly hairy for records contained in attribute values in the entity store -- I believe these records go through a hashmap representation in our current datapath, so we lose ordering information and also duplicate keys (silently). Since the values in these records are restricted expressions, there are very few ways they can produce evaluation errors, but they still can, in the case of calls to extension functions. This is probably a potential source of difference between the errors produced in the Rust implementation and in the formal spec; I wonder why DRT hasn't caught this, particularly in the case where we drop a duplicate key that was mapped to an erroring expression. Maybe the Dafny JSON parser silently takes last-value-wins? or maybe we don't test extension values in attribute values?
```
khieta, line 107: I believe that during DRT we pass in the same post-hashmap entity store & context to both the Dafny and Rust, which means attributes were already silently de-duped in both cases. 

After poking around a bit, it looks like the only place we can generate records with duplicates is when [generating `Expr`s to include in the policy](https://github.com/cedar-policy/cedar-spec/blob/3f3e25b86c4bcd8c67567923dfe57e90e2ceb767/cedar-policy-generators/src/schema.rs#L2972). I believe evaluation will return an error in both the [Rust](https://github.com/cedar-policy/cedar/blob/e940bdf48d750834b96c3279ebd4f96ff8b0eb3d/cedar-policy-core/src/evaluator.rs#L589) and [Dafny](https://github.com/cedar-policy/cedar-spec/blob/3f3e25b86c4bcd8c67567923dfe57e90e2ceb767/cedar-dafny/def/engine.dfy#L138) if any value in a record fails to evaluate (regardless of duplicate keys).

khieta (approved):

anwarmamat (approved):

andrewmwells-amazon (approved):

D-McAdams (approved):

mwhicks1 (approved):

text/0020-unique-record-keys.md
cdisselkoen: Updated the RFC text to reflect that we're not making further changes in this RFC

cdisselkoen:
The final comment period (FCP) for this RFC is starting now, with intent to accept. The FCP will end 2023-08-04 at noon PT / 3pm ET / 7pm UTC. Please add comments, and especially any objections, if you have any. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

hakanson:
Silent handling of duplicate keys in Terraform (HCL language parser) lead to a recent security issue documented at https://securitylabs.datadoghq.com/articles/exploring-github-to-aws-keyless-authentication-flaws/#root-cause-analysis 

Failing fast seems preferable.

aaronjeline (approved):

PR #21, opened 2023-07-15
Add `any` and `all` operators

emina:
[Rendered](https://github.com/cedar-policy/rfcs/blob/any-and-all-operators/text/0021-any-and-all-operators.md)

### DIFF
--- a/text/0021-any-and-all-operators.md
+++ b/text/0021-any-and-all-operators.md
@@ -9,7 +9,7 @@
 
 - Start Date: 2023-07-14
 - Date Entered FCP: 2023-10-27
-- Date Accepted:
+- Date Accepted: 2023-11-08
 - Date Landed:
 
 ## Summary
--- a/text/0021-any-and-all-operators.md
+++ b/text/0021-any-and-all-operators.md
@@ -38,7 +38,7 @@ unless {
 };
 ```
 
-The two operators behave analogously to `&&` and `||`. Given negation, each can be written in terms of the other. We include both for ergonomics.
+The two operators behave analogously (though [not equivalently](#extending-the-semantics)) to `&&` and `||`. Given negation, each can be written in terms of the other. We include both for ergonomics.
 
 ## Motivation
 
--- a/text/0021-any-and-all-operators.md
+++ b/text/0021-any-and-all-operators.md
@@ -62,7 +62,7 @@ The new operators provide a built-in way to concisely express these universal an
 
 ## Detailed design
 
-The Cedar syntax, semantics, and type system are all extended to support the new operators. These extensions are backward compatible, and existing policies won't be affected.
+The Cedar syntax, semantics, and type system are all extended to support the new operators. These extensions are backward compatible, and existing policies won't be affected. In particular, we will keep the `.containsAny` and `.containsAll` operators on sets, even though they can be expressed in terms of `any` and `all`.
 
 ### Extending the syntax
 
--- a/text/0021-any-and-all-operators.md
+++ b/text/0021-any-and-all-operators.md
@@ -1,0 +1,165 @@
+# Add `all` and `any` operators
+
+## Related issues and PRs
+
+- Reference Issues:
+- Implementation PR(s):
+
+## Timeline
+
+- Start Date: 2023-07-14
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+This RFC proposes extending the language with `all` and `any` operators that allow checking if all elements or any element in a set satisfies a given predicate. The `all` operator returns `true` if the provided predicate is true for all elements in the set. The `any` operator returns true if the predicate is true for any element in the set.
+
+## Basic example
+
+Consider writing a policy that allows access when all port numbers in the `context` are between 8000 and 8999.  This can't be expressed in Cedar today.
+
+With the `all` operator, we can write this policy as follows:
+
+```
+permit(principal, action, resource)
+when {
+  context.portNumbers.all(it >= 8000 && it <= 8999)
+};
+```
+
+The same policy can be written using the `any` operator instead:
+
+```
+permit(principal, action, resource)
+unless {
+  context.portNumbers.any(it < 8000 || it > 8999)
+};
+```
+
+The two operators behave analogously to `&&` and `||`. Given negation, each can be written in terms of the other. We include both for ergonomics.
+
+## Motivation
+
+The `all` and `any` operators support a common use case: iterating over a set and checking if a predicate holds for any or all members of the set.
+
+Cedar currently lacks a general way to express this use case. The only alternative is to expand the `all` or `any` checks into conjunctions or disjunctions. For example:
+
+```
+// Using all
+["a", "ab", "abc"].all(it like "a*")
+
+// Using &&
+("a" like "a*") &&
+("ab" like "a*") &&
+("abc" like "a*")
+```
+
+This manual expansion works for literal sets but cannot be applied to non-literal sets like `context.portNumbers`.
+
+The new operators provide a built-in way to concisely express these universal and existential quantifications over a set.
+
+## Detailed design
+
+The Cedar syntax, semantics, and type system are all extended to support the new operators. These extensions are backward compatible, and existing policies won't be affected.
+
+### Extending the syntax
+
+This RFC proposes extending the Cedar grammar as follows:
+
+```
+Access    :=
+  '.' QUANTIFIER '(' Predicate ')'
+  | '.' IDENT ['(' [ExprList] ')']
+  | '[' STR ']'
+Quantifier := 'all' | 'any'
+Predicate  := ...
+```
+
+The `Predicate` expression differs from `Expr` in two ways:
+
+1. Predicates can use the keyword `it`. A `Predicate` defines the body of function with a single parameter, and we use `it` as the parameter name. This is similar to [Kotlin's `it` keyword](https://kotlinlang.org/docs/lambdas.html#it-implicit-name-of-a-single-parameter).
+
+2. Predicates cannot contain nested `any` or `all` expressions. We disallow nested quantifiers for both performance and analyzability reasons. Nested quantifiers allow writing expressions that run in $O(2^n)$ time, where $n$ is the number of nested quantifiers. For example:
+
+```
+[0,1].all(
+  [0,1].all(
+    ...
+      [0,1].all(true)...))
+```
+
+Because nested quantifiers are prohibited, there is no need to support predicates with arbitrary parameter names. The `it` keyword sufficiently extends the grammar.
+
+### Extending the semantics
+
+The semantics of `all` and `any` are straightforward when there are no errors:
+- `E.all(P)` evaluates to true when `P[e/it]` is true for every element `e` in `E`.
+- `E.any(P)` evaluates to true when `P[e/it]` is true for some element `e` in `E`.
+
+But what happens if evaluating `P` errors on some element `e`? In this case, the evaluation of `all` or `any` terminates abruptly with a distinguished `QuantifierError`.
+
+The `QuantifierError` error may include static information (such as source location), but no dynamic information (such as the value that caused the error to be thrown). This error handling approach ensures that evaluating  `all` and `any` is _deterministic_: the result is the same regardless of iteration order.
+
+For example, consider the expression `[1, true].all(it like "foo")`. Regardless of the order in which the predicate is applied, the expression returns the same `QuantifierError`.
+
+This semantics lets us think about `all` and `any` in terms of their expansion to conjunctions and disjunctions. The quantified expression errors if and only if the expansion would error.
+
+For example:
+```
+// Throws QuantifierError
+[1, true].all(it like "foo")
+
+// Throws a TypeError: "like" can't be applied to an integer
+1 like "foo" && true like "foo"
+
+// Throws a TypeError: "like" can't be applied to an boolean
+true like "foo" && 1 like "foo"
+```
+
+The only difference is the error being thrown: `QuantifierError` for the quantified expression versus an element-specific error for the explicit expansion.
+
+### Extending the type system
+
+The type system needs to be extended with rules for type checking quantified expressions `E.all(P)` and `E.any(P)`. This should be fairly standard: we know the type of `it` from the type of the set `E`, and we can use this to typecheck the predicate `P`. We also need to propagate effects through predicates.
+
+For example, the following should typecheck when `context.limit` is an optional integer attribute, and `resource.things` is a set of entities with an optional integer attribute `.size`:
+
+```
+context has limit &&
+resource.things.all(
+  it has size && it.size < context.limit
+)
+```
+
+## Drawbacks
+
+- This proposal requires significant development effort. We'll need to modify the parser, the CST to AST conversion, CST/AST/EST, evaluator, validator, models, proofs, and DRT.
+- The `any` and `all` operators may encourage writing complex expressions with poor performance. For example, `context.portNumbers.any(it == 8010)` is $O(n)$ versus $O(1)$ for `context.portNumbers.contains(8010)`.
+- Supporting these operators will likely have a negative impact on analysis performance, especially for complex predicates.
+
+## Alternatives
+
+We considered two alternative semantics for these operators.
+
+### Non-deterministic error behavior
+
+Given `E.any(P)` or `E.all(P)`, we apply `P` to elements of `E` in arbitrary order, terminating on the first error and returning that error as the result.
+
+Under this semantics, the same policy could produce different errors depending on iteration order. For example, `[1, true].all(it like "foo")` may result in a type error that says `like` cannot be applied to an integer or a boolean. In other words, Cedar semantics would become non-determinstic.
+
+We decided against this semantics for three reasons:
+* Violates Cedar's design as a deterministic language. Error non-determinism would be visible to `is_authorized` callers.
+* Applications may also become non-deterministic if they examine errors and act on them.
+* Non-determinism complicates models and proofs.
+
+Keeping deterministic semantics is important for Cedar as an authorization language.
+
+### Treating errors as `false`
+
+Given `E.any(P)` or `E.all(P)`, we apply `P` to elements of `E` in any order, treating errors as false and continuing evaluation on the remaining elements.
+
+With this approach, `any` and `all` become total functions: they never error. For example, `["one", 1].any(it < 2)` evaluates to true because there is one element for which the predicate holds. And `["one", 1].all(it < 2)` is simply false, because `"one" < 2` is treated as false.
+
+We decided against this semantics because it breaks the interpretation of `all` and `any` as shorthands for conjunctions and disjunctions. If we expand `["one", 1].all(it < 2)` to a conjunction, the result always errors rather than evaluating to false. This discrepancy between the quantified and expanded forms might be confusing and undesirable.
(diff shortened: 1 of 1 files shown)


text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+### Extending the syntax
+
+This RFC proposes extending the Cedar grammar as follows:
+
+```
+Access    :=
+  '.' QUANTIFIER '(' Predicate ')'
+  | '.' IDENT ['(' [ExprList] ')']
+  | '[' STR ']'
+Quantifier := 'all' | 'any'
```
mwhicks1, line 76: `Quantifier` --> `QUANTIFIER`

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+
+The semantics of `all` and `any` are straightforward when there are no errors:
+- `E.all(P)` evaluates to true when `P[e/it]` is true for every element `e` in `E`.
+- `E.any(P)` evaluates to true when `P[e/it]` is true for some element `e` in `E`.
+
+But what happens if evaluating `P` errors on some element `e`? In this case, the evaluation of `all` or `any` terminates abruptly with a distinguished `QuantifierError`.
+
+The `QuantifierError` error may include static information (such as source location), but no dynamic information (such as the value that caused the error to be thrown). This error handling approach ensures that evaluating  `all` and `any` is _deterministic_: the result is the same regardless of iteration order.
+
+For example, consider the expression `[1, true].all(it like "foo")`. Regardless of the order in which the predicate is applied, the expression returns the same `QuantifierError`.
```
mwhicks1, line 105: Add:

> Compare this to the alternative semantics that just propagates the error discovered, when it is discovered. Then evaluating this expression could lead to the error that "a Long was found where a String was expected" or it could lead to the error "A Boolean was found where a String was expected." Which one depends on evaluation order, but no clear order exists because set elements are specifically unordered.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+- This proposal requires significant development effort. We'll need to modify the parser, the CST to AST conversion, CST/AST/EST, evaluator, validator, models, proofs, and DRT.
+- The `any` and `all` operators may encourage writing complex expressions with poor performance. For example, `context.portNumbers.any(it == 8010)` is $O(n)$ versus $O(1)$ for `context.portNumbers.contains(8010)`.
+- Supporting these operators will likely have a negative impact on analysis performance, especially for complex predicates.
+
+## Alternatives
+
+We considered two alternative semantics for these operators.
+
+### Non-deterministic error behavior
+
```
mwhicks1, line 147: You should indicate the advantage of this approach, though: It provides more specific error information. The generic error is not particularly helpful. There are lots of errors that could have happened and now the user has to guess which one. If you provided either of the two specific errors in the example, that would provide more information to the user to fix the bug.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+## Drawbacks
+
+- This proposal requires significant development effort. We'll need to modify the parser, the CST to AST conversion, CST/AST/EST, evaluator, validator, models, proofs, and DRT.
+- The `any` and `all` operators may encourage writing complex expressions with poor performance. For example, `context.portNumbers.any(it == 8010)` is $O(n)$ versus $O(1)$ for `context.portNumbers.contains(8010)`.
+- Supporting these operators will likely have a negative impact on analysis performance, especially for complex predicates.
+
+## Alternatives
+
```
mwhicks1, line 143: What about this alternative:

You evaluate the predicate on _every_ member of the set, and collect all of the errors to be returned in a deterministic order? So in your example `[1, true].all(it like "foo")` you'd return the equivalent of 
```
[(true,TypeError("Boolean found where String expected"),
 (1,TypeError("Long found where String expected"))]
```
The order in the list is based on a global order defined on Cedar values, e.g., that all Booleans come before all Longs.

mwhicks1: Alternatively, you could sort based on error type, and ignore the contents of the error message and the expression that accompanies it. I believe that matches the level of abstraction we have today when considering Cedar error messages in the model.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+The type system needs to be extended with rules for type checking quantified expressions `E.all(P)` and `E.any(P)`. This should be fairly standard: we know the type of `it` from the type of the set `E`, and we can use this to typecheck the predicate `P`. We also need to propagate effects through predicates.
+
+For example, the following should typecheck when `context.limit` is an optional integer attribute, and `resource.things` is a set of entities with an optional integer attribute `.size`:
+
+```
+context has limit &&
+resource.things.all(
+  it has size && it.size < context.limit
+)
```
mwhicks1, line 133: While the effect prior to a `.any` or `.all` is clear, what about the effect after one? The simplest thing would be to ignore the effects, but we could be more precise. For example
```
resource.things.all(it has size && it.size < 5) &&
context has limit &&
resource.things.all(it.size < context.limit)
```
The second use of `all` could benefit from the first, which has confirmed that all the elements of `resource.things` have the `size` attribute. A simpler example would be
```
resource.things.all(context has limit && it.size < context.limit) &&
context.limit > 5
```
Here, the redundant use of `context.limit` in the `all` carries an effect outside to the `>` expression.

text/0021-any-and-all-operators.md
emina: I don't think that using a list is a viable alternative because it forces the specification to impose an (arbitrary) order on values and/or error types.  Then, every implementation has to enforce the same order.

If we want to go this route, how about returning a _set_ of error messages instead?  

Not that this a non-trivial change because it changes the return type of the evaluator from `Result<Value, Error>` to `Result<Value, Set<Error>>`. 

Another alternative is to package this error set into the QuantifierError.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+Under this semantics, the same policy could produce different errors depending on iteration order. For example, `[1, true].all(it like "foo")` may result in a type error that says `like` cannot be applied to an integer or a boolean. In other words, Cedar semantics would become non-determinstic.
+
+We decided against this semantics for three reasons:
+* Violates Cedar's design as a deterministic language. Error non-determinism would be visible to `is_authorized` callers.
+* Applications may also become non-deterministic if they examine errors and act on them.
+* Non-determinism complicates models and proofs.
+
+Keeping deterministic semantics is important for Cedar as an authorization language.
+
+### Treating errors as `false`
```
emina, line 159: A big advantage of this approach is that it is much cheaper to analyze than any of the alternatives that propagate errors.  Will add to the RFC.

text/0021-any-and-all-operators.md
emina: Good callout.  Yes, this would increase precision, though it might not be worth the extra complication in the implementation.  For these examples at least, there are natural ways to rewrite them so that don't rely on the extra precision.  For example, the first expression could be rewritten as

```
context has limit &&
resource.things.all(it has size && it.size < 5 && it.size < context.limit)
```

text/0021-any-and-all-operators.md
emina: We can still preserve the mapping:  use a map from values to errors.  That still avoids having to impose an arbitrary ordering in the specification, and allows implementations to use a HashMap if desired.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+- `E.all(P)` evaluates to true when `P[e/it]` is true for every element `e` in `E`.
+- `E.any(P)` evaluates to true when `P[e/it]` is true for some element `e` in `E`.
+
+But what happens if evaluating `P` errors on some element `e`? In this case, the evaluation of `all` or `any` terminates abruptly with a distinguished `QuantifierError`.
+
+The `QuantifierError` error may include static information (such as source location), but no dynamic information (such as the value that caused the error to be thrown). This error handling approach ensures that evaluating  `all` and `any` is _deterministic_: the result is the same regardless of iteration order.
+
+For example, consider the expression `[1, true].all(it like "foo")`. Regardless of the order in which the predicate is applied, the expression returns the same `QuantifierError`.
+
+This semantics lets us think about `all` and `any` in terms of their expansion to conjunctions and disjunctions. The quantified expression errors if and only if the expansion would error.
```
cdisselkoen, line 107: I'm not seeing how this works (and maintains determinism).  The expansion will have short-circuiting behavior, but I don't see how the quantifier could have short-circuiting behavior without relying on the order of elements in the set.

Example: `[2, "foo"].any(it < 3)`. Is this `true`, or a quantifier error? In the expansion, this could be either `true` or an error, depending on the order of the set elements in the expansion.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Extending the semantics
+
+The semantics of `all` and `any` are straightforward when there are no errors:
+- `E.all(P)` evaluates to true when `P[e/it]` is true for every element `e` in `E`.
+- `E.any(P)` evaluates to true when `P[e/it]` is true for some element `e` in `E`.
+
+But what happens if evaluating `P` errors on some element `e`? In this case, the evaluation of `all` or `any` terminates abruptly with a distinguished `QuantifierError`.
+
+The `QuantifierError` error may include static information (such as source location), but no dynamic information (such as the value that caused the error to be thrown). This error handling approach ensures that evaluating  `all` and `any` is _deterministic_: the result is the same regardless of iteration order.
```
cdisselkoen, line 103: I find this unsatisfying. I'd love for there to be a solution that at least provides one specific error message for some erroring element.  Users will want to know what kind of error was encountered during evaluation of the predicate, especially for complex predicates.  Attribute not existing is very different from a type error which is very different from integer overflow, and these all have very different solutions from the user's perspective.

cdisselkoen: I see that I've partially duplicated @mwhicks1 comment on lines 146-147

text/0021-any-and-all-operators.md
emina: Here, the expansion is done in the order of evaluation.  Expansions make sense only for set literals, and set literals assume left-to-right evaluation order, like `&&` and `||`.  So, this assumption is extended to the expansion.  Will clarify.

cdisselkoen: Ah, so the expansion analogy is limited to set literals and doesn't hold for nonliterals

text/0021-any-and-all-operators.md
emina: Thinking about this more:  returning a map from values to errors makes both evaluation and authorization asymptotically more expensive.  

In the worst case, the size of the answer returned by the authorizer will be linear in the size of both the policy slice _and_ the entity slice with this approach.  For example, imagine a set that contains 1,000 elements of the wrong type.  Then this alternative will return a map from 1,000 of these elements to the errors they caused. 

This alternative is fundamentally changing the size complexity of the evaluation and authorization results.  I don't think we should adopt it for this reason.  It may be impossible to return the full result over the wire if/when that is desired (e.g., AVP and similar services).

mwhicks1: The return value is now O(n) in the size of the policy, whereas before it was O(1), but the evaluation time is O(n) either way, is that right?

What if we had some sort of debug mode, where you return "quantifierError" in the general case, but return the map in debug mode?

emina: It's somewhat worse.  

For evaluation, the size of the return value becomes O(n) in the size of the input, where it used to be O(1) with respect to both input and policy size.  For authorization, the size of the return value becomes O(n * p + p) where n is the size of the input and p is the number of policies. 

Is the idea that we surface a debug mode in the language spec too?  Or don't surface / test / model / prove anything about it?

cdisselkoen (approved):
I'll approve this as-is because I think it would be an improvement to the Cedar language and unlock additional use-cases.  I'm still somewhat unsatisfied with an opaque `QuantifierError` instead of any specific, actionable feedback when a policy fails due to an error in a predicate.  But I think the RFC as-written is still a positive change vs the status quo.

emina:
> I'll approve this as-is because I think it would be an improvement to the Cedar language and unlock additional use-cases. I'm still somewhat unsatisfied with an opaque `QuantifierError` instead of any specific, actionable feedback when a policy fails due to an error in a predicate. But I think the RFC as-written is still a positive change vs the status quo.

Agree on all counts.  One thing that might help in the future is developing debugging tools for Cedar.  Both @aaronjeline and @mwhicks1 have mentioned this in the past and in this conversation.  This would likely need an alternative interpreter and/or interpretation mode, and we could support detailed error messages in the debugger.

text/0021-any-and-all-operators.md
mwhicks1: Yes, I think we surface some sort of debug mode. My thinking goes like this:
1. In production, an authorization request returns a response that indicates a policy has produced an error
2. The customer application notes this fact in its own logs, along with information about the request, i.e., the PARC and provided entity data
3. Later, the dev team can replay the request against the identified policy in debug mode, which then produces the detailed error message indicating what element of the list actually failed

As long as step 1 returns sufficient information for customers to do steps 2 and 3, once a debug-mode exists, I'm OK with this RFC as is.

mwhicks1: Debug mode probably is another evaluator implementation which we have to prove/test is equivalent to the production one.

text/0021-any-and-all-operators.md
```diff
@@ -0,0 +1,165 @@
...
+The same policy can be written using the `any` operator instead:
+
+```
+permit(principal, action, resource)
+unless {
+  context.portNumbers.any(it < 8000 || it > 8999)
+};
+```
+
+The two operators behave analogously to `&&` and `||`. Given negation, each can be written in terms of the other. We include both for ergonomics.
```
anwarmamat, line 41: `any` behaves differently than`||`. `any` does not short-circuit while `||` does. For example. ` 2 > 1  || true > 1` evaluates to `true`, but `[2, true].any(it > 1)` errors.

anwarmamat (approved):

text/0021-any-and-all-operators.md
mwhicks1: @emina I think @anwarmamat makes an important point here, which is glossed over in the text of the RFC: using `||` and `&&` is order dependent due to short circuiting, whereas use of `any` and `all` cannot short circuit in order to give an order independent result. I'd aim to clarify this point in the RFC text throughout. (There is one place that talks about the use of `contains` as opposed to its encoding using `any`, but this is an indirect statement about the avoidance of short circuiting, not a direct one.)

emina: This is explained in detail in the semantics section: https://github.com/cedar-policy/rfcs/blob/any-and-all-operators/text/0021-any-and-all-operators.md#extending-the-semantics

In the case of errors, `.any` and `.all` will error when _any_ expansion to `||` and `&&` errors.  

So, the meaning is "analogous" but not "equivalent".

emina: Added a forward pointer to the semantics section in the intro, explicitly stating that the expansion is not equivalent to the use of `||` and `&&`.

text/0021-any-and-all-operators.md
cdisselkoen: This sounds reasonable to me.  

As an implementation note, the natural choice for the map from `foo` element to error would probably be a map from `Expr` to error. (Note that `Expr` implements `Hash` so it can be put in an unordered hashmap.)  But I'm a little less clear on what type the keys should be if you are calling `.any()` or `.all()` on a set that originates from the entity data or `Context`.  `Expr` is still one option (we could easily convert the `RestrictedExpr`s to `Expr` and return them in that form), but RFC 34 proposes not keeping around the `RestrictedExpr` by the time we get to evaluation.  At that point you'd have to have the key be `Value`, but those are not `Hash`.  `Value`s are totally ordered, so you can put them in a treemap, but then your concern above about appearing unordered is kind of a fiction.  Moreover, if `Value`s are indeed the key type, they'd have to be made public -- something which is also not currently the case but also proposed in RFC 34.

cdisselkoen: Perhaps an acceptable workaround to the above would be to have the key be the string representation of the `Value`; the string representation would be trivially total-ordered, hashable, and useful for debugging, while dodging the issues about whether `Value` is made public.

shaobo-he-aws (approved):

text/0021-any-and-all-operators.md
emina: I think we need to keep the size of the evaluator output constant.  Letting it grow linearly with input size (policies + request +  entities) means that the authorizer's output is worst-case quadratic in input size.  This is undesirable even if rare.

I've revised the RFC to relax the specification of `QuantifierError` so that it can include more information  in the error message, as long as the additional information is computed deterministically and is bounded in size.

mwhicks1 (approved):
Looks good! Let's make it happen.

andrewmwells-amazon (approved):

text/0021-any-and-all-operators.md
john-h-kastner-aws: I don't see a natural way to rewrite when the second quantifier is an `any`
```
context has limit &&
resource.things.all(it has size && it.size < 5) &&
resource.things.any(it.size < context.limit)
```

text/0021-any-and-all-operators.md
khieta: I had the same question as @anwarmamat here. I think the text could be more clear to callout that the behavior for `any` and `||` is different in the presence of errors.

khieta (approved):
Approving as-is, but I do have two small comments:
1. Added to the [thread above](https://github.com/cedar-policy/rfcs/pull/21#discussion_r1376493586).
2. What happens to `containsAll` and `containsAny`? Should we drop them from the language now that they can be implemented using `all` and `any`?

john-h-kastner-aws (approved):
I'm convinced that a single `QuantifierError` won't be good enough, but this still sound good as is.

mwhicks1:
> What happens to containsAll and containsAny? Should we drop them from the language now that they can be implemented using all and any?

I think we leave them, because their implementation can be more efficient than the corresponding implementations with `.any` and/or `.all`.

shaobo-he-aws (approved):

emina:
> > What happens to containsAll and containsAny? Should we drop them from the language now that they can be implemented using all and any?
> 
> I think we leave them, because their implementation can be more efficient than the corresponding implementations with `.any` and/or `.all`.

And removing them would be a breaking change.

emina:
> I'm convinced that a single `QuantifierError` won't be good enough, but this still sound good as is.

I'm curious, why don't you think this will be sufficient?  Note that `QuantifierError` can include additional implementation-specific information, as long as it's computed deterministically and bounded in size.

khieta:
> I think we leave them, because their implementation can be more efficient than the corresponding implementations with .any and/or .all.

Sounds good. Let's update the RFC text in case the implementer has this question down the line.

khieta (approved):

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

emina:
After further investigation, we've found that even the simplified proposal for `.any? / .all?` is not amenable to automated reasoning.  To support analysis in principle, we would have to restrict the operators even more:  specifically, the expressions on the left and right hand side of the operators cannot both refer to the same variable.  For practical analyzability, we believe that this restrictions would need to be much tighter:  the expression on the right-hand side should be a constant (literal).  

It's not clear at this point that such a highly restricted version of `.any? / .all?` is sufficiently useful.  

For that reason, I propose that we reject this RFC. 

We can consider revisiting this decision in the future if there is enough demand for specialized any/all checks, such as `anyLike? <pattern> / allLike? <pattern>` on sets of strings.

PR #24, opened 2023-07-24
Custom schema syntax

mwhicks1:
Proposal to add a custom syntax for Cedar schemas that is more readable than the current JSON.

[Rendered](https://github.com/cedar-policy/rfcs/blob/schema-syntax/text/0024-schema-syntax.md)

### DIFF
--- a/text/0024-schema-syntax.md
+++ b/text/0024-schema-syntax.md
@@ -8,7 +8,7 @@
 ## Timeline
 
 - Start Date: 2023-07-24
-- Date Entered FCP:
+- Date Entered FCP: 2023-08-22
 - Date Accepted:
 - Date Landed:
 
@@ -244,11 +244,10 @@ The custom syntax does not require such designators, for readability, so we need
 4. Reserve `__cedar` as a namespace to disambiguate extension/primitive types from others
 5. Handle entity/common typename overlaps in translation
 
-These rules aim to address several tenets
+These rules aim to address several (additional) tenets
 
 * Conflicting type names are an exception to be discouraged, rather than a norm, for readability and clarity
 * Schemas should not break when new extension types are added to Cedar
-* Conversion should be possible from the JSON to custom format, and vice versa, as much as possible
 * Backward-incompatible changes should be minimized
 
 #### Rule 1: Issue a warning (for both syntaxes) when a schema defines the same typename twice
@@ -293,11 +292,12 @@ If a schema defines an entity type and a common type with the same name, the ref
 
 Custom syntax users can rectify the situation by changing either the common type or entity type to a different name (or placing them in different namespaces). JSON syntax users can refer to both types distinctly, but will be warned that translation to the custom syntax will not be possible without changing the name. Translation tools can easily make this name change automatically.
 
-Evaluating these rules against our tenets above, we can see:
+Note that translation from JSON to custom syntax is only problematic when a schema defines both a common and entity type of the same name. Translation tools can easily address this issue by renaming the common type (and offering to change the type name in the original JSON)
+
+Evaluating these rules against our added tenets above, we can see:
 
 * Conflicting type names are discouraged by issuing a warning when there is overlap
 * Due to the choice in priority order, schemas will not break when new extension types are added
-* Translation from JSON to custom syntax is only problematic when a schema defines both a common and entity type of the same name. Translation tools can easily address this issue by renaming the common type (and offering to change the type name in the original JSON)
 * Only rule 4 (introducing the `__cedar` reserved namespace) is backward incompatible, and it is very unlikely to affect existing schemas.
 
 #### Rejected alternative designs for disambiguation
--- a/text/0024-schema-syntax.md
+++ b/text/0024-schema-syntax.md
@@ -1,0 +1,333 @@
+# Cedar Schema Syntax
+
+## Related issues and PRs
+
+- Reference Issues: 
+- Implementation PR(s): 
+
+## Timeline
+
+- Start Date: 2023-07-24
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+This document proposes a custom syntax for Cedar schemas, as an alternative to its JSON-based syntax. 
+
+The syntax was developed with the following goals in mind:
+
+* Use the same concepts in both the JSON-based and custom syntax, allowing them to be naturally inter-converted
+* Reuse Cedar policy syntax as much as possible, to help with intuition
+* When no Cedar policy syntax analogue exists, leverage ideas from other programming languages that favor readability
+* Use as few syntactic concepts as possible; similar ideas in the schema should look similar syntactically
+
+## Basic example
+
+Here is the proposed syntax for the [TinyTodo schema](https://github.com/cedar-policy/cedar-examples/blob/main/tinytodo/tinytodo.cedarschema.json), whose auto-formatted JSON version is 160 lines long.
+```
+namespace "" {
+    entity Application;
+    entity User in [Team,Application] { name: String };
+    entity Team in [Team,Application];
+    entity List in [Application] {
+        owner: User,
+        name: String,
+        readers: Team,
+        editors: Team,
+        tasks: Set<{name: String, id: Long, state: String}>
+    };
+
+    action CreateList, GetLists 
+        appliesTo { principal: [User], resource: [Application] };
+
+    action GetList, UpdateList, DeleteList, CreateTask, UpdateTask, DeleteTask, EditShares 
+        appliesTo { principal: [User], resource:[List] };
+}
+```
+## Motivation
+
+Cedar schemas for non-toy applications are hard to read and write. The root cause of the difficulty is the use of JSON:
+
+* JSON has low information density. For three applications we’ve modeled — [TinyTodo](https://github.com/cedar-policy/cedar-examples/blob/main/tinytodo/tinytodo.cedarschema.json), [Document Cloud](https://github.com/cedar-policy/cedar-examples/blob/main/cedar-example-use-cases/document_cloud/schema.json), and [GitHub](https://github.com/cedar-policy/cedar-examples/blob/main/cedar-example-use-cases/github_example/schema.json) — schemas span 4-5 pages.
+* JSON does not support comments. This means that any design intent for a schema cannot be expressed inline.
+
+We believe that a custom syntax for Cedar schemas can help. It can be more information-dense, support comments, and leverage familiar syntactic constructs. We think customers will be happier with a custom syntax: we regularly hear favorable reviews of Cedar's custom policy syntax, as compared to a JSON-based syntax.
+
+## Detailed Design
+
+### Comments
+
+Cedar schemas can use line-ending comments in the same style as Cedar, e.g., `// this is a comment`.
+
+### Namespace format
+
+```
+namespace "My::Namespace" {
+  ...
+}
+```
+
+A Schema can have multiple namespace declarations. The contents of each (`...` above) comprise entity, action, and common-type declarations, all of which are unordered; i.e., they are all allowed to mutually refer to one another (but see discussion on common types below). The recommended style is to put entity declarations first, followed by action declarations, with common type declarations interspersed as needed. Entity, action, and common types referenced in the same namespace in which they are defined need not be prefixed by that namespace.
+
+### Entity declarations
+
+Here is an example illustrating the entity declaration syntax.
+
+```
+entity User in [Group] {
+    personalGroup: Group,
+    blocked: Set<User>
+};
+entity Document {
+    owner: User,
+    isPrivate: Boolean,
+    publicAccess: String,
+    viewACL: DocumentShare,
+    modifyACL: DocumentShare,
+    manageACL: DocumentShare
+};
+entity DocumentShare;
+entity Public in [DocumentShare];
+```
+
+The first declaration indicates that `User` is an entity type, can be a member of the `Group` type, and has two attributes: `personalGroup` with entity type `Group`, and `blocked` with type `Set<User>`.
+
+To allow syntactic parity with common `type` declarations, discussed below, we can optionally add `=` just before the `{ ... }` defining an entity's attributes, e.g., `entity Dog = { name: String }` rather than `entity Dog { name: String }`. Allowing `=` also supports a future extension in which an entity's shape is a type other than a record.
+
+Entity declarations may or may not have the `in ...` component, and they may or may not have attributes. A singleton list can be expressed without brackets, e.g., `entity User in Group = ...` The last element of an the attribute list can optionally include a comma; e.g., we could have written `blocked: Set<User>,` for `User`'s  `blocked` attribute declaration. 
+
+Entity types that share a definition can be declared together. For example, the following declares two entity types `User` and `Team`, both of which can be members of `Team` and both of which have a `name` attribute.
+```
+entity User, Team in [Team] { name: String };
+```
+
+### Action declarations
+
+Here is an example illustrating the action declaration syntax.
+```
+action ReadActions;
+action WriteActions;
+action CreateDocument 
+    appliesTo { principal: [User], resource: [Drive] };
+action DeleteDocument in [WriteActions] 
+    appliesTo { principal: [User], resource: [Document] };
+action ViewDocument in [ReadActions] appliesTo {
+    principal: [User,Public],
+    resource: Document,
+    context: {
+        network: ipaddr,
+        browser: String
+    }
+};
+```
+
+A declared action indicates either none, one, or both of the following:
+
+1. What action groups it may be a member of, using `in`
+2. What principal, resource, and context types it may be coupled with in a request (if any), using `appliesTo`
+
+Specifying `[]` for the `in` component is equivalent to omitting `in` entirely. An action that has no `appliesTo` component essentially names an action group: having no possible principal and/or resource types means there’s no way to submit a legal request with the action. 
+
+The `appliesTo` specificiation uses record syntax, where the "attributes" of the record may be any combination of `principal`, `resource`, and `context`, with the following constraints.
++ If `principal` and/or `resource` is given, the accompanying type must be either
+  - a single entity type,
+  - a list of entity types, or
+  - the syntax `[*]`, which is interpreted as allowing _any_ entity type (equivalently, the "unspecified" entity).
++ If a `principal` or `resource` element is not given, that means that any entity can be used for that element in a request (and is equivalent, therefore, to it having type `[*]`).
++ If the type given with `principal` or `resource` is `[]`, that means that _no_ entity is permitted in a request with this action, which essentially means that the action can only be used as a group
++ If `context` is given, the accompanying type must be a record type. If it is not given, the type `{}` is assumed.
+
+Here are two additional examples that follow these rules:
+
+```
+action ViewDocument in ReadActions appliesTo {
+    resource: Document,
+    context: {}
+};
+action CreateDocument appliesTo {
+    principal: [User],
+    resource: [*]
+};
+```
+
+Since actions are entities, action names are entity IDs, which can be arbitrary strings. Thus we admit the following more general syntax as well. 
+
+```
+action "Delete Document $$" in ["Write Actions"] appliesTo {
+    principal: [User],
+    resource: [Document]
+};
+```
+
+We anticipate future support for attributes as part of action entities. If/when they are supported, we can specify action attributes prior to the `appliesTo` part as an inline record using Cedar syntax. For example, here’s an extension of `ViewDocument` with action attributes.
+
+```
+action ViewDocument in [ReadActions] {
+  version: 1,
+  group: "meta"
+} appliesTo {
+  principal: [User,Public],
+  resource: Document,
+  context: {
+    network: ipaddr,
+    browser: String
+  }
+};
+```
+
+### Common types
+
+You can define names for types (as in Typescript, OCaml, Haskell, and many other languages) and then use them in `action` and `entity` declarations. Here is an example:
+
+```
+type authcontext = {
+    ip: ipaddr,
+    is_authenticated: Boolean,
+    timestamp: Long
+};
+entity Ticket {
+  who: String,
+  operation: Long,
+  request: authcontext
+};
+action view appliesTo { context: authcontext };
+action upload appliesTo { context: authcontext };
+```
+Note the use of `=` for `type` declarations, but the lack of (required) `=` for `entity` declarations; the `=` is needed because the definining type may not be a record, e.g., `type time = Long`.
+
+As with `entity` and `action` declarations, the name implicitly includes the surrounding namespace as a prefix. 
+
+While common types can be declared anywhere within a namespace and safely referenced by entity and action declarations, they cannot refer to one another, to avoid introducing definitional cycles. (To allow mutual reference, we could make declaration order matter: `entity`, `action`, and other `type` declarations cannot refer to a common type before it is defined.)
+
+### Mockup: Document Cloud
+
+Here is a version of the [Document Cloud](https://github.com/cedar-policy/cedar-examples/blob/main/cedar-example-use-cases/document_cloud/schema.json) schema.
+
+```
+namespace "" {
+    entity User in [Group] {
+        personalGroup: Group,
+        blocked: Set<User>
+    };
+    entity Group in [DocumentShare] { 
+        owner: User 
+    };
+    entity Document {
+        owner: User,
+        isPrivate: Boolean,
+        publicAccess: String,
+        viewACL: DocumentShare,
+        modifyACL: DocumentShare,
+        manageACL: DocumentShare
+    };
+    entity DocumentShare;
+    entity Public in [DocumentShare];
+    entity Drive;
+
+    action CreateDocument, CreateGroup 
+        appliesTo { principal: [User], resource: [Drive] };
+    action ViewDocument 
+        appliesTo { principal: [User,Public], resource: [Document] };
+    action DeleteDocument, ModifyDocument, EditIsPrivate, AddToShareACL, EditPublicAccess 
+        appliesTo { principal: [User], resource: [Document] };
+    action ModifyGroup, DeleteGroup 
+        appliesTo { principal: [User], resource: [Group] };
+}
+```
+
+### Mockup: GitHub Model
+
+Here is a version of the [GitHub](https://github.com/cedar-policy/cedar-examples/blob/main/cedar-example-use-cases/github_example/schema.json) schema.
+
+```
+namespace "" {
+    entity User in [UserGroup,Team];
+    entity UserGroup in [UserGroup];
+    entity Repository {
+        readers: UserGroup,
+        triagers: UserGroup,
+        writers: UserGroup,
+        maintainers: UserGroup,
+        admins: UserGroup   
+    };
+    entity Issue {
+        repo: Repository,
+        reporter: User
+    };
+    entity Org {
+        members: UserGroup,
+        owners: UserGroup,
+        memberOfTypes: UserGroup
+    };
+    action pull, fork, push, add_reader, add_triager, add_writer, add_maintainer, add_admin 
+        appliesTo { principal: [User], resource: [Repository] };
+    action delete_issue, edit_issue, assign_issue 
+        appliesTo { principal: [User], resource: [Issue] };
+}
+```
+
+## Drawbacks
+
+Here are two reasons not to create a custom schema syntax:
+- Adding another schema format means another format for customers to potentially need to learn. Mitigating this problem would be tooling that allows the custom syntax to be easily converted to the JSON syntax, which is easily achieved by parsing in the new format and pretty-printing in JSON.
+- Supporting a new syntax is an extra implementation cost, now and going forward, and an added potential for bugs. Mitigating this problem: The Rust internals already parse the JSON to a schema data structure, so adding a new format only involves adding a parser to this data structure and not, for example, making changes to the validator. We can use property-based testing to ensure that both formats are interconvertible, and correctly round-trip, i.e., that *parse(pretty-print(AST)) == AST*.
+
+## Unresolved questions
+
+One unresolved challenge is type name resolution. There are four kinds of type name:
+1. Primitive types, e.g., `Long`, `String`, etc.
+2. Extension types, currently just `ipaddr` and `decimal`
+3. Entity types
+4. Common types
+
+The first two are built in. The last two are schema-declarable; they have an associated namespace prefix, but references within the defining namespace can drop the prefix.
+
+In the current JSON syntax, common types may not overlap with primitive types; e.g., declaring a common type `Long` is disallowed. However, common types are permitted to overlap with extension types; e.g., declaring common type `ipaddr` is allowed. This works in the JSON because specifying an extension type requires an explicit `"type"` designator, as in the following:
+```
+"addr": {
+    "ip": { "type": "Extension", "name": "ipaddr" },
+}
+```
+If we _also_ had defined a common type `ipaddr` we would reference it by writing
+```
+"addr": {
+    "myip": { "type": "ipaddr" },
+}
+```
+Common types may also overlap with entity types, for similar reasons: To reference an entity type you must write `{ "type": "Entity", "name": "Foo" }` while to reference a common type you'd just write `{ "type": "Foo" }`.
+
+The custom syntax does not require such designators, for readability. How should type names that are ambiguous therefore be resolved? For example, what should happen when a schema references type `ipaddr` when it also defines `entity ipaddr` and `type ipaddr` ? 
+
+The main proposal already allows using namespace prefixes for disambiguation. If we had entity `ipaddr` and extension type `ipaddr` we could write `My::Namespace::ipaddr` to distinguish the former from the latter, assuming we are using a non-empty namespace. We can do likewise to distinguish entity type `Long` from primitive type `Long`. Namespaces are not sufficient to disambiguate entity types from common types, though, since they both can have namespace prefixes. Moreover, long namespace prefixes could create readability problems.
+
+Here are two possible solutions:
+
+### Typename prefixes
+
+The first solution is to create type-name prefixes for the purpose of disambiguation, which parallel the `"type"` part of the JSON syntax. Proposal:
+1. Extension types can be optionally prefixed with `+`, e.g., `+ipaddr` and `+decimal`
+2. Entity types can be optionally prefixed with `&`, e.g., `&User` and `&My::Namespace::List`.
+
+With these prefixes, we could write the following contrived schema:
+```
+entity ipaddr { name: String };
+type ipaddr = Long;
+entity container {
+    issuer: &ipaddr, // entity ipaddr
+    time: ipaddr,    // type ipaddr
+    addr: +ipaddr    // extension type ipaddr
+};
+```
+You would only write the prefix if there was a possibility for ambiguity. If you leave off a prefix but a name does not uniquely resolve, the schema parser will complain about the ambiguity.
+
+### Forbidding type name overlap
+
+As an alternative solution, we could forbid type-name overlaps entirely. Specifically, we could disallow creating any entity or common types that overlap with extension or primitive types, and we could disallow such types from mutually overlapping within the same namespace.
+
+This solution would lead to better readability (due to fewer concepts imposed on the reader for interpreting type names), but would require imposing a similar restriction on the JSON form of the syntax, which is a breaking change.
+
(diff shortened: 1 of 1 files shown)


shaobo-he-aws:
It looks great to me. It would be great if we can work out the grammar first so it can be reasoned about.

mwhicks1:
> It would be better if we can work out the grammar first so it can be reasoned about.

Do you think it's necessary to spell that out before approving? I figured that the example-driven presentation was easier to understand, and the details of the grammar won't be so hard to work out, once we've agreed on what's here.

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+
+The grammar applies the following the conventions. Capitalized words stand for grammar productions, and lexical tokens are given in all-caps. When productions or tokens match those in the Cedar policy grammar, we use the same names (e.g., `IDENT` and `Path`).
+
+For grammar productions it uses `|` for alternatives, `[]` for optional productions, `()` for grouping, and `{}` for repetition of a form zero or more times. 
+
+Tokens are defined using regular expressions, where `[]` stands for character ranges; `|` stands for alternation; `*` , `+` , and `?` stand for zero or more, one or more, and zero or one occurrences, respectively; `~` stands for complement; and `-` stands for difference. The grammar ignores whitespace and comments.
+
+```
+Schema    := {Namespace}
+Namespace := 'namespace' '"' Path '"' '{' {Decl} '}'
```
shaobo-he-aws, line 286: `Path` should be optional here?

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+Decl      := Entity | Action | TypeDecl
+Entity    := 'entity' IDENT ['in' (EntType | '[' [EntTypes] ']')] [['='] RecType] ';'
+Action    := 'action' Name ['in' (Name | '[' [Names] ']')] [AppliesTo] ';'
+TypeDecl  := 'type' IDENT '=' Type ';'
+Type      := PRIMTYPE | IDENT | SetType | RecType
+EntType   := Path
+SetType   := 'Set' '<' Type '>'
+RecType   := '{' [AttrDecls] '}'
+AttrDecls := Name ':' Type [',' | ',' AttrDecls]
```
shaobo-he-aws, line 295: In Schema-based format, record attributes have `required` fields. They are missing here.

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+Schema    := {Namespace}
+Namespace := 'namespace' '"' Path '"' '{' {Decl} '}'
+Decl      := Entity | Action | TypeDecl
+Entity    := 'entity' IDENT ['in' (EntType | '[' [EntTypes] ']')] [['='] RecType] ';'
+Action    := 'action' Name ['in' (Name | '[' [Names] ']')] [AppliesTo] ';'
+TypeDecl  := 'type' IDENT '=' Type ';'
+Type      := PRIMTYPE | IDENT | SetType | RecType
+EntType   := Path
+SetType   := 'Set' '<' Type '>'
+RecType   := '{' [AttrDecls] '}'
```
shaobo-he-aws, line 294: Should we allow empty records here?

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+For grammar productions it uses `|` for alternatives, `[]` for optional productions, `()` for grouping, and `{}` for repetition of a form zero or more times. 
+
+Tokens are defined using regular expressions, where `[]` stands for character ranges; `|` stands for alternation; `*` , `+` , and `?` stand for zero or more, one or more, and zero or one occurrences, respectively; `~` stands for complement; and `-` stands for difference. The grammar ignores whitespace and comments.
+
+```
+Schema    := {Namespace}
+Namespace := 'namespace' '"' Path '"' '{' {Decl} '}'
+Decl      := Entity | Action | TypeDecl
+Entity    := 'entity' IDENT ['in' (EntType | '[' [EntTypes] ']')] [['='] RecType] ';'
+Action    := 'action' Name ['in' (Name | '[' [Names] ']')] [AppliesTo] ';'
```
shaobo-he-aws, line 289: It should be `Names` instead of `Name` here?

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+TypeDecl  := 'type' IDENT '=' Type ';'
+Type      := PRIMTYPE | IDENT | SetType | RecType
+EntType   := Path
+SetType   := 'Set' '<' Type '>'
+RecType   := '{' [AttrDecls] '}'
+AttrDecls := Name ':' Type [',' | ',' AttrDecls]
+AppliesTo := 'appliesTo' '{' AppDecls '}'
+AppDecls  := VAR ':' Type [',' | ',' AppDecls]
```
shaobo-he-aws, line 297: `Type` should be `'[' [EntTypes] ']'` here?

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+    context: {
+        network: ipaddr,
+        browser: String
+    }
+};
+```
+
+A declared action indicates either none, one, or both of the following:
+
+1. What action groups it may be a member of, using `in`
```
cdisselkoen, line 130: what action groups it _is_ a member of?

cdisselkoen (approved):
I'm in favor of this.

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+
+    action GetList, UpdateList, DeleteList, CreateTask, UpdateTask, DeleteTask, EditShares 
+        appliesTo { principal: [User], resource:[List] };
+}
+```
+## Motivation
+
+Cedar schemas for non-toy applications are hard to read and write. The root cause of the difficulty is the use of JSON:
+
+* JSON has low information density. For three applications we’ve modeled — [TinyTodo](https://github.com/cedar-policy/cedar-examples/blob/main/tinytodo/tinytodo.cedarschema.json), [Document Cloud](https://github.com/cedar-policy/cedar-examples/blob/main/cedar-example-use-cases/document_cloud/schema.json), and [GitHub](https://github.com/cedar-policy/cedar-examples/blob/main/cedar-example-use-cases/github_example/schema.json) — schemas span 4-5 pages.
```
shaobo-he-aws, line 53: Wouldn't yaml kind of solve these two issues?

text/0024-schema-syntax.md
mwhicks1: I think we want to change this so that the entire `namespace { ... }` part is optional -- if you just put in declarations with no namespace then that implies it's the empty namespace.

text/0024-schema-syntax.md
mwhicks1: Good point! I will modify this to add an optional `?` after the name, in the style of Typescript. So the grammar should be
```
AttrDecls := Name ['?'] ':' Type [',' | ',' AttrDecls]
```

text/0024-schema-syntax.md
mwhicks1: It would solve the problem with comments. I'm not sure about information density. And I'm not sure it's an improvement of a syntax that more closely matches Cedar itself. (Otherwise: Why didn't we use Yaml for Cedar?)

text/0024-schema-syntax.md
mwhicks1: Yes, agreed!

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Alternatives
+
+One alternative would be to _replace_ the current JSON-based sytnax with the one in this proposal. This proposal would avoid the "cognitive burden" drawback mentioned above, but would be a disruptive, backward-incompatible change, and would lose the JSON format's benefits of existing tooling and easier programmatic schema construction.
+
+Another alternative would be to adopt a [Yaml](https://en.wikipedia.org/wiki/YAML)-based syntax. This approach would meet our goals of greater information density and support for comments, and it would come with some existing tooling (such as IDE extensions). A downside of Yaml is that it provides _more_ than we need, with a lack of conciseness leading to confusing. We could make our own parser for a subset of Yaml we wish to support for schemas, but that may lead to a confusing user experience. Yaml's indentation-sensitive parsing also means that an indentation mistake will be silently accepted, leading to a confusing user experience. Our custom syntax is whitespace-insensitive, and having total control over the grammar means better context for error messages.
```
khieta, line 347:
```suggestion
Another alternative would be to adopt a [Yaml](https://en.wikipedia.org/wiki/YAML)-based syntax. This approach would meet our goals of greater information density and support for comments, and it would come with some existing tooling (such as IDE extensions). A downside of Yaml is that it provides _more_ than we need, with a lack of conciseness leading to confusion. We could make our own parser for a subset of Yaml we wish to support for schemas, but that may lead to a confusing user experience. Yaml's indentation-sensitive parsing also means that an indentation mistake will be silently accepted, leading to a confusing user experience. Our custom syntax is whitespace-insensitive, and having total control over the grammar means better context for error messages.
```

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,405 @@
...
+```
+You would only write the prefix if there was a possibility for ambiguity. If you leave off a prefix but a name does not uniquely resolve, the schema parser will complain about the ambiguity.
+
+### Forbidding type name overlap
+
+As an alternative solution, we could forbid type-name overlaps entirely. Specifically, we could disallow creating any entity or common types that overlap with extension or primitive types, and we could disallow such types from mutually overlapping within the same namespace.
+
+This solution would lead to better readability (due to fewer concepts imposed on the reader for interpreting type names), but would require imposing a similar restriction on the JSON form of the syntax, which is a breaking change.
+
+As a middle ground, we could choose the first solution above, but _warn_ about overlapping type definitions, to discourage users from defining a common type with the same name as an entity or extension type.
```
khieta, line 403: My preference is for forbidding name overlap, and making the corresponding breaking change in the JSON syntax.

khieta (approved):

shaobo-he-aws (approved):

text/0024-schema-syntax.md
cdisselkoen: I'm in favor of forbidding the name overlap in the case of a conflict between an entity type and a common type in particular.  For conflicts involving extension types it's a little less clear, because then if we implement a new extension type, suddenly some people's existing schemas could become invalid if they happened to use an entity type or common type that has the same name as the new extension type.  This isn't hard to imagine either -- imagine someone writes `type date = int` or something for clarity of attribute types, and then later we actually implement a `date` extension type

mwhicks1: To summarize your position, @cdisselkoen: for extension types we would use `+` as a prefix for disambiguation but prevent overlap with entity and common types?

cdisselkoen: I only said it's less clear what to do about the extension types overlap, and I would say my mind is not made up yet.

Note that requiring a disambiguation prefix where none used to be required (eg because we added a `date` extension type) is also still a breaking change for user's schema

max2me:
Thank you for putting this together. Very interesting stuff. A few thoughts:

**Comments**  
Not strictly related to this proposal, but I am wondering why we don't support /* */ style comments?
I also wonder, what comments would be used for in schema most commonly? To describe what entity represents? If we can distill their use to specific cases, we could add 'proper' ways to address them through additional properties (that can be supported in JSON and visual modes).

**Extensibility**  
If tomorrow we will add new  attribute to entity types, for example `description`, what would that look like in this syntax?

**Syntax** 
* `in` vs `memberOfType` when defining entity types. I appreciate elegance and brevity of `in` but I do wonder if we should optimize for continuity between this syntax and JSON one. between two formats.
* I am very much in favor of having less 'optional' syntax rules. If `entity Dog = { name: String }` is allowed, then lets make `=` required everywhere. It does seem though that it might be unnecessary if we will support `type age Long` instead of `type age = Long` (I have not spent time studying grammar)
* "Bulk declarations" of multiple things with the single statement - although I see value of that, it does come with tradeoffs, especially when creating visual tools for editing schema. If two actions are declared together in a single statement, should visual editor represent it as such (when user edits this thing then both actions will be updated) or should it treat them as two separate items that get updated independently. This is something that's disambiguous in JSON syntax.
* Random thought: `,` separator between attributes could be optional?
* I find `appliesTo` syntax harder to comprehend when it's combined with action attributes. Although JSON goes overboard in terms of nested curly braces, there is certain clarity in what applies where.

```
action ViewDocument in [ReadActions] {
  version: 1,
  group: "meta"
} appliesTo {
  principal: [User,Public],
  resource: Document,
  context: {
    network: ipaddr,
    browser: String
  }
};
```
is a bit more natural to read in this format for me:
```
action ViewDocument in [ReadActions] {
  appliesTo {
    principal: [User,Public],
    resource: Document,
    context: {
      network: ipaddr,
      browser: String
    }
  }, 
  attributes {
    version: 1,
    group: "meta"
  } 
};
```

Swolebrain:
Generally, I really like this format a lot. 

I like the way `=` is being treated (optional for entities, non optional for commonTypes). This is very congruent with typescript.

Like max, I'm also curious about how you'll handle conversions back and forth between JSON and this syntax when you have composite declarations (eg. `action pull, fork, push, add_reader appliesTo { principal: [User], resource: [Repository] };`. 

I strongly agree with Craig's position that we should bake in a disambiguator mechanism in order to prevent future breaking changes. Thank you for thinking of that. Breaking changes represent a big burden for products which use Cedar as a platform. Is there a reason why the disambiguator prefix has to be one character? Why can't it be something like `__cedar_` or `__cedar.`? This reminds me of a situation you run into relatively often with javascript where you might name a local variable something like `escape`, but if you need to use the `escape()` function that's provided by JS, you can use `window.escape()`. I feel that a nice, ledible disambiguator prefix is the way to go here.

One question, how are we thinking about treating trailing commas? I hope we make them allowed (this is quite an annoyance with JSON).

One final thought I'm having is all the extensibility this syntax will allow. For instance, going from here to union and intersection types is conceptually a small jump. Union and intersection types would be a massive improvement in usability and would save space as well. If we did that though, we'd have to support it in JSON syntax as well. How feasible do we think this will be?

max2me (approved):

ChunghaSung:
I like the proposed custom syntax.

Swolebrain (approved):

mwhicks1:
Thanks @Swolebrain @max2me for your comments. I've just pushed an update that addresses several of them. In detail, here are changes I did not make:

* Cedar does not support bracketed comments because of ambiguity with parsing (I believe); we are following Cedar's lead.
* Using `in` rather than `memberOfType`, I agree, is one more thing a JSON and custom syntax reader should understand. But `in` is closer to core Cedar syntax, and indeed `memberOfType` is a concept that connects to `in` in Cedar policies. I don't think this is a big leap. Ultimately we want this custom syntax to be the default that people look at, not the JSON. The JSON is just for auto-manipulation, and to be the target of auto-generation tools
* Per @Swolebrain, optional `=` matches the approach of Typescript, is more future compatible. It also allows using a common type for the record type of an entity, e.g., `type f = { foo: String }; entity g = f;` Writing `type age Long` seems very awkward to me (and is unlike existing PLs). 
* Not having `,` between attributes would make either make whitespace significant, which we don’t want, and would diverge from regular Cedar syntax

For the rest:

* Bulk declarations are for readability, and I don’t think we want to drop that, since readability is a significant justification for the custom syntax. But I take the point of wanting to go back and forth from custom to JSON, so I added that as a tenet. To be able to support that, we can add an extra attribute to the JSON to indicate the grouping, e.g., for each action that is grouped in the custom syntax, you could have  "group" : "XXX" in the JSON; then all of those actions in the JSON with the same group ID are merged in the custom presentation.
* I like the proposal for the syntax for adding action attributes, and included that (with a slight tweak).
* Trailing commas are allowed
* I agree about 1-character disambiguating prefixes being odd, so I put together a proposal that avoids them, favoring a `__cedar` reserved namespace instead. It's been put in the main body of the proposal.

anwarmamat (approved):

shaobo-he-aws:
Correction of the grammar: Action attributes are unlike record type attributes. The former are mappings from names to Cedar values (i.e., restricted expressions) whereas the latter are mappings from names to types.

shaobo-he-aws:
> I have just one semi-related question. are there plans to provide a rust struct with public internals into which a schema (defined as json or with this new syntax) can be parsed. I am asking this because for example I now have a json schema and in my application I parse it into a type that I defined myself so that I can retrieve a list of actions and resources it applies to. with a schema defined with the new syntax I won't be able to do so so I'll have ot stick to the json schema for that reason.

Yes, we do have Rust structs that serve ASTs for schemas. I think what you want is in the [`cedar-policy-validator`](https://docs.rs/cedar-policy-validator/2.3.2/cedar_policy_validator/) crate. For instance, `cedar_policy_validator::ValidatorSchema` is likely sufficient. You can choose `cedar_policy_validator::SchemaFragment` if you want more flexibility. To be honest, I'm not sure what's best. @john-h-kastner-aws @cdisselkoen could you weigh in here?

cdisselkoen:
We don't want folks to have to use `cedar_policy_validator` directly, as we don't commit to keeping its interface stable.  I think a better solution would be to have more getters on [cedar_policy::Schema](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Schema.html) or [cedar_policy::SchemaFragment](https://docs.rs/cedar-policy/latest/cedar_policy/struct.SchemaFragment.html).

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,449 @@
...
+Action    := 'action' Names ['in' (Name | '[' [Names] ']')] [AppliesTo] [ActAttrs]';'
+TypeDecl  := 'type' IDENT '=' Type ';'
+Type      := PRIMTYPE | IDENT | SetType | RecType
+EntType   := Path
+SetType   := 'Set' '<' Type '>'
+RecType   := '{' [AttrDecls] '}'
+AttrDecls := Name ['?'] ':' Type [',' | ',' AttrDecls]
+AppliesTo := 'appliesTo' '{' AppDecls '}'
+ActAttrs  := 'attributes' '{' AttrDecls '}'
+AppDecls  := VAR ':' EntOrTyps [',' | ',' AppDecls]
```
shaobo-he-aws, line 400: This rule is not correct: What's after `context :` should be a `RecType`.

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,449 @@
...
+EntTypes  := Path {',' Path}
+EntOrTyps := EntType | '[' [EntTypes] ']'
+Name      := IDENT | STR
+Names     := Name {',' Name}
+Idents    := IDENT {',' IDENT}
+
+IDENT     := ['_''a'-'z''A'-'Z']['_''a'-'z''A'-'Z''0'-'9']* - PRIMTYPE
+STR       := Fully-escaped Unicode surrounded by '"'s
+PRIMTYPE  := 'Long' | 'String' | 'Bool'
+VAR       := 'principal' | 'action' | 'resource' | 'context'
```
shaobo-he-aws, line 411: It doesn't need to contain `action`, right?

shaobo-he-aws:
I updated my prototype parser and added it to the CLI. Now you can parse schemas in the new syntax and print them to JSON. For instance,

```
echo 'namespace GitHub {
    entity User in [UserGroup,Team];
    entity UserGroup in [UserGroup];
    entity Repository {
        readers: UserGroup,
        triagers: UserGroup,
        writers: UserGroup,
        maintainers: UserGroup,
        admins: UserGroup
    };
    entity Issue {
        repo: Repository,
        reporter: User
    };
    entity Org {
        members: UserGroup,
        owners: UserGroup,
        memberOfTypes: UserGroup
    };
    action pull, fork, push, add_reader, add_triager, add_writer, add_maintainer, add_admin
        appliesTo { principal: [User], resource: [Repository] };
    action delete_issue, edit_issue, assign_issue
        appliesTo { principal: [User], resource: [Issue] };
}'  | target/debug/cedar translate-schema | jq
{
  "GitHub": {
    "commonTypes": {},
    "entityTypes": {
      "Issue": {
        "memberOfTypes": [],
        "shape": {
          "type": "Record",
          "attributes": {
            "repo": {
              "type": "Entity",
              "name": "Repository",
              "required": true
            },
            "reporter": {
              "type": "Entity",
              "name": "User",
              "required": true
            }
          },
          "additionalAttributes": false
        }
      },
      "Org": {
        "memberOfTypes": [],
        "shape": {
          "type": "Record",
          "attributes": {
            "memberOfTypes": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            },
            "members": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            },
            "owners": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            }
          },
          "additionalAttributes": false
        }
      },
      "Repository": {
        "memberOfTypes": [],
        "shape": {
          "type": "Record",
          "attributes": {
            "admins": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            },
            "maintainers": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            },
            "readers": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            },
            "triagers": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            },
            "writers": {
              "type": "Entity",
              "name": "UserGroup",
              "required": true
            }
          },
          "additionalAttributes": false
        }
      },
      "User": {
        "memberOfTypes": [
          "UserGroup",
          "Team"
        ],
        "shape": {
          "type": "Record",
          "attributes": {},
          "additionalAttributes": false
        }
      },
      "UserGroup": {
        "memberOfTypes": [
          "UserGroup"
        ],
        "shape": {
          "type": "Record",
          "attributes": {},
          "additionalAttributes": false
        }
      }
    },
    "actions": {
      "add_triager": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "pull": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "add_admin": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "add_reader": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "add_writer": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "edit_issue": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Issue"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "add_maintainer": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "fork": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "assign_issue": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Issue"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "push": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Repository"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      },
      "delete_issue": {
        "attributes": null,
        "appliesTo": {
          "resourceTypes": [
            "Issue"
          ],
          "principalTypes": [
            "User"
          ],
          "context": {
            "type": "Record",
            "attributes": {},
            "additionalAttributes": false
          }
        },
        "memberOf": null
      }
    }
  }
}
```

emina (approved):

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,449 @@
...
+* Support converting back and forth between the JSON-based schema without losing information
+
+This is _not_ a proposal to _replace_ the JSON-based syntax; the aim is to provide an additional syntax that better satisfies the goals above. The JSON syntax is appropriate in many circumstances, such as when constructing schemas systematically. Adopting this proposal would put schemas on the same footing as Cedar policies, which have a custom and a JSON syntax.
+
+## Basic example
+
+Here is the proposed syntax for the [TinyTodo schema](https://github.com/cedar-policy/cedar-examples/blob/main/tinytodo/tinytodo.cedarschema.json), whose auto-formatted JSON version is 160 lines long.
+```
+entity Application;
+entity User in [Team,Application] { name: String };
```
emina, line 32: Suppose that we decide at some point to support named (transitively closed) relations, in addition to the (anonymous) hierarchy membership relation.  

What would be a nice way to extend the schema syntax to support these?

cdisselkoen: shouldn't we first decide what the Cedar syntax would be, and then use that to inform the schema syntax?

Good chance it could be the same. for instance if (strawman) the syntax is `A in(mycustomrelationname) B`, then the schema could use the same

emina: Yes, I imagine we'd follow the Cedar syntax.   My question would be more precisely phrased as this:  do we see something about the current proposal that's a one-way door and would make such an addition particularly onerous?  (I don't but I haven't spent as much time thinking about this design as the other commenters :) )

max2me:
I wonder if we should carry over attribute names from current syntax? So that `appliesTo.principal` would instead be `appliesTo.principalTypes`.

text/0024-schema-syntax.md
```diff
@@ -0,0 +1,449 @@
...
+namespace My::Namespace {
+  ...
+}
+```
+
+A Schema can have zero or more namespace declarations. The contents of each (`...` above) comprise entity, action, and common-type declarations, all of which are unordered; i.e., they are all allowed to mutually refer to one another (but see discussion on common types below). The recommended style is to put entity declarations first, followed by action declarations, with common type declarations interspersed as needed. Entity, action, and common types referenced in the same namespace in which they are defined need not be prefixed by that namespace.
+
+Entity, action, and common-type declarations can appear on their own, outside of any namespace; in that case they are implicitly within the empty namespace.
+
+### Entity declarations
```
caguado, line 77: The syntax and AST look very similar to Smithy and it would be a natural fit to model an API contract with the authorization domain model for it. Currently Smithy can model namespaces, generic data structures, and primitive types as well as `service`, `resource`, `operation` entities. We could define also something like `principals` and cedar traits to model the authz context type for the supported cedar operators on them.

An additional benefit of modelling the authz model for a cedar engine in smithy would be to also generate the SSDK stub for a service to fill in at runtime, exactly as the Smithy team does for the service payload.

I posit this would also help bringing up the SDF/CFN definitions and the reasoning efforts behind them.

So, have we considering such Smithy path? or is the intent to define a Cedar DSL and later bridge the two with a Smithy custom plugin to keep these two investments separate?

hakanson:
**CLI**

Should the RFC address impacts to the Cedar CLI, or is that an implementation concern?  Do we treat the CLI commands and parameters as a public api that need to be backward compatible?

For example, `cedar validate --schema XXX ...` or `cedar evaluate --schema XXX ...` both include a `--schema` parameter referencing a file, which today is the JSON format.  Would the CLI try and auto-detect the schema syntax (either by pre-parsing or trying both), require a `--format=json|rfc24` parameter, or other ?

**JSON**

Are changes being proposed to the JSON format to support comments? Options like the `__comment` key hack, or parsing against jsonc or [json5](https://json5.org/).  VS Code has some support for [JSON with Comments (jsonc)](https://code.visualstudio.com/docs/languages/json#_json-with-comments) but it has [issues](https://github.com/microsoft/vscode/issues/100688).

Should a top level `@version` or `$version` key be added to the JSON schema format (as well as the native syntax).  Maybe this gets added / required for Cedar 3.0 to assist in compatibility.  When not present, assume a value of `2` (representing major version of Cedar), and `3` could trigger different parsing / validation behaviors.

Should JSON format just be deprecated?  Future versions can still parse it, but may not support new language features.  This is slightly different than the alternative "to replace" w.r.t backward compatibility. 

**Annotations**

Should the custom syntax also support annotations to parallel cedar policies, possibly only at namespace level?  Maybe I want an `@author('hakanson')` which I could do with comments using a front-matter style syntax (`// author: hakanson`).

Should the schema provide a way to define Cedar policy annotations, either for required fields or to assist in editing/auto-complete?  For example, a `annotation` keyword at the same level as namespace.

```
annotation A1; // required
annotation A2?; // optional
namespace N {
  entity E;
  action a;
}
```

```
@A1('string')
permit (principal == N::E::"e", action == N::Action::"a", resource);
```

**Multiple files**

As schema size gets larger, or development teams want to split out namespace per file, are uses expect to concatenate all files together or should there be native support to "!include" one file into another?  Utilities like [cfn-include](https://github.com/monken/cfn-include) were created to pre-process CloudFormation templates and avoid merging conflicts on large template files.  Not sure if `include` should be a new keyword, overload annotation syntax like `@include`, or treat like an extension method.  This is always something that can be added later.

```
include('namespace1.cedarschema')
include('namespace2.cedarschema')
```

text/0024-schema-syntax.md
cmoher: I agree that this looks similar to Smithy.  I strongly recommend that consideration be given to consider using the Smithy IDL and perhaps contributing additional required traits/features/plugins.

mwhicks1:
[evgeniy-terekhin](https://github.com/evgeniy-terekhin) wrote

> are there plans to provide a rust struct with public internals into which a schema (defined as json or with this new syntax) can be parsed.

Good question. Two thoughts:

1) We aim to provide translation to/from the JSON format. In that case, the custom syntax could be translated to JSON, which you could then translate into your Rust struct as you are doing now.

2) We could aim to provide parity to what is done now with Cedar policies. We currently have the "extended syntax tree" (EST) into which you can parse Cedar policies from JSON. It seems like we should provide something similar for the schema.

mwhicks1:
https://github.com/cedar-policy/rfcs/pull/24#issuecomment-1703371260 suggested that instead of `principal` we use `principalTypes`. My feeling is to keep things as proposed. My reasoning is that the keyword in the custom syntax is slightly different than the same word in the JSON version. In particular, when I write
```
{ principal: [User], resource: [List] }
```
I am in effect defining a pseudo-"record" with a `principal` attribute and a `resource` attribute, and the the types of those attributes are `User` and `List`. The name of the attribute is `principal` because that's how I refer to it in a Cedar policy, i.e., as the variable `principal`.

I understand that this is a point of potential disconnect with the JSON version, but I'm not to worried about it for two reasons: 

1. I believe the custom syntax should be the primary way that users view Cedar schemas, going forward, just like Cedar policy custom syntax is the primary view. From this perspective, it makes sense to aim for syntactic similarity Cedar policies, which customers also will be looking at, rather than JSON schemas, which they will not

2. JSON schemas will exist primarily for transformation/analysis purposes, in code. Therefore, disconnects like this one will affect only those writing such transformation code, and this population is a lot smaller than the population writing schemas.

text/0024-schema-syntax.md
mwhicks1: I don't see any issues, no. Basically: Whatever we can work into Cedar policy syntax _should_ work out Ok for schemas, since there is a lot of similarity between the two.

mwhicks1:
Addressing questions from https://github.com/cedar-policy/rfcs/pull/24#issuecomment-1710906641

1. CLI: We should adjust the CLI, agreed. I'd suggest that by default we support the custom syntax, i.e., if no format flag is given the assumption is that it's the custom syntax. We could attempt to auto-infer the use of JSON, or we could have a separate flag for that. My reasoning is that we have a JSON format for Cedar policies too, but we default to the custom syntax.

2. Comments: What to do about comments is a good question. My suggestion is that when we translate from custom syntax with comments to JSON we could use the "key hack" as a best effort. We would not promise that round-trip conversion, i.e., custom --> JSON --> custom would perfectly preserve comment formatting. The start and finish in this round trip might land the comments in different places.

3. Versions: A top level `@version` or `$version` key makes sense -- whatever we do for policies we should do for schemas too.

4. Annotations and multiple files: These are interesting and useful ideas but I'd leave them to a future RFC. The goal is to get parity with current JSON for now, and it doesn't support this.

text/0024-schema-syntax.md
cdisselkoen: We met with the Smithy folks and, although there are ways for Cedar and Smithy to collaborate in the future, the Cedar schema isn't an IDL and there are enough differences between the goals of this syntax and the goals of the Smithy IDL that in my opinion we should proceed with this RFC as proposed.  Down the line we can discuss generating a Cedar schema from a Smithy API specification, but that would apply only for the particular case that Cedar is authorizing calls to a web API.

text/0024-schema-syntax.md
mwhicks1: Thanks for these. Updated the grammar.

emina (approved):

shaobo-he-aws (approved):

cdisselkoen (approved):

andrewmwells-amazon (approved):

text/0024-schema-syntax.md
caguado: Ack, thank you! Leaving out the reasoning aspects about a service posture that Smithy already models or what APIs Cedar is being used for, I'd argue that Cedar slices modelled with this doc's schema are an IDL, an authorization one if you will (and very powerful by the way). As a developer, I'd still have to create software types to marshall them when passing them over the wire to the service authorization subsystem and without supported typings I'll have to lift this undifferentiated work to bridge them with the entities of my application domain model already in Smithy, or Smithy-generated code for the language of choice, and which Cedar still requires to authorize on, and verify to get them right.

To be clear on the XY problem, given this schema syntax is to be its own DSL, are you considering improving dev-experince via code-generation tools for a polyglot service environment using Cedar?

cdisselkoen: We are thinking about the dev-experience; Cedar already has its own [VSCode extension](https://github.com/cedar-policy/vscode-cedar/) which supports the current JSON format for schemas and will also support this new DSL.  The dev experience for using Cedar from any language other than Rust is currently subpar and we're well aware of that, but that is a more fundamental problem that extends beyond schema syntax.  Cedar simply doesn't currently have APIs in non-Rust languages that have functionality parity with its Rust API.

> As a developer, I'd still have to create software types to marshall them when passing them over the wire

This is true for some applications using Cedar; but some Cedar users never send Cedar data over the wire at all, they just call the Cedar library directly.

@mwhicks1 do you have comments on any of this?

max2me:
Custom schema syntax already provides an explicit way of specifying that action applies to all principal types and resources types via `principal: [*]` and `resource: [*]`. I've [filed a feature request](https://github.com/cedar-policy/cedar/issues/320) to add this explicit way to standard JSON syntax.

Could you please consider changing semantics of undefined principal / resource attributes on action from 'all' to 'none' to reduce risk of customers unintentionally specifying that action applies to all entity types?

emina:
I would vote against providing a shorthand (such as `[*]`) for specifying that an action can apply to _all_ types.  In general, we want to encourage users to minimize the applicability / scope of their actions, for understandability, security, and efficiency of validation.  It's safest to explicitly specify (enumerate) the set of types to which an action applies.

I agree that we do want to provide some syntax for saying that the principal or resource type is "unspecified" (or "don't care").  The unspecified type is disjoint from all other types and has no attributes.  [RFC 19](https://github.com/cedar-policy/rfcs/blob/main/text/0019-stricter-validation.md#unspecified-principalresource-entity-types) has the details.

aaronjeline (approved):

PR #25, opened 2023-08-14
Update 0005-is-operator.md

verrchu:
provide another alternative


andrewmwells-amazon:
Thanks for pointing out this alternative. I'm going to close this PR because we've accepted the RFC and will be adding the `is` operator. I think it will be more readable for people use the `is` operator rather than creating a new `Type::"any"` entity as a workaround.

PR #27, opened 2023-10-02
Map types for records

mwhicks1:
Proposal for a way to treat records as maps in Cedar.

[Rendered](https://github.com/cedar-policy/rfcs/blob/record-map/text/0027-record-as-map.md)

### DIFF
--- a/text/0027-record-as-map.md
+++ b/text/0027-record-as-map.md
@@ -1,4 +1,4 @@
-# RFC Template (Fill in your title here)
+# Support to treat records as maps
 
 ## Related issues and PRs
 
@@ -14,7 +14,7 @@
 
 ## Summary
 
-Cedar records are fundamentally static. They force a user to specifically enumerate all possible record attributes in the schema, and they require an attribute to be mentioned as a literal in the policy. 
+Cedar records are fundamentally static. They force a user to specifically enumerate all possible record attributes in the schema, and they require an attribute to be mentioned as a literal when used in a policy.
 
 This RFC proposes to support a limited form of maps in Cedar, adding greater dynamism, by slightly extending current support for records. In the extension, not all attribute names need to be known in advance, and can be taken from run-time data. There are two changes:
 - Add a new type `map<T>` which represents records that optionally have _any_ attribute, where if an attribute is present it will have type `T`.
@@ -34,13 +34,13 @@ or
 ```
 when { resource.tags["projects"].contains(principal.tags["project"])}
 ```
-This would be difficult to do when trying to encode tags as sets (see discussion at the end of this RFC). `map`-based tags are most valuable when you don't know the tag names when writing policies, else you could extend the schema to name the tags explicitly.
+This would be difficult to do when trying to encode tags as sets. `map`-based tags are most valuable when you don't know the tag names when writing policies, else you could extend the schema to name the tags explicitly. See the [Workarounds](#workarounds) section below for discussion.
 
 #### Timeboxes
 
 A second example is the extension of [`TinyTodo`](https://github.com/cedar-policy/cedar-examples/tree/main/tinytodo) to include `TimeBox`es, which are fixed periods during which a principal is granted read-only access to a todo `List`. [Right now](https://github.com/cedar-policy/cedar-examples/blob/features/timebox/tinytodo/src/timebox.rs), this feature is implemented by generating a fresh policy whenever timed access is requested. This approach is unsatisfying because part of the application's authorization model thus lives in the application code, rather than in the policy file (i.e., the part that is generating the `TimeBox`-based policies). It also results in stale policies, since they remain in the store even after the timebox range has expired.
 
-With `map` types, `TimeBox`es for individual `User`s can be maintained separately from the policies. Each `List` could have a `timeboxes` map from a `User`'s name to time range that the `List` is readable.
+With `map` types, `TimeBox`es for individual `User`s can be maintained separately from the policies. Each `List` could have a `timeboxes` map from a `User`'s name to a time range that the `List` is readable.
 ```
 permit(
     principal,
@@ -52,7 +52,7 @@ permit(
     resource.timeboxes[principal.name].end < context.now
 }
 ```
-This approach adds just the above policy to the store at the outset, for all users and lists. The application manages timebox ranges -- adding and updating them in the relevant `List` entity -- the same as it was doing in the old approach.
+This approach adds just the above policy to the store at the outset, and it applies to all users and lists. The application manages timebox ranges -- adding and updating them in the relevant `List` entity -- the same as it was doing in the old approach.
 
 ## Detailed design
 
@@ -91,14 +91,14 @@ The JSON entity format does not change to support maps: We use the same format a
 
 ### Schema
 
-We extend schemas to support a new type `map` which is parameterized by the type of its values. In the JSON format of Cedar schema, a map to strings would be written
+We extend schemas to support a new type `map` which is parameterized by the type of its values. In the JSON format of Cedar schema, the type of a map to strings would be written
 ```
     "type": "Map",
     "element": {
         "type": "String"
     }
 ```
-In the proposed [custom schema syntax](https://github.com/cedar-policy/rfcs/pull/24), we would add the type `map<T>` where `T` is any type (e.g., `map<String>`), analogous to the existing type `set<T>`.
+In the proposed [custom schema syntax](https://github.com/cedar-policy/rfcs/pull/24), we would add the type `map<T>` where `T` is any type (e.g., `map<String>` or `map<map<Long>>`), analogous to the existing type `set<T>`.
 
 ### Validation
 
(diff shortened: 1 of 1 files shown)


max2me:
Thank you for writing this up!

If we assume that [RFC 21 Add all and any operators](https://github.com/cedar-policy/rfcs/blob/any-and-all-operators/text/0021-any-and-all-operators.md#add-all-and-any-operators) can be extended to support `.find()` operator (similar to [Array.find() in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)) that returns first matching item in the set, then we have two proposals that would help achieve the same outcomes in provided tags and timebox examples.

This RFC tackles these use cases through new data structure and enforces uniqueness of the key.
RFC 21 deals with it through new operators.

Motivating example using Map types
```
when { resource.tags["project"] == principal.tags["project"] } 
```

can then be rewritten as 
```
when { resource.tags.find(it.key == "project").value == principal.tags.find("it.key == "project").value }
```

I personally lean towards using predicates for their flexibility although I appreciate simplicity of having Map types for when element is a primitive. It does seem that RFCs are generally compatible so we could do both.

emina:
> when { resource.tags.find(it.key == "project").value == principal.tags.find("it.key == "project").value }

Following up on @max2me's example from above, we can express this constraint without `.find`, using just `.any`, like this:

```
when { resource.tags.any(it.key == "project" && principal.tags.contains({key: "project", value: it.value})) }
```

cdisselkoen:
> `.find()` operator that returns first matching item in the set

Cedar sets are unordered, so "first" doesn't make sense.  You could potentially make a version of `.find()` that just returns _any_ matching item in the set, though. That said, @emina's refactoring to use just `.any()` is probably the easier/preferable way to do this

cplee:
This would address https://github.com/cedar-policy/cedar/issues/305

max2me:
> > `.find()` operator that returns first matching item in the set
> Cedar sets are unordered, so "first" doesn't make sense. You could potentially make a version of `.find()` that just returns _any_ matching item in the set, though. 
Fair point, any one matching item in the set works too.

> That said, @emina's refactoring to use just `.any()` is probably the easier/preferable way to do this
I'd love to understand what makes it easier/preferrable. I find that approach very creative but harder to read. It also assumes that you know the entire value upfront.

emina:
The approach with .any doesn’t require that you know the value upfront.  It’s equivalent to the .find approach in that sense.  Note that we effectively find it.value for one set and then use that value to perform the membership test in the second set.

emina:
The .find operator is tricky to define. There is no notion of “first” and we can’t pick any value non-deterministically because the language must remain deterministic, for safety.

max2me:
Here is an example where this approach might not be sufficient. Currently each tag is a `{key, value}` record. Imagine that instead each tag is a Record with 3 attributes `{key, value, updatedAt}`. 

Although this is not an example based on real use case, if I understand approach correctly, contains won't work because it will require supplying valid `updatedAt` attribute that we do not know.

emina:
The approach still works, even with three attributes, assuming that we want to check equality of the entire record (as in the previous example).  It would look like this:

```
when { 
  resource.tags.any(
    it.key == "project" && 
    principal.tags.contains({key: "project", value: it.value, updatedAt: it.updatedAt})) 
}
```

It wouldn’t work if we wanted to ignore the `.updatedAt` attribute in the equality comparison, but in that case, we could represent the data differently.   We can use _two_ sets of key/value pairs.  So, we’d have `.tags`, which maps keys to values, and also `.updatedAt`, which maps keys to timestamps. 

Another disadvantage of `.find`, besides the tricky semantics, is that it negatively affects analyzability.  We could define `.find` to return a value only if it’s the sole such value in the set; otherwise, it errors.  This can be encoded for analysis, but it’s unclear how well the resulting encoding would be perform.

mwhicks1:
I updated RFC to include discussion above, in the "workarounds" part at the end. I'm becoming more convinced that we don't see this RFC, and that RFC 21 plus what we have is the way to go.

andrewmwells-amazon:
> I updated RFC to include discussion above, in the "workarounds" part at the end. I'm becoming more convinced that we don't see this RFC, and that RFC 21 plus what we have is the way to go.

I agree. Should we move to FCP with intent to reject?

mwhicks1:
I've thought a little more about what we gain/lose by using an encoding of maps using sets. 

From what I can see, the encoding is:
- a value of type `map<K,V>` has type `set<{ key: K, value: V }>`
- given map `m`, a predicate (boolean expression) `E` involving `m[k]` would be written `m.any(it.key == k && E)` where occurrences of `m[k]` in `E` are replaced with `it.value`. For example, checking whether `m["foo"] == "bar"` would be written `m.any(it.key == "foo" && it.value == "bar")`.

The predicate `E` is limited by our inability to nest `any` and `all`, though. 

In particular if `E` mentions any other map lookups, you can't use this encoding. There are some tricks you can play by using `contains` or `containsAny`, though. For example, to encode `m[k] == n[k]`, you could write `m.any(it.key == k && n.contains({key: k, value: it.value}))`. But you can't encode `m[k] > n[k]` because `contains` is limited to use of `==`. One possible solution to any inadequacy we find is to add more set operators like `contains` to fill the gap, but I'm not sure what they would be.

So the question is: Is this limitation significant enough that the encoding is inadequate? What other expressions `E` per the above might we like to see in policies that could involve multiple maps? @max2me @emina

emina:
In my view, the cost of adding Maps (today) significantly outweighs the benefits of doing so (today).

For real use cases we've seen so far, there are viable alternative solutions.  And while there are hypothetical use cases that can't be handled, they don't sufficiently justify the added complexity brought by Maps.

Adding Maps is an invasive change to the language:  it's adding a new type, changing the syntax and semantics, changing validation, analysis etc.  This invasive change does add expressiveness but at the following cost:

1. It adds a non-trivial cognitive burden on users.  Cedar used to have Maps (prior to the open source release), and we removed them from the language based on the feedback from users.  Users found the distinction between Maps and records subtle and confusing.  It wasn't clear when to use one vs the other when both applied.  And the differences in how the two behave with respect to indexing (static vs dynamic) were also confusing. Some of this can be addressed with better documentation, but fundamentally, the differences between the two are subtle: records are heterogenous maps over a fixed domain of string keys, and Maps are homogenous maps over an unknown set of string keys.
2. Related to (1), Maps add sharp corners the language, outlined in this RFC.  For example, there is no way to write a Map literal without having to also add type casts to the language.   
3. Maps are analyzable by reduction to the theory of arrays.  But it's not clear how well such an analysis would scale, especially in combination with all the other theories that Cedar analysis requires. 

So, on the balance, I vote to reject this proposal for the time being.

emina:
Thinking about this more, here is a not-fully-fleshed-out alternative idea for adding map-like functionality to Cedar, without introducing a new type.  This idea should help with concerns (1) and (2) in the [above comment](https://github.com/cedar-policy/rfcs/pull/27#issuecomment-1750970745), though (3) will remain a challenge.

Extend the record type so that a record can behave as _closed_ or _open_.  Closed records behave exactly as they do today with respect to syntax, semantics, validation, and analysis.

But open records have an additional feature.  In addition to a pre-defined set of closed attributes, their schema also declares a _default_ type $T$.  For all strings $f$ that are _not_ declared as attributes, an instance of an open record implicitly has an optional attribute $f$ of type $T$. 

With this alternative, we're just generalizing our current record type rather than introducing a new one.  The syntax of accessing and creating records etc. remains the same.  Validation does need to change to account for  these implicit attributes and so does the analysis.  But overall, the change is less invasive.

We can define what it means to have dynamic attribute access on such records, but this is not urgent.  All examples we've seen so far have used static (known) attribute names to access record values.   If needed, we can add dynamic access in the future.   Again, this is analyzable though there will be scaling limitations.

mwhicks1:
If keys are always static (i.e., they appear as literals in the policies), then I don't think we need to make any change and we can cancel this RFC. As I wrote in Workarounds, with static keys you can model a map as a record with each key as an optional attribute, and just enumerate those attributes/keys in the schema. The only reason you couldn't do this would be if policies are being generated on the fly but you could not update the schema to match them. Emina's open record proposal would help a little here, but this is a narrow use-case I think, and there are potential technical problems, see below.

To me, the only reason to add maps is if keys are dynamic. You can get dynamic keys with the set-of-records encoding, but you lack some flexibility, as I wrote above. I'd like to hear more from @max2me about whether there are applications that can't use the set-based encoding I gave above. If there are, then we might think more.

Back to open records: The problem here is that they do not play well with width subtyping (permissive mode). With width subtyping, you can do `{ foo: String, bar: String }` <: `{ foo: String }`, i.e., you can treat a record with two attributes as if it had just one. Now suppose that the record is open. That means you basically have a rule `{ foo: String } <: { foo: String, bar: T }` for the default type `T`. This rule is unsound in combination with the first rule: stringing them together, we make it so that a field with a `String` is treated as it had a `T` instead. I wrote about this in the RFC and pointed out that subtyping a record to a map can only work if the record is _closed_, so we are sure it doesn't have any non-listed fields (i.e., it's type is complete). I'm sure we could figure something out, but it might be kludgey, and that speaks against doing this too.

emina:
Ah, I missed the part about permissive validation.  But this should be fine for strict validation because there is no subtyping, right?

In any case, I agree that this is not ideal either.  I'm still for maintaining the status quo.

mwhicks1:
I am going to close this RFC. I believe that optional keys in records, or sets of key/value records along with `any` / `all`, should get the job done sufficiently well. We can re-open this RFC if a use-case emerges that can't be handled in either of these ways (or some other one) and the downsides to analysis can be overcome.

mwhicks1:
Reopening because the reason that we closed this RFC no longer applies: We are no longer proposing a sufficiently general `any?`/`all?` feature.

PR #33, opened 2023-10-13
Support version annotations on Cedar inputs

mwhicks1:
Support version annotations on Cedar inputs.

[Rendered](https://github.com/cedar-policy/rfcs/blob/input-versions/text/0033-input-versions.md)

### DIFF
--- a/text/0033-input-versions.md
+++ b/text/0033-input-versions.md
@@ -167,6 +167,8 @@ Adding an annotation adds some maintenance burden on customers. When a version a
 
 If an input format changes in a way that older syntax is interpreted differently, such as the `appliesTo` in the JSON schema as per the request in [issue 351](https://github.com/cedar-policy/cedar/issues/351), the fact that a version annotation is optional presents a problem: A newer and older parser could both accept the same input, but interpret it differently. If a version annotation was required, such an ambiguous interpretation would not be possible.
 
+Providing language-version annotations in code files is unusual. Go permits [specifying the language minimum version in a module's metadata file](https://go.dev/ref/mod#go-mod-file-go), but not in the code itself.
+
 ## Alternatives
 
 ### Do nothing to Cedar; version info can appear in application metadata
@@ -183,9 +185,11 @@ Rather than add a version annotation in the Cedar input syntax directly, version
     }
 }
 ```
-Such annotations could then trigger different application actions, e.g., calling different versions of the Cedar evaluator.
+Such annotations could then trigger different application actions, e.g., calling different versions of the Cedar evaluator. As a real-world example, the [`go` directive](https://go.dev/ref/mod#go-mod-file-go) in Go-language module metadata is used to select which toolchain to run.
 
 Versioning is especially useful when the same syntax is interpreted differently in later versions. This situation should be (very) rare, though. When it happens, the newer parser can warn about the change in syntax, especially in the case of errors that would trigger due to the change.
+
+Some comments on earlier versions of this RFC suggested that versioning might be most naturally given within a single configuration file that applies to a group of schemas, entity data, policies, etc. rather than associated with individual files.
 
 ### Change, or drop, the policy annotation format
 
--- a/text/0033-input-versions.md
+++ b/text/0033-input-versions.md
@@ -17,7 +17,7 @@ Reserve a way to indicate the Cedar version of an input format, so that future v
 
 Cedar policies:
 ```
-__cedar::version("3.0");
+@cedar_version("3.0");
 permit(principal,action,resource) when {
     principal is User && resource.owner == principal
 };
@@ -25,7 +25,7 @@ permit(principal,action,resource) when {
 JSON Cedar entities:
 ```
 {
-    "__cedar::version": "2.4",
+    "@cedar_version": "2.4",
     "contents": [
         {
             "uid": { "__entity": { "type": "User", "id": "alice"} },
@@ -38,7 +38,7 @@ JSON Cedar entities:
 JSON Cedar schema:
 ```
 {
-    "__cedar::version": "2.4",
+    "@cedar_version": "2.4",
     "PhotoApp": {
         "entityTypes": {
             "User": {
@@ -53,7 +53,7 @@ JSON Cedar schema:
 ```
 [Custom Cedar schema](0024-schema-syntax.md):
 ```
-@__cedar::version("3.0");
+@cedar_version("3.0");
 entity UserGroup;
 entity User in [UserGroup];
 entity List {
@@ -65,9 +65,9 @@ entity List {
 
 Cedar will occasionally make breaking changes which invalidate prior Cedar input formats. For example [RFC 20](0020-unique-record-keys.md) requires that keys in records be unique, both in policies and in entities, whereas pre-RFC 20 they did not have to be. 
 
-This RFC proposes to define annotations that signal the associated version of a particular input format, so that future versions of Cedar (or tools/services that use Cedar) can handle the situation gracefully. 
+This RFC proposes to define annotations that signal the associated version of a particular input format, so that future versions of Cedar (or tools/services that use Cedar) can handle the situation gracefully.
 
-For example, by annotating a policy or JSON input file with version 2.4, the parser for Cedar version 3.0 can flag an input with a 2.4-only feature as no longer supported, rather than just failing to parse. Or, a service that uses Cedar can run multiple versions of the Cedar tools, selecting the version that matches the input version they are given.
+This RFC makes no specification about what "gracefully" means, but minimally it can be used to indicate that parse errors might be due to version format incompatibilities. For example, by annotating a policy or JSON input file with version 2.4, the parser for Cedar version 3.0 can flag an input with a 2.4-only feature as no longer supported, rather than just failing to parse. Or, a service that uses Cedar can run multiple versions of the Cedar tools, selecting the version that matches the input version they are given.
 
 ## Detailed design
 
@@ -89,7 +89,7 @@ If an input file is annotated with an ill-formed version string, the parser shou
 
 For Cedar policies and custom schema syntax, we propose adding a new top-level declaration:
 ```
-__cedar::version(XXX);
+@cedar_version(XXX);
 ```
 where `XXX` is a string with the version in it.
 
@@ -98,25 +98,25 @@ This declaration applies to all policies or schema elements that are parsed in t
 Here is the amended grammar for a policies file:
 ```
 File    := {Policy | Version}
-Version := '__cedar::version' '(' STRING ')' ';'
+Version := '@cedar_version' '(' STRING ')' ';'
 Policy  := {Annotation} Effect '(' Scope ')' {Conditions} ';'
 ...
 ```
 Here is the amended grammar for a custom schema:
 ```
 Schema    := {Namespace | Version}
-Version   := '__cedar::version' '(' STRING ')' ';'
+Version   := '@cedar_version' '(' STRING ')' ';'
 Namespace := ('namespace' Path '{' {Decl} '}') | {Decl}
 ...
 ```
-In both cases there is the semantic constraint that `VERSION` appears at most once.
+In both cases there is the semantic constraint that `Version` appears at most once.
 
 ### Entities JSON
 
 The Cedar entities JSON is now defined as an array of entity records. We propose to update it to be a record containing a version declaration followed by that array. In particular:
 ```
 {
-    "__cedar::version": XXX,
+    "@cedar_version": XXX,
     "contents": [
         ...
     ]
@@ -139,10 +139,10 @@ The Cedar schema JSON is now defined as a record whose elements are individual n
     }
 }
 ```
-We propose to extend this format so that one of the "namespace" elements can be `"__cedar::version"`, and it will be associated with the version string XXX, i.e.,
+We propose to extend this format so that one of the "namespace" elements can be `"@cedar_version"`, and it will be associated with the version string XXX, i.e.,
 ```
 {
-    "__cedar::version": XXX, 
+    "@cedar_version": XXX,
     "Name::space::one": {
         ...
     },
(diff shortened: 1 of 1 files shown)


text/0033-input-versions.md
```diff
@@ -0,0 +1,166 @@
...
+
+Each input file will be associated with a version string. This is the version the input was written for, e.g., `"2.4.1"` or `"3.0"`. It could be that newer Cedar versions are compatible with older formats, but it is up to newer parsers to know that. For example, a new version of Cedar will add the [`is` operator](0005-is-operator.md) to policies, but an earlier version without that operator may still parse properly. The newer parser can account for this when parsing.
+
+If an input file is annotated with an ill-formed version string, the parser should treat it as a version it does not know about, e.g., as if for a future version of Cedar. It can issue a warning and proceed. If an input file is version-annotated more than once, then the parser should signal an error.
+
+### Policies and Custom Schema
+
+For Cedar policies and custom schema syntax, we propose adding a new top-level declaration:
+```
+__cedar::version(XXX);
```
emina, line 92: While I like the idea of adding a language version specification to policies, this syntax doesn't look nice or consistent with the syntax within Cedar policies themselves.  To me at least, keeping this consistent with policy syntax is more important than keeping it consistent with JSON. 

How about generalizing the annotation mechanism so that it can apply globally, to a whole policy file, as well as locally, individual policies?  

To my eyes, it would much nicer and more consistent to write just `version("2.4.1");` or `@cedar_version("2.4.1");` at the top of the policy file.  

And we may eventually want to add other global annotations, which would be enabled by a general global annotation mechanism.

text/0033-input-versions.md
```diff
@@ -0,0 +1,166 @@
...
+- Reference Issues:
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2013-10-13
+
+## Summary
+
+Reserve a way to indicate the Cedar version of an input format, so that future versions of Cedar can recognize and gracefully handle earlier input formats.
```
yuana97, line 14: love it, lets us bump versions (including to those with breaking changes) without needing to refactor our codebase immediately

anwarmamat: We need to be clear about "gracefully handle earlier input formats". It can be as simple as returning an error message saying you are using an old version of the schema and need to update.

cdisselkoen (approved):
Strongly in favor of this as a concept.  Indifferent about some of the syntax choices.

text/0033-input-versions.md
cdisselkoen: > I understand that we use `__cedar` in JSON

We don't.  `__cedar` is only proposed to be reserved as a namespace in RFC 24.  It's not currently proposed to be reserved in any other context.  

I think the concern with reserving the `cedar` namespace (in RFC 24 or here) is backwards compatibility.  It's much more likely that existing users are using the `cedar` namespace than the `__cedar` one.

emina: Got it; thanks.

In that case, is there any way for us to assess how many users, if any, would be affected if we reserve `cedar`?

I think it would be a pity to give up on the more natural namespace prematurely, especially since we don't know for sure that ` __cedar` is not a breaking change as well.

aaronjeline (approved):

text/0033-input-versions.md
max2me: I wholeheartedly agree on aesthethics of the prefix. Strings starting with underscores are.. strange. `cedar` feels a lot more natural. 

~~We should also allow 🌲 emoji as a legitimate alternative to `cedar` as in `🌲::version("1.2.0")`~~ 😇

At this time I am not concerned about backward compatibility. Although we don't have data on this, it seems unlikely that significant number of customers would use either `cedar` for their own purposes. We can provide guidance as part of upgrade instructions.

max2me:
I'm, as well, very much in support of versioning as a concept. 

> Optional: No version annotation is required. If none is given, the Cedar parser will assume the most recent version.

I appreciate simplicity of annotations being optional. But I also see advantages of requiring version to be specified for all subsequent major releases, especially for schemas and entity slice. Cost is minimal but it eliminates guesswork that parsers and tools have to do. Policies are likely to be on the same version for a given set anyway so specifying them once somewhere makes more sense.

It also makes it easier and safer for users of 3rd party tools that generate schema/entity slice because they will know exactly which SDK version outputs are valid for.

> This is the version the input was written for, e.g., "2.4.1" or "3.0"

How likely are we to change formats between minor releases? The experience would be simpler if schemas/policies/slices specified major version only `@cedar::version(3)` as customers wouldn't have to answer questions like 'My schema says its a version `2.3.1` but I'm going to upgrade to `2.4.1` SDK, do I need to make changes to the version?'

Downsides of requiring major versions only are all additional constraints we'd have to abide by for the sake of simpler experience.

> Implementing this RFC will add some complexity to the various Cedar parsers. They will now have to accept inputs that have a version annotation, and those that do not.

I wonder if we should be opinionated about handling of mismatches of different versions and/or older versions? Right now it's left up to the tool developers. However, I can see the benefits of being more prescriptive and encouraging tools to reject input  when its major version that doesn't match SDK used as that would protect from unintended consequences.

text/0033-input-versions.md
mwhicks1: I'll try a poll in Slack.

mwhicks1:
> But I also see advantages of requiring version to be specified for all subsequent major releases

I didn't want to require that inputs always have a version annotation because version annotations did not always exist. For example, version 2.0 does not have version annotations. So, newer-version parsers that want to recognize version-2.0 inputs have to deal with them being unlabeled. I suppose you could do that by just rejecting them, but that sounds like a bad CX -- if you upgrade to 3.0 then none of your policies, entities, etc. will work, even if they are compatible with the format changes being made.

> How likely are we to change formats between minor releases? 

Adding a new language feature, with new syntax, can be done in a minor release. I don't think we should attempt to deal with format incompatibilities by limiting the form of the version string, or limiting when format changes can happen. We already have major/minor/patch releases for distinguishing between breaking and non-breaking changes, and I think we should just go with what it prescribes. More below.

> I wonder if we should be opinionated about handling of mismatches of different versions and/or older versions?

The RFC doesn't say what a parser should do if it sees an older-version label on an input. It could do a few things:

1. Reject the input straightaway
2. Try to parse the input only if the old version is known to be compatible with the current version
3. Try to parse the input regardless of version, with the caveat that all parses in the new version would produce the same AST/meaning as the labeled version. Reject outright, if not. (An example here is the proposed change to `appliesTo` -- the new version and old version will parse the same files, but the meaning of the files would be different.) If parsing can go ahead, but fails, point out the version difference in the error message and whether we expect the versions to be compatible.

I lean toward (3) because it provides a safe, pleasant CX. This option is also compatible with making version annotations optional. In that case, you will basically be doing (3).

text/0033-input-versions.md
mwhicks1: I disagree with `__cedar` being "unnatural" as a reserved namespace item. Macro names starting with `__` are reserved in C and C++, for example. If I saw `__cedar` in a policy, I would immediately think of it as reserved/special, and wouldn't be surprised by the syntax.

I take your point that `java` in `java.lang.Integer` is also "reserved" effectively. However, it hasn't been reserved to this point, and I'm a little concerned about someone already using it. I would be surprised if someone was using it, but I'd be more surprised if someone was using `__cedar`.

Are you also proposing to change RFC 24 so that instead of using `__cedar` for a namespace there, we should use `cedar` instead? Recall:

we propose reserving the namespace `__cedar` and allowing it to be used as a prefix for both primitive and extension types. This way, an user-defined entity type `ipaddr` can be referred to directly, and the extension type `ipaddr` can be referred to as `__cedar::ipaddr`. Likewise, this allows a user to (say) define an entity type `String` that can be be referenced distinctly from `__cedar::String` (though this practice should be discouraged!).

I don't like the idea of using `cedar` in one/some places, and `__cedar` in others. To me it feels better to always use one or the other.

> I understand that we use __cedar in JSON, and that's fine.

Why is the use of `__cedar` fine in the JSON? Why not use `cedar::version` there too?

max2me:
> I didn't want to require that inputs always have a version annotation because version annotations did not always exist. For example, version 2.0 does not have version annotations. ...

What I was thinking is that if version is not supplied -- then parsers treat it as v2.x.x. If it's there, then parsers use the version specified. Whether it will be rejected or not -- that's up to individual parser that it supports. It does mean that if you want to use new features from v3, for example, you have to supply explicit version. But if you don't -- everything continues to work as it did in v2.

> Adding a new language feature, with new syntax, can be done in a minor release. 

Makes sense. If new language features can be introduced as part of minor release then yeah, we have to have full version.

> The RFC doesn't say what a parser should do if it sees an older-version label on an input. It could do a few things...

(3) does feel like a very graceful way of handling it.

mwhicks1:
> if version is not supplied -- then parsers treat it as v2.x.x. 

I see. That makes sense.

However, I still prefer the RFC as specified. There are two reasons:
1. I can imagine digesting policies for which a version annotation in the syntax would be inconvenient. E.g., perhaps I am reading an input that is intended to be a single policy. With a required version annotation, the input must be the version and the policy. This is just more complicated.
2. Explicit version identifiers are unusual, perhaps even nonexistent, in the concrete syntax of programming languages. Languages do have annotation mechanisms (e.g., `[# ...]` in Rust or `@XXX(YYY)` in Java) but no mechanism to say what version the language syntax is, in the syntax itself. So I'd like to support that expectation in Cedar.

I could imagine version information in the _metadata_ of a program. E.g., a JSON object with a `version` field and a `policy` field, where the former contains the version of the policy contained in the latter. This sort of arrangement would speak to parameterizing the actual parsing API functions with an optional, expected version ID. Should we do that?

text/0033-input-versions.md
mwhicks1: Changed proposed prefix to `@cedar_version` (see RFC for why this works).

glb:
@mwhicks1 you said

> Explicit version identifiers are unusual, perhaps even nonexistent, in the concrete syntax of programming languages. 

Is it too navel-gazey of me to ask whether you see Cedar (or at least the elements you're describing in this RFC) as a programming language or as a data representation format? Are schemas / entities / policies code or data? As a newbie outsider, I have had the mental model of these being data that the Cedar engine operates on rather than being elements of a programming language, despite the DSL nature of the typical (pre-JSON) Cedar policy representation.

There are lots of precedents<sup>citation needed</sup> for including versioning in data formats, and having the version embedded explicitly in the same file as the data feels like a really good way to make sure it's there. I would also echo @max2me 's comment about requiring the version information for the next major version, with some reasonable default during a transition period (as he suggested, assuming 2.x if not present).

There are also possibly precedents in the Go ecosystem as well, where the version of Go specified in the `go.mod` file affects toolchain behaviour when building the module ([reference](https://go.dev/ref/mod#go-mod-file-go)).

You also said

> I can imagine digesting policies for which a version annotation in the syntax would be inconvenient. E.g., perhaps I am reading an input that is intended to be a single policy.

Outside of toy / demo scenarios, is a single-policy input a common case or an exceptional one? I'm not familiar enough with Cedar in the wild to speculate usefully how others are using it, but picking on this particular example it seems to me that a single-policy input would be incredibly rare. Every application I've contemplated using Cedar with has had dozens to thousands of policies, so I'm curious about scenarios that could be handled with much-smaller policy sets or even a single policy.

emina:
> Is it too navel-gazey of me to ask whether you see Cedar (or at least the elements you're describing in this RFC) as a programming language or as a data representation format? Are schemas / entities / policies code or data?

It's a good question!

Cedar policies are code, and Cedar entities are data.  Cedar schemas are metadata (type annotations) about policies and code.  They don't affect runtime semantics (evaluation outcome), but they are useful for detecting and preventing errors (via validation).

max2me:
It would be good to get @hakanson perspective on this as he is the primary maintainer of vscode-cedar and has practical experience of adapting to new language features. 

I suspect that having explicit version annotations would make authoring tools like this easier as there would be a clear and deterministic way for which version of Cedar tool should use. At the same time, I'm leaning towards stating policy version in application metadata due to the number of policies/files with policies that a single project might have (this also sets up expectation that all policies should be of the same version).

hakanson:
As I look at my projects using Cedar,  I have:
* Cedar only projects with a schema and single policies file
* Projects with multiple folders of independent sets of schema, policies, and entities (some for testing, some as a monorepo support multiple Verified Permissions policy stores)
* AWS specific projects targeting Verified Permissions using CDK and constructing both static and template-linked policies in code.
* Test scripts that concatenate a set of "1 policy per .cedar file" into a single file containing all policies for CLI based unit testing

I mentally group the policies (code), entities (data), schema (metadata), and template links (data?) together as one logical construct where everything interrelates.  This makes a version per input item hard to reason about, especially when multiple policy files or even schema fragments are combined.  I would expect more a "application metadata" approach specifying a single version similar to the `go.mod` mentioned about or NodeJS `package.json` (with [`engines`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines)) or .NET `.csproj` (with `<TargetFramework>`) or others.  This combined code / data / metadata constructs also reminds of databases (DML, data, DDL) and SQL Server "[database compatibility level](https://learn.microsoft.com/en-us/sql/relational-databases/databases/view-or-change-the-compatibility-level-of-a-database?view=sql-server-ver16)".

It was mentioned in the RFC that "presence of a version annotation may imply more than this RFC is actually offering" is a big risk - with possible behavior ranging from just a "documentation comment" to a "fully backward compatible runtime."  This is where a single project-level version simplifies the approach.

What are thoughts on if the version number string support for pinning `=2.4.1`, minimum major version `^2.3`, or [ranges](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges) `[2.2, 3.0)` (meaning 2.2 ≤ x < 3.0) ?

In addition to version, an "application metadata" / config file can specify the file location / names of the schema file (instead of inline) and entities.  This would help tools like IDEs or the CLI to use this file to provide defaults to `--schema` parameters.  Definition of this file should be it's own RFC, but I vote TOML over JSON and have other idea here that help with unit testing.

FWIW, I liked `@cedar::version` better that `@cedar_version`

text/0033-input-versions.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Basic example
+
+Cedar policies:
+```
+@cedar_version("3.0");
```
aaronjeline, line 20: I did initially misread this as being an annotation on the policy, fwiw

aaronjeline (commented):
Strongly in favor, personally in favor of moving the syntax to `#version("1.1")`

mwhicks1:
Relevant comment to this RFC, raised in discussion elsewhere: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1381736348

text/0033-input-versions.md
```diff
@@ -0,0 +1,209 @@
...
+The use of `;` to distinguish per-policy annotations from file-wide policy annotations may be too subtle. Users could mistakenly write the following:
+```
+@cedar_version("2.4.0")
+permit(
+    principal == User::"Alice",
+    action == Action::"viewPhoto",
+    resource == Photo::"vacation97.jpg"
+);
+```
+They might think that the version applies to the whole file, but the lack of a semi-colon means the annotation just applies to the given policy.
```
khieta, line 201: I didn't notice the semicolon when this syntax was first presented, and assumed the version applied on a per-policy basis. So I was confused when the RFC said later that it applied to the whole file. If we decide to keep policy annotations, then I'd vote for a different syntax like `#cedar_version(...)`.

I like the idea of only using versions for entity data & schemas. But, like this RFC says, that implies that the policy syntax is always backwards compatible. Unfortunately, that's not the case due to [RFC20](https://github.com/cedar-policy/rfcs/blob/input-versions/text/0020-unique-record-keys.md).

text/0033-input-versions.md
```diff
@@ -0,0 +1,209 @@
...
+
+Adding an annotation adds some maintenance burden on customers. When a version annotation is present in a policy file, if the policies are changed to use a newer version of Cedar, the version annotation needs to be changed; it is not hard to forget to do so.
+
+If an input format changes in a way that older syntax is interpreted differently, such as the `appliesTo` in the JSON schema as per the request in [issue 351](https://github.com/cedar-policy/cedar/issues/351), the fact that a version annotation is optional presents a problem: A newer and older parser could both accept the same input, but interpret it differently. If a version annotation was required, such an ambiguous interpretation would not be possible.
+
+## Alternatives
+
+### Do nothing to Cedar; version info can appear in application metadata
+
+Rather than add a version annotation in the Cedar input syntax directly, version information can just be stored in metadata. For example, an application could store Cedar schemas in a document database and use a format like
```
khieta, line 174: @hakanson raises a good point that for a single application, the policies, entities, schema, and template links can be viewed as a cohesive unit that should all use the same version. This suggests that we might want a "config file" for Cedar applications that specifies the version, rather than syntax to add a version to each file independently.

khieta (approved):
I'm a fan of this RFC in general, although I'd prefer a different syntax for policy versions.

text/0033-input-versions.md
max2me: I've been trying to figure out why am I in favor of including version into entity data and schemas but opposed to doing the same for policies. I have a hunch:

* In most products built on Cedar schema and entity data would be represented as single 'document'. Entire schema is a single JSON file and same goes for request data.
* On the other hand, policies are going to be split across multiple documents in which case we either need to annotate each policy (not practical) or have version of the policy set stored in the config somewhere. 

Each product tailors policy management to their own unique need and having inline annotation of the version for policies seems to be less ideal in that case.

emina: If users manage their own policy versioning, why would they not want to manage their schema and entity versioning too?

I like @khieta's solution from [this comment](https://github.com/cedar-policy/rfcs/pull/33#discussion_r1381889216), which is to put all versioning information into a single configuration file, rather than spread it across many files (schemas, entity data, policies, etc).

text/0033-input-versions.md
```diff
@@ -0,0 +1,209 @@
...
+Here is the amended grammar for a custom schema:
+```
+Schema    := {Namespace | Version}
+Version   := '@cedar_version' '(' STRING ')' ';'
+Namespace := ('namespace' Path '{' {Decl} '}') | {Decl}
+...
+```
+In both cases there is the semantic constraint that `Version` appears at most once.
+
+### Entities JSON
```
cdisselkoen, line 114: Not mentioned in this RFC: if we have explicit version markers for the entities JSON format, would we also have/need explicit version markers for entities constructed programmatically (without JSON)?  I would say this RFC should take a stance on this (yes/no), and if yes, then at least sketch how the APIs would change.

This might be an additional argument in support of some kind of global configuration rather than per-file or more granular version markers (as discussed in other threads, e.g., [this one](https://github.com/cedar-policy/rfcs/pull/33#discussion_r1381889216), [this one](https://github.com/cedar-policy/rfcs/pull/33#discussion_r1381930212), or [this one](https://github.com/cedar-policy/rfcs/pull/33#issuecomment-1773893005)).

mwhicks1:
After some discussion today, the Cedar team feels that this RFC as written should **not** be accepted. Consuming version syntax adds complexity to the Cedar tools, especially if versions are to be _interpreted_ by those tools (e.g., carrying out parsing differently based on version).

With the status quo, Cedar users can specify version metadata on containers of Cedar policies, schemas, entities, etc. and interpret that metadata as best suits their needs, e.g. (as with Go), using it to select the right version of the Cedar toolchain to use.

The _motivation_ for this RFC still needs to be addressed. In particular, developers who might want to upgrade Cedar need an easy way to know if any changes to the language/schema/entity syntax or semantics affect their existing Cedar artifacts. The version syntax as proposed here would not help them -- at best, it would simply cause their application to start denying requests due to a policy/schema/entity version being out-of-date. To address poor customer experience due to language/schema changes, we must take care to

1. Not make breaking changes to the language/schema/entity syntax/semantics except in very rare cases, and only after ensuring customer impact is minimal.
2. If a breaking change is implemented, the language team should provide tools with which users can easily test whether their existing policy/schema/entity files are affected, and ideally can propose automatic fixes.

philhassey:
I don't feel the need for this right now.  As in Go, if I wanted to mark a large group of cedar policies as being a certain version I'd do it in a separate file instead of having to add a version marker into each file.  

If there was a version marker in the file, it would probably do best as a comment at the top (similar to Go build tags). Having it be part of the actual cedar code makes it such that you almost have to parse the file itself to discover the version.

PR #34, opened 2023-10-24
Precomputed entity attributes

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/precomputed-entity-attributes/text/0034-precomputed-entity-attributes.md)

### DIFF
--- a/text/0034-precomputed-entity-attributes.md
+++ b/text/0034-precomputed-entity-attributes.md
@@ -1,0 +1,186 @@
+# Precomputed Entity Attributes
+
+## Related issues and PRs
+
+- This RFC is mostly directly taken from [RFC 28](https://github.com/cedar-policy/rfcs/pull/28);
+  it splits out the changes from RFC 28 that relate to precomputed entity
+  attributes, without the larger changes in RFC 28 regarding `EntityDataSource`.
+- This RFC is also the successor to [RFC 23](https://github.com/cedar-policy/rfcs/pull/23).
+
+## Timeline
+
+- Started: 2023-10-24 (predecessor [#28](https://github.com/cedar-policy/rfcs/pull/16) started on 2023-10-03, and predecessor [#23](https://github.com/cedar-policy/rfcs/pull/23) started on 2023-07-18)
+- Entered FCP:
+- Accepted:
+- Landed:
+
+## Summary
+
+Internally, Cedar entity objects will store their attribute values as
+precomputed `Value`s, rather than in `RestrictedExpression` form.
+This entails some minor breaking changes to the Cedar public API, but provides
+important efficiency gains.
+
+## Motivation
+
+1. Today, every `is_authorized()` request re-evaluates all of the attributes in the
+`Entities`; they are stored as restricted expressions and evaluated to `Value`
+on every call to `is_authorized()`. (See
+[cedar#227](https://github.com/cedar-policy/cedar/issues/227).)
+This is inefficient.
+
+2. Today, `Entities::evaluate()` is implemented on `main` and provides a way to
+mitigate this issue, but it is opt-in and we'd rather have it be the default
+behavior.
+Notably, as of this writing, `Entities::evaluate()` is not released yet, so we
+can make changes in this area without breaking our users.
+
+3. Today, `Entity::new()` requires callers to pass in attribute values as
+`RestrictedExpression`.
+For some callers, evaluating these `RestrictedExpression`s is an unnecessary
+source of runtime evaluation errors (and performance overhead).
+
+## Detailed design
+
+This RFC has three components:
+
+### Component 1: Store attributes as precomputed `Value`s
+
+First, we redefine `Entities` and `Entity` to hold attributes as `Value`s
+instead of as restricted expressions.
+This addresses the first point in [Motivation](#motivation).
+
+This change is breaking for the public API in two (small) ways:
+- `Entity::new()`, `Entities::from_json_*()`, and friends can take the same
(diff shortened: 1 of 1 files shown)


andrewmwells-amazon:
For Component 2 I slightly prefer the 2nd option (consolidating `EvalResult` and `Value`), but I'm fine with the RFC as-is.

andrewmwells-amazon (approved):

shaobo-he-aws (approved):

anwarmamat (approved):

text/0034-precomputed-entity-attributes.md
```diff
@@ -0,0 +1,186 @@
...
+return `EntitiesError` but not an evaluation error.
+- Conversely, `Entity::attr()` need not return an error anymore, as all of the
+attributes will already be evaluated. (If we are very concerned about backward
+compatibility, we could keep the current signature and just never return `Err`.
+Currently, this RFC is proposing we change the signature to not contain
+`Result`, since we're already making the related breaking change to
+`Entity::new()`.)
+
+Accepting these breaking changes allows us to give users the best-performing
+behavior by default.
```
mwhicks1, line 68: What will happen to customer code when they upgraded? I think they will get a type error, which they can easily fix, is that right?

text/0034-precomputed-entity-attributes.md
```diff
@@ -0,0 +1,186 @@
...
+- not expose `Value` and just use `EvalResult`, even in the new constructors.
+This provides maximum backwards compatibility, like the first option, and
+also avoids the possible confusion generated by the first option, but comes
+at a notable performance cost (due to the inefficient
+`EvalResult`-to-`Value` conversion required under the hood) and might
+generate a different kind of confusion as `EvalResult` is an awkward name in
+the context of these constructors.
+
+This RFC currently proposes we take the third option.
+
```
mwhicks1, line 108: Same question as above: What happens when you try to compile existing code, which will use `EvalResult` to use the new API? I think you will get a type error that is easily fixed?

text/0034-precomputed-entity-attributes.md
```diff
@@ -0,0 +1,186 @@
...
+In addition to everything proposed in the current RFC, we could provide the
+status-quo definitions of `Entities` and `Entity` under new names: e.g.,
+`UnevaluatedEntities` and `UnevaluatedEntity` (naming of course TBD).
+This is very unlikely to be of any use to new users, but existing users really
+wanting the old error behavior could use these types and functions instead of
+`Entities` and `Entity`.
+
+To me, this provides little benefit, at the expense of more confusion and more
+APIs to maintain. It also doesn't make any of the changes more back-compatible;
```
mwhicks1, line 160: +1

text/0034-precomputed-entity-attributes.md
```diff
@@ -0,0 +1,186 @@
...
+We could do only first and third components of this RFC, and not the second
+component, at least for now.
+(The second component -- adding the new constructors for `Entities` and `Entity`
+-- is severable and could be its own RFC, or perhaps a change that doesn't rise
+to the level of requiring an RFC.)
+
+In essence, we could decide that the second component is more trouble than it's
+worth -- that it provides only marginal benefit to users at the cost of
+cluttering our API, both with additional constructors, and with the public
+`Value` type.
```
mwhicks1, line 138: I tend to agree with this assessment.

mwhicks1 (commented):
I'm leaning toward Alternative A, but I'm curious about your response to my questions.

text/0034-precomputed-entity-attributes.md
cdisselkoen: Yes, users will get type errors (or other compile errors) for all the breaking changes contemplated in this RFC.  As far as "easily fix", existing users of `EvalResult` (if there are any) will have the most involved fixes, as the underlying structure of `EvalResult` will have changed (if we take the third option in Component 2, as currently proposed).  The fixes for the changes in Component 1 should be easy for users. (and Component 3 is not a breaking change for any users on a released Cedar version -- and even if someone is on a pre-release version of Cedar and is using `.evaluate()`, they simply have to remove it.)

text/0034-precomputed-entity-attributes.md
cdisselkoen: See my answer above.  You will get a compile error and have to adapt to the new structure, depending on how you are using the `EvalResult`.  I don't know if anyone is actually using `EvalResult`, and if so, how they are using it.  If they are just printing the result to a string/console (as the CLI does), then the fix is minimal or zero, as the new version will be equally formattable with the Rust `Display` trait.

text/0034-precomputed-entity-attributes.md
cdisselkoen: Updated this RFC to move the second component to Alternative A (and adopt the previous Alternative A as the main proposal).  See my comment below.

text/0034-precomputed-entity-attributes.md
cdisselkoen: Resolving this thread now that this component is no longer in the main proposal.

text/0034-precomputed-entity-attributes.md
cdisselkoen: Now that the RFC has been updated to remove Component 2 (leaving it in Alternative A), the only breaking changes that remain are the ones in Component 1, which should be easy for users (and yes, surface as type errors or other compile errors).

mwhicks1 (approved):
LGTM

khieta (approved):

PR #35, opened 2023-10-25
Change the semantics of `appliesTo` fields in the schema

khieta:
[Rendered](https://github.com/cedar-policy/rfcs/blob/khieta/schema-applies-to/text/0035-schema-applies-to-unspecified.md)

### DIFF
--- a/text/0035-schema-applies-to-unspecified.md
+++ b/text/0035-schema-applies-to-unspecified.md
@@ -40,16 +40,16 @@ This feature request is consistent with the semantics for `appliesTo` proposed i
 
 ## Detailed design
 
-This RFC will require updating the schema parser to enforce suggestions 1-3 above.
+This RFC will require updating the schema parser to enforce proposals 1-3 above.
 It requires no changes to the validator itself.
 
 ### Notes on Backwards Compatibility
 
 This RFC proposes a **breaking change** that requires a major version bump to Cedar.
-For the sake of argument, say that the current version of Cedar X and that this RFC will be implemented in Cedar Y (which has no other breaking changes).
+For the sake of argument, say that the current version of Cedar is X and that this RFC will be implemented in Y (which has no other breaking changes).
 The rest of this section is written for users who are upgrading from X to Y.
 
-The changes proposed in this RFC will _only_ affect you use the following in your schema:
+The changes proposed in this RFC will _only_ affect you if you use the following in your schema:
 
 1. You intend for an action to apply to _no_ principals or resources, i.e.,
 
@@ -78,9 +78,45 @@ This will have the same meaning in both X and Y: the action applies to an unspec
 
 ## Drawbacks
 
-This is a breaking change that will alter the meaning of current schemas, see the discussion on backwards compatibility [above](#notes-on-backwards-compatibility).
-Although this change may cause developer pain in the near term, we hope that it will make schemas easier to understand in the long term.
+This is a breaking change that will alter the meaning of current schemas.
+See the discussion on backwards compatibility [above](#notes-on-backwards-compatibility) and discussion on backwards-compatible alternatives [below](#alternatives).
 
 ## Alternatives
 
-- Keep the semantics as is, and improve documentation to reduce confusion.
+### Alternative A
+
+The main downside of the current proposal is that, in some cases (see **case 1** [above](#notes-on-backwards-compatibility)), there will be no way to update your schema before you upgrade to Cedar Y -- you will have to update your schema simultaneously with the upgrade.
+This will be a problem for large applications where updating policies, schemas, and the Cedar SDK cannot be done atomically.
+
+We could mitigate this by instead implementing the following changes.
+
+| Scenario | Status Quo | Proposal |
+| --------- | -------- | ------- |
+| Specified `appliesTo` but omitted `principalTypes`/`resourceTypes` | Authz request with this action can only have an unspecified entity as the principal/resource | Same as status quo |
+| `appliesTo.principalTypes: []` AND `appliesTo.resourceTypes: []` | Action cannot be used in authz request | Same as status quo |
+| `appliesTo.principalTypes: []` OR `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
+| Omitted `appliesTo` | Authz request with this action can only have unspecified entities as the principal and resource | ❌ Not allowed |
+
+These changes remove some of the confusing behavior of the current implementation (e.g., setting only one of `appliesTo.principalTypes`/`appliesTo.resourceTypes` to be the empty list), while maintaining backwards compatibility.
+Under this alternative, some schemas will produce parse errors after upgrading to Y, but users can fix these error proactively.
+For every valid schema in Y, there is some schema in X that has identical behavior.
+
+The downside of this alternative is that it still allows the empty list to mean that an action applies to no entities, which may be confusing for users.
+Specifying that an action applies to no entities also requires more characters with this alternative, possibly making it more annoying to type out.
+
+This alternative will require updates to [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) in order to maintain consistency between the two schema formats.
+
+### Alternative B
+
+Another way to address backwards compatibility concerns is to combine this change with an implementation of schema versioning (see [RFC 33](https://github.com/cedar-policy/rfcs/pull/33)), supporting both the old and new behavior for schema parsing (for some period of time), allowing users to choose which behavior they want by setting a flag.
+
+There are at least two reasonable options here:
+
+- Force Cedar Y schemas to include an explicit version marker saying version Y. Schemas with no version marker will be interpreted using the Cedar X behavior, even if you're using Cedar Y. This gives full backwards-compatibility: no one's schema changes behavior when upgrading from Cedar X to Cedar Y, unless/until you add a version marker saying version Y. However, this means the behavior proposed in this RFC won't be default, which seems undesirable.
+- Same thing, but interpret schemas with no version marker using the Cedar Y behavior. This is still a breaking change, because unmarked schemas will change behavior when upgrading from Cedar X to Cedar Y. But it solves the problem of "no way to write a schema so that validation behavior will be the same between X and Y." Users just add a "version X" marker to their schema, and then they can upgrade from Cedar X to Cedar Y without that schema changing behavior.
+
+The primary arguments against this alternative are that (1) it blocks this RFC on the acceptance of RFC 33, and (2) it commits us to maintaining multiple versions of the schema parser and sets a precedent for supporting deprecated data formats going forward.
+
+## Unresolved questions
+
+- Should we provide an automated transformation for people using the existing schemas? How should we provide this transformation: as a new API in `cedar-policy`, or a standalone executable? How long should we maintain the transformation code? How can we ensure that users run the transformation at upgrade time?
--- a/text/0035-schema-applies-to-unspecified.md
+++ b/text/0035-schema-applies-to-unspecified.md
@@ -41,7 +41,7 @@ Under this proposal, some schemas will produce parse errors when users upgrade 
 
 ## Detailed design
 
-This RFC will requires updating the schema parser.
+This RFC requires updating the schema parser.
 It requires no changes to the validator itself.
 
 ## Drawbacks
--- a/text/0035-schema-applies-to-unspecified.md
+++ b/text/0035-schema-applies-to-unspecified.md
@@ -13,14 +13,14 @@
 
 | Scenario | Status Quo | Proposal |
 | --------- | -------- | ------- |
-| Specified `appliesTo` but omitted `principalTypes`/`resourceTypes` | Authz request with this action can only have an unspecified entity as the principal/resource | Same as status quo |
-| Omitted `appliesTo` | Authz request with this action can only have unspecified entities as the principal and resource | Action cannot be used in authz request |
-| `appliesTo.principalTypes: []` / `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
-| `appliesTo` is defined but all sub-elements are omitted | Authz request with this action can only have unspecified entities as the principal and resource; context is an empty record | ❌ Not allowed |
+| Specified `appliesTo` but omitted `principalTypes` and/or `resourceTypes` | Authz request with this action can only have an unspecified entity as the principal and/or resource | Same as status quo |
+| `appliesTo.principalTypes: []` AND `appliesTo.resourceTypes: []` | Action cannot be used in authz request | Same as status quo |
+| `appliesTo.principalTypes: []` OR `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
+| Omitted `appliesTo` | Authz request with this action can only have unspecified entities as the principal and resource | ❌ Not allowed |
 
 ## Motivation
 
-In Cedar 2.x, if an action's `appliesTo.principalTypes` (or `appliesTo.resourceTypes`) is not given, then it's interpreted as an action that applies to only unspecified principals (or resources).
+In Cedar 2.x and 3.0, if an action's `appliesTo.principalTypes` (or `appliesTo.resourceTypes`) is not given, then it's interpreted as an action that applies to only unspecified principals (or resources).
 If an action's `appliesTo` is not given, then it's interpreted as an action that applies only to unspecified principals _and_ resources.
 An _unspecified_ entity has a type that is distinct from any defined type.
 It is created by passing the `None` option in the principal and/or resource component of a [`Request`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Request.html).
@@ -29,27 +29,101 @@ If an action's `appliesTo.principalTypes` or `appliesTo.resourceTypes` are the e
 This takes precedent over any other information in the `appliesTo`.
 For example: if `appliesTo.principalTypes` is the empty list, then the validator expects the action to never apply, regardless of whether `appliesTo.resourceTypes` is omitted or set to a particular list of entity types.
 
-The difference between omitting a field vs. setting it to the empty list has caused confusion for users.
+The difference between omitting a field vs. setting it to the empty list, and the relationship between the two, has caused confusion for users.
 This RFC proposes to:
+
+1. Require that either both `appliesTo.principalTypes` and `appliesTo.resourceTypes` are empty lists, or neither are.
+2. Disallow a missing `appliesTo` field.
+
+These changes remove some of the confusing behavior of the current implementation while maintaining backwards compatibility.
+In particular, with this change there is only _one way_ to define an action group (both `appliesTo.principalTypes` and `appliesTo.resourceTypes` must be empty), and _one way_ to define an action where both the principal and resource are unspecified (omit the `appliesTo.principalTypes` and `appliesTo.resourceTypes` fields).
+Under this proposal, some schemas will produce parse errors when users upgrade to a new version of Cedar, but users can fix these error proactively because for every valid schema under the new proposal, there is some schema in current implementation that has identical behavior.
+
+## Detailed design
+
+This RFC will requires updating the schema parser.
+It requires no changes to the validator itself.
+
+## Drawbacks
+
+### This proposal is a compromise
+
+This RFC originally proposed Alternative A [below](#alternative-a).
+That alternative completely disallows empty lists in the `appliesTo` field, which have been a source of confusion.
+The current proposal still allows empty lists, but at least removes some cases where the behavior is confusing (e.g., when one of the `appliesTo` fields is an empty list, but not the other).
+Unfortunately, that alternative causes problems for users upgrading between versions of Cedar -- see discussion in the section on Alternative A.
+
+### This proposal is inconsistent with RFC24
+
+This proposal is inconsistent with [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md), where an omitted `appliesTo` field is used to define an action group, and empty `appliesTo.principalTypes` or `appliesTo.resourceTypes` lists are forbidden.
+In particular, this proposal is inconsistent with RFC24 in two cases:
+
+1. Declaring an action group `"actionGroup"`.
+
+JSON:
+
+```json
+"actionGroup": {
+    "appliesTo": {
+        "principalTypes": [],
+        "resourceTypes": []
+    }
+},
+```
+
+Natural syntax:
+
+```
+action actionGroup;
+```
+
+2. Declaring an action `"action"` that applies to an unspecified principal _and_ resource.
+
+JSON:
+
+```json
+"action": {
+    "appliesTo": {}
+},
+```
+
+Natural syntax:
+
+```
+action action
+  appliesTo { context: {} };
+```
+
+(Note that in the natural syntax, `context: {}` is required.)
+
+## Alternatives
+
+### Alternative A
+
+**NOTE:** this alternative was the original proposal
+
+| Scenario | Status Quo | Proposal |
+| --------- | -------- | ------- |
+| Specified `appliesTo` but omitted `principalTypes` and/or `resourceTypes` | Authz request with this action can only have an unspecified entity as the principal and/or resource | Same as status quo |
+| Omitted `appliesTo` | Authz request with this action can only have unspecified entities as the principal and resource | Action cannot be used in authz request |
+| `appliesTo.principalTypes: []` / `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
+| `appliesTo` is defined but all sub-elements are omitted | Authz request with this action can only have unspecified entities as the principal and resource; context is an empty record | ❌ Not allowed |
+
+This alternative proposes to:
 
 1. Change an omitted `appliesTo` to mean that the action cannot be used in a request.
 2. Disallow empty arrays for `appliesTo.principalTypes` / `appliesTo.resourceTypes`.
 3. Disallow an empty `appliesTo` attribute thus requiring at least one of `principalTypes`, `resourceTypes` or `context` to be specified if `appliesTo` is specified.
 
-This feature request is consistent with the semantics for `appliesTo` proposed in [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) for the alternate Cedar schema syntax.
+This semantics is consistent with the semantics for `appliesTo` proposed in [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) for the alternate Cedar schema syntax.
 
-## Detailed design
+The main downside of this alternative is that, in some cases (see **case 1** [below](#notes-on-backwards-compatibility)), there will be no way to update your schema before you upgrade Cedar versions -- you will have to update your schema simultaneously with the upgrade.
+This is a no-go for large applications where updating policies, schemas, and the Cedar SDK cannot be done atomically.
 
-This RFC will require updating the schema parser to enforce proposals 1-3 above.
-It requires no changes to the validator itself.
+#### Notes on Backwards Compatibility
 
-### Notes on Backwards Compatibility
-
-This RFC proposes a **breaking change** that requires a major version bump to Cedar.
-For the sake of argument, say that the current version of Cedar is X and that this RFC will be implemented in Y (which has no other breaking changes).
-The rest of this section is written for users who are upgrading from X to Y.
-
-The changes proposed in this RFC will _only_ affect you if you use the following in your schema:
+Say that the current version of Cedar is X and that this alternative is implemented in Y (which has no other breaking changes).
+When upgrading from X to Y, the changes proposed in this alternative will _only_ affect you if you use the following in your schema:
 
 1. You intend for an action to apply to _no_ principals or resources, i.e.,
 
@@ -76,36 +150,6 @@ In **case 2**, you should update your schema to use the following syntax instead
 
 This will have the same meaning in both X and Y: the action applies to an unspecified principal and resource, and the empty context.
 
-## Drawbacks
-
-This is a breaking change that will alter the meaning of current schemas.
-See the discussion on backwards compatibility [above](#notes-on-backwards-compatibility) and discussion on backwards-compatible alternatives [below](#alternatives).
-
-## Alternatives
-
-### Alternative A
-
-The main downside of the current proposal is that, in some cases (see **case 1** [above](#notes-on-backwards-compatibility)), there will be no way to update your schema before you upgrade to Cedar Y -- you will have to update your schema simultaneously with the upgrade.
-This will be a problem for large applications where updating policies, schemas, and the Cedar SDK cannot be done atomically.
-
-We could mitigate this by instead implementing the following changes.
-
-| Scenario | Status Quo | Proposal |
-| --------- | -------- | ------- |
-| Specified `appliesTo` but omitted `principalTypes`/`resourceTypes` | Authz request with this action can only have an unspecified entity as the principal/resource | Same as status quo |
-| `appliesTo.principalTypes: []` AND `appliesTo.resourceTypes: []` | Action cannot be used in authz request | Same as status quo |
-| `appliesTo.principalTypes: []` OR `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
-| Omitted `appliesTo` | Authz request with this action can only have unspecified entities as the principal and resource | ❌ Not allowed |
-
-These changes remove some of the confusing behavior of the current implementation (e.g., setting only one of `appliesTo.principalTypes`/`appliesTo.resourceTypes` to be the empty list), while maintaining backwards compatibility.
-Under this alternative, some schemas will produce parse errors after upgrading to Y, but users can fix these error proactively.
-For every valid schema in Y, there is some schema in X that has identical behavior.
-
-The downside of this alternative is that it still allows the empty list to mean that an action applies to no entities, which may be confusing for users.
-Specifying that an action applies to no entities also requires more characters with this alternative, possibly making it more annoying to type out.
-
-This alternative will require updates to [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) in order to maintain consistency between the two schema formats.
-
 ### Alternative B
 
 Another way to address backwards compatibility concerns is to combine this change with an implementation of schema versioning (see [RFC 33](https://github.com/cedar-policy/rfcs/pull/33)), supporting both the old and new behavior for schema parsing (for some period of time), allowing users to choose which behavior they want by setting a flag.
@@ -117,6 +161,7 @@ There are at least two reasonable options here:
 
 The primary arguments against this alternative are that (1) it blocks this RFC on the acceptance of RFC 33, and (2) it commits us to maintaining multiple versions of the schema parser and sets a precedent for supporting deprecated data formats going forward.
 
-## Unresolved questions
+## Outcomes of Discussion
 
-- Should we provide an automated transformation for people using the existing schemas? How should we provide this transformation: as a new API in `cedar-policy`, or a standalone executable? How long should we maintain the transformation code? How can we ensure that users run the transformation at upgrade time?
+- _Alternative A_ was rejected because of backwards-compatibility concerns.
+- _Alternative B_ was rejected because [RFC 33](https://github.com/cedar-policy/rfcs/pull/33) was rejected. See the discussion on the associated PR.
--- a/text/0035-schema-applies-to-unspecified.md
+++ b/text/0035-schema-applies-to-unspecified.md
@@ -37,7 +37,7 @@ This RFC proposes to:
 
 These changes remove some of the confusing behavior of the current implementation while maintaining backwards compatibility.
 In particular, with this change there is only _one way_ to define an action group (both `appliesTo.principalTypes` and `appliesTo.resourceTypes` must be empty), and _one way_ to define an action where both the principal and resource are unspecified (omit the `appliesTo.principalTypes` and `appliesTo.resourceTypes` fields).
-Under this proposal, some schemas will produce parse errors when users upgrade to a new version of Cedar, but users can fix these error proactively because for every valid schema under the new proposal, there is some schema in current implementation that has identical behavior.
+Under this proposal, some schemas will produce parse errors when users upgrade to a new version of Cedar, but users can fix these errors proactively because for every valid schema under the new proposal, there is some schema in the current implementation that has identical behavior.
 
 ## Detailed design
 
--- a/text/0035-schema-applies-to-unspecified.md
+++ b/text/0035-schema-applies-to-unspecified.md
@@ -21,16 +21,19 @@ This RFC proposes to change the semantics of the `appliesTo` field in the schema
 
 ## Motivation
 
-**Current behavior**  
+### Current behavior
+
 In Cedar 2.x, if an action's `appliesTo.principalTypes` or `appliesTo.resourceTypes` is not given (or if the entire `appliesTo` element is not given), then it's interpreted as an action that applies to all principal types and resource types.
 
-**Challenges**  
+### Challenges
+
 Discussion on [RFC 24](https://github.com/cedar-policy/rfcs/pull/24) highlighted challenges this introduces, primarily the risk of unintentionally specifying that an action applies to all principal types / resource types and related complications it causes for analysis and the experience of policy validation as error messages become more confusing.
 
-**Feature request**  
+### Feature request
+
 We can mitigate the challenges listed above by making following changes:
 
-1. Specified `appliesTo` but omitted `appliesTo.principalTypes` / `appliesTo.resourceTypes` means that the request component is unspecified, i.e., corresponding to the `None` option in the principal and/or resource component of a [`Request`](https://docs.rs/cedar-policy/2.4.0/cedar_policy/struct.Request.html). Omitted `appliesTo.context` means that the context is an empty record (same as the status quo).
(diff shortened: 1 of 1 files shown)


max2me:
Kesha @khieta, thank you for converting issue to the RFC, it looks great!

khieta:
@max2me Thanks for filing a clear issue that was easy to convert 😄

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Summary
+
+This RFC proposes to change the semantics of the `appliesTo` field in the schema, as summarized by the following table.
+| Scenario | Status Quo | Proposal |
+| --------- | -------- | ------- |
+| (1) Specified `appliesTo` but omitted `principalTypes`/`resourceTypes` | Authz request can have a principal/resource entity of any type, or the unspecified entity | Authz request with this action can only have unspecified entity as principal/resource |
+| (2) Omitted `appliesTo` | Authz request can have a principal/resource entity of any type, or the unspecified entity | Action cannot be used in authz request |
+| (3) `appliesTo.principalTypes: []` / `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
+| (4) `appliesTo` is defined but all sub-elements are omitted | Authz request can have a principal/resource entity of any type, or the unspecified entity; context is an empty record | ❌ Not allowed |
```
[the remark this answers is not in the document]
max2me: I think either option would achieve the goal of requiring customers to be explicit about principals and resources action applies to. 

Personally, I find idea of empty array indicating that unspecified principal/resource less intuitive but open to changing my mind about it. 

Another alternative to consider is introducing reserved entity type for unspecified principal/resource, something like `cedar::unspecified` (or whichever prefix is currently reserved). Having unspecified type be explicitly specified (no pun intended :) would help with clarity and comprehension. Thoughts?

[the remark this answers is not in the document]
mwhicks1: I prefer the current proposal, but I'm open to a change if many others are. Note that if we change it, we have to also change the custom syntax, which matches the proposal here.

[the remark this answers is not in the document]
khieta: Actually, I like the proposal to swap (1) and (3). Here's my reasoning:
* `principalTypes: []` / `resourceTypes: []` conveys that an action should not be used with any entity type defined in the schema. I think it's fair to interpret this as saying that the action _can_ be used with an entity type not in the schema (which is what an unspecified entity is).
* It feels a little cleaner to allow either `appliesTo` with both `principalTypes` and `resourceTypes` or none of the above.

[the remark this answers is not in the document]
mwhicks1: I personally don't like this because it's just more text to write. I'd prefer to omit the `principalTypes` line than to write `principalTypes: []`.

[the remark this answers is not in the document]
cdisselkoen: That is fair.  For folks actually using unspecified principal, `appliesTo { resourceTypes: [...] }` is probably clearer than `applies To { principalTypes: [], resourceTypes: [...] }`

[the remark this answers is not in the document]
mwhicks1: Right -- the principal type is literally unspecified. 😄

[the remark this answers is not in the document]
khieta: Ok, I've been convinced. Will leave the current proposal as-is.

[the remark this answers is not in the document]
Swolebrain: expressing the restriction of non-empty arrays is a little more challenging in the type system. While it's definitely possible to encode this in Rust's type system, I reckon it's going to be basically impossible to achieve automatic struct conversion from rust to typescript of this constraint. 

How do you plan on expressing the constraint that one of the keys (context, principalTypes, resourceTypes) must be set? For this one, i can't think of a way to do it at all other than imperatively. Why do we need this anyway? Why can't we assume {}, [], []? For this case, it also feels to me upon first thought that we will have a difficult if not impossible time achieving automatic type conversion to other types like ts.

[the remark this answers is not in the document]
mwhicks1: To clarify, the modified restrictions given here are on Cedar schemas, not on Rust code. In the `appliesTo` part of a schema, the attempt to set `principalTypes` to `[]` (say) would be rejected by the schema parser. There is no requirement to express the constraint that the corresponding Rust datastructure, after parsing, will always be nonempty. Likewise, there is no need to express, in the Rust type system, that other elements of `appliesTo` will be nonempty. That will be checked by the parser too, which populates the corresponding data structures.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,62 @@
...
+This RFC will require the following changes:
+
+1. The permissive validator must be updated to treat an omitted `appliesTo.principalTypes` / `appliesTo.resourceTypes` as the _unspecified_ entity type, rather than _any_ entity type. The difference is that the _unspecified_ entity type is a particular type that is different from any defined type, whereas _any_ entity type could be the unspecified type or any other type defined in the schema. Note that this change is only required in the _permissive_ validator, because the corresponding change was already made to strict validation as part of [RFC 19](https://github.com/cedar-policy/rfcs/blob/main/text/0019-stricter-validation.md).
+2. The validator must be updated to treat an omitted `appliesTo` attribute as indication that an action does not apply to any principal or resource types.
+3. The schema parser must be updated to reject empty arrays for `appliesTo.principalTypes` / `appliesTo.resourceTypes`.
+4. The schema parser must be updated to reject empty `appliesTo` attributes.
+
+## Drawbacks
+
+This is a breaking change that will alter the meaning of current schemas.
```
aaronjeline, line 51: Can we provide some kind of automated transformation for people using the existing schemas?

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,62 @@
...
+1. Specified `appliesTo` but omitted `appliesTo.principalTypes` / `appliesTo.resourceTypes` means that the request component is unspecified, i.e., corresponding to the `None` option in the principal and/or resource component of a [`Request`](https://docs.rs/cedar-policy/2.4.0/cedar_policy/struct.Request.html). Omitted `appliesTo.context` means that the context is an empty record (same as the status quo).
+2. Omitted `appliesTo` means that the action cannot be used in a request. It is used exclusively as an action group.
+3. Disallow empty arrays for `appliesTo.principalTypes` / `appliesTo.resourceTypes`.
+4. Disallow an empty `appliesTo` attribute thus requiring at least one of `principalTypes`, `resourceTypes` or `context` to be specified if `appliesTo` is specified.
+
+This feature request is consistent with the semantics for `appliesTo` proposed in [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) for the alternate Cedar schema syntax.
+
+## Detailed design
+
+This RFC will require the following changes:
```
john-h-kastner-aws, line 42: This RFC should also consider the semantics of `null` values. The schema parser currently represents this as `Option<Vec<_>>` which `serde_json` handles by deserializing both a `null` or an omitted value as `None`. This might not be what we want.

john-h-kastner-aws (approved):
The new semantics sound good.

text/0035-schema-applies-to-unspecified.md
mwhicks1: I don't think `null` values belong here, at least not at this late date as things are winding down. Let's make a separate RFC for that.

john-h-kastner-aws: The RFC was posted a week ago and isn't in FCP? Does not seem too late to add this

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. The permissive validator must be updated to treat an omitted `appliesTo.principalTypes` / `appliesTo.resourceTypes` as the _unspecified_ entity type, rather than _any_ entity type. The difference is that the _unspecified_ entity type is a particular type that is different from any defined type, whereas _any_ entity type could be the unspecified type or any other type defined in the schema. Note that this change is only required in the _permissive_ validator, because the corresponding change was already made to strict validation as part of [RFC 19](https://github.com/cedar-policy/rfcs/blob/main/text/0019-stricter-validation.md).
+2. The validator must be updated to treat an omitted `appliesTo` attribute as indication that an action does not apply to any principal or resource types.
+3. The schema parser must be updated to reject empty arrays for `appliesTo.principalTypes` / `appliesTo.resourceTypes`.
+4. The schema parser must be updated to reject empty `appliesTo` attributes.
+
+### Notes on Backwards Compatibility
+
```
khieta, line 53: Based on offline discussion with @Swolebrain I added this new section on backwards compatibility, describing in more detail how things will break if we implement this RFC. Please comment!

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,100 @@
...
+2. You intend for an action to apply to _any_ principal and/or resource types, i.e.,
+
+- omitted `appliesTo` field
+- omitted `appliesTo.principalTypes` field
+- omitted `appliesTo.resourceTypes` field
+
+In **case 1**, in preparation to upgrade, you will need to update your schema to omit the `appliesTo` field.
+Your current schema will result in error when you upgrade to Y.
+If you use your modified schema with X, you will be able to validate policies that will not validate in Y.
+There is _no way_ to write a schema so that validation behavior will be the same between X and Y.
```
khieta, line 74: It seems undesirable that there is no way to write a schema with the same meaning before & after this change. An easy fix would be to leave `principalTypes: []` and `resourceTypes: []` as the only way to specify that an action can apply to no entity types. This is arguably more intuitive too: rather than omitting a field (which otherwise means that something is unspecified) users have to explicitly say that an action should not be used in requests.

text/0035-schema-applies-to-unspecified.md
khieta: It's a little tricky, for reasons discussed in my recent commit adding a "Notes on Backwards Compatibility" section. TLDR: if a user has omitted an `appliesTo` field from an action declaration, then they may have intended for the action to apply to any defined entity or the unspecified entity. These two cases were previously conflated, and we'd need user input to decide which they intended.

But aside from that: how would we provide this automated transformation? As a new API in `cedar-policy`, or a standalone executable? How could ensure that users know to run the transformation?

text/0035-schema-applies-to-unspecified.md
khieta: Not too late 🙂 but I think @mwhicks1's point was that we want a separate RFC to decide on uniform treatment for `null` across all Cedar JSON representations, rather than just this RFC.

Personally, I think it makes sense to treat `null` values the same way as an omitted value.

text/0035-schema-applies-to-unspecified.md
khieta: More concretely, I would propose updating this RFC to the following:
| Scenario | Status Quo | Proposal |
| --------- | -------- | ------- |
| Specified `appliesTo` but omitted `principalTypes`/`resourceTypes` | Authz request can have a principal/resource entity of any type, or the unspecified entity | Authz request with this action can only have unspecified entity as principal/resource |
| `appliesTo.principalTypes: []` AND `appliesTo.resourceTypes: []` | Action cannot be used in authz request | Action cannot be used in authz request  |
| One of `appliesTo.principalTypes: []` OR `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed  |
| Omitted `appliesTo` | Authz request can have a principal/resource entity of any type, or the unspecified entity | ❌ Not allowed|

In all cases, if `context` is not specified in the `appliesTo` field, then it defaults to the empty record `{}`.

text/0035-schema-applies-to-unspecified.md
mwhicks1: >  it makes sense to treat null values the same way as an omitted value.

Agree. Let's put that it in its own RFC. At this rate, it will get approved faster than this one!

text/0035-schema-applies-to-unspecified.md
mwhicks1: > they may have intended for the action to apply to any defined entity or the unspecified entity.

Did we actually implement this previously? We might have suggested in the docs they could apply to any defined entity, but did validation actually consider each entity individually? I didn't think that it did, at least in strict mode; it basically just considered a hypothetical, unknown entity. @john-h-kastner-aws 

If I am right, then whatever the user intended, they were getting validation tantamount to the `unspecified` entity, so we don't really need to distinguish.

text/0035-schema-applies-to-unspecified.md
khieta: An alternative that also makes sense to me: combine the first and last rows of the table above, so that omitting `appliesTo` has the same meaning as omitting `principalTypes` and `resourceTypes`.

text/0035-schema-applies-to-unspecified.md
khieta: Yes, the latest released version of Cedar uses a hypothetical, unknown entity.

In many cases, this is the same as validating with an `unspecified` entity (e.g., it is not possible to access an entity's fields without a `has` check). But the `unspecified` type allows more precision, meaning that a policy may fail to validate (with an `ImpossiblePolicy` error) where it previously succeeded.

text/0035-schema-applies-to-unspecified.md
mwhicks1: For case 1: I'm confused. If you wrote `principalTypes: []` or `resourceTypes: []` before, in order to indicate that an action cannot be used in requests, then you should change your schema to remove the `appliesTo` clause instead. So you can still express in the proposal what you could express before. 

Moreover, it's much more obvious what you intend: With no `appliesTo` it's clear that this action is not "applyable" in a request. Part of the motivation for this proposal is to rule out nonsensical and/or confusing specs. @max2me specifically didn't like the idea that once you had `principalTypes: []` then it didn't matter what else you specified for the other bits -- you can't use this action in requests. The proposed change eliminates this confusion. I much prefer its clarity.

This change will be made evident by the validator. If you previously wrote `principalTypes: []` etc. then under this proposal you'll get an error. The fix is clear: remove the `appliesTo`. If you previously omitted `appliesTo` entirely but used the action in policies, then you'll get an error under this proposal, which should prompt you to add `appliesTo { }` instead.

For case 2: as I wrote in a comment above, I don't think we actually supported the idea that you could apply to _any_ type, in the sense that we specifically considered every type during validation.

text/0035-schema-applies-to-unspecified.md
max2me: > then they may have intended for the action to apply to any defined entity or the unspecified entity

Although this is possible, the more likely use case is actually reverse one -- they were creating an action group and intended it to not apply to any entity types.

> But aside from that: how would we provide this automated transformation? As a new API in cedar-policy, or a standalone executable? How could ensure that users know to run the transformation?

I don't have an opinion on standalone vs API but I do think this capability would be beneficial for all customers upgrading to latest Cedar. To make them aware, we'd call it out as part of release notes.

text/0035-schema-applies-to-unspecified.md
khieta: > If you wrote `principalTypes: []` or `resourceTypes: []` before, in order to indicate that an action cannot be used in requests, then you should change your schema to remove the `appliesTo` clause instead. So you can still express in the proposal what you could express before.

Yes, both proposals (the original RFC and my table above) allow you to express the same thing. The difference is that one is backwards compatible: it means the same thing in both before and after the proposed change. With the proposal in the original RFC, there is no way to update your schema to prepare for this change -- you have to update your schema simultaneously with the upgrade to the new version of Cedar.

> once you had `principalTypes: []` then it didn't matter what else you specified for the other bits

I also think that allowing only one of `principalTypes` / `resourceTypes` to be an empty list is confusing (since it means the same thing as having both be the empty list). I propose disallowing this above.

text/0035-schema-applies-to-unspecified.md
khieta: > Can you give me an example?

Consider this policy
```cedar
permit(principal == User::"alice", action == Action::"action", resource);
```

In the latest released version of Cedar, if `appliesTo` or `appliesTo.principalTypes` are omitted for `Action::"foo"`, then this policy will validate.

After implementing this RFC, the user could choose either to have the action apply to any defined type, or an unspecified entity. In the former case, the policy will continue to validate. In the latter case it will not because [RFC19](https://github.com/cedar-policy/rfcs/blob/main/text/0019-stricter-validation.md) will type `principal == User::"alice"` as `False`.

mwhicks1: > are omitted for Action::"foo", then this policy will validate.

Did you mean `Action::"action"` rather than `Action::"foo"` ?

> After implementing this RFC, the user could choose either to have the action apply to any defined principal type,

By "any defined principal type" you mean "any entity type", right? The validator cannot know which ones are intended to be used as principals. So if the user intended this behavior they'd have to list all the entity types explicitly, right?

I find it unlikely that any user intends a policy to actually apply to any entity type in a way that would not match treatment of the unspecified entity. It's a misfeature that the current schema treats a missing principal type as "unspecified and/or any other entity type". This RFC fixes that misfeature.

Perhaps we can write an analysis that warns users of policies like this and prompts them to make the appropriate change?

text/0035-schema-applies-to-unspecified.md
emina: Late to the conversation, but I prefer the original proposal given in the RFC.  Omitting `appliesTo` is clearer (to my eyes) and more ergonomic than having to set _both_ `principalTypes` and `resourceTypes` to the empty list. 

Since the change will be breaking no matter what, I'd vote for adopting the simplest long-term solution, even though it incurs more short-term inconvenience (having to update the schema simultaneously with the upgrade to the new version of Cedar).

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,100 @@
...
```
cdisselkoen, line 74: starting a different thread based on the same RFC text about backwards compatibility:

One way to mitigate some of these backwards compatibility problems would be to combine this with [RFC 33](https://github.com/cedar-policy/rfcs/pull/33).  For instance, at least two options:
* force Cedar Y schemas to include an explicit version marker saying version Y.  Schemas with no version marker will be interpreted using the Cedar X behavior, even if you're using Cedar Y.  This gives full backwards-compatibility: no one's schema changes behavior when upgrading from Cedar X to Cedar Y, unless/until you add a version marker saying version Y.  However, this means the behavior proposed in this RFC won't be default, which seems undesirable.
* same thing, but interpret schemas with no version marker using the Cedar Y behavior.  This is still a breaking change, because unmarked schemas will change behavior when upgrading from Cedar X to Cedar Y.  But it solves the problem of "no way to write a schema so that validation behavior will be the same between X and Y."  Users just add a "version X" marker to their schema, and then they can upgrade from Cedar X to Cedar Y without that schema changing behavior.

Of these two options, I prefer the latter because it makes this RFC's proposed behavior the default.  But both of these options of course come with the caveats/downsides discussed on RFC 33.

text/0035-schema-applies-to-unspecified.md
Swolebrain: One key thing to understand is that these updates are not atomic. People will have to update their policies and schemas to be forwards compatible, wait some non-zero time period (potentially weeks or months), then perform the update to cedar 3.0. 

In that interim time (during those few weeks/months) there is no path for customers to express their actual intent in the schema. This is because the same syntax means different things in both proposals.

text/0035-schema-applies-to-unspecified.md
Swolebrain: If someone out there was relying on schemas, and they were storing a creationDate and lastModifiedDate, they could use that information to decide which version to use in schema parsing and validation. If the lastModifiedDate < some cutoff, then use Cedar X. Otherwise use Cedar Y. I think craig is on to something here.

mwhicks1: Keep in mind that for RFC33 the original proposal was that later versions of the software will not (necessarily) accept earlier versions of the inputs. You are suggesting to that later software versions interpret earlier input versions ("Schemas with no version marker will be interpreted using the Cedar X behavior, even if you're using Cedar Y."). 

We didn't want to commit to doing this in general. If we are not going to do it, I realize that versions could just be stored as metadata in whatever way an application likes, rather than in policies/JSON/etc so that Cedar's tools have to specifically know about it. In particular, we could just add a general file-wide annotation mechanism that generalizes the per-policy mechanism we have now, and apps can use it to add version info however they like.

In sum: We should only add versioning info to Cedar itself if we are going to have later versions of tools/code interpret out-of-date versions of inputs (which Craig is proposing to do here). But agreeing to understand past input versions complicates development, testing, etc. so we should be really sure about what we are getting into.

anwarmamat: If we introduce another breaking change to the schema in Cedar Z, then how do we handle the schemas in versions X,Y, and Z? Can Y schema just delete the version Y so that it will be parsed with the Cedar Y? What will happen to the X schemas with no version marker?

cdisselkoen: > If we introduce another breaking change to the schema in Cedar Z

then Cedar Z would have to handles schemas from X, Y, and Z.  This is why @mwhicks1 calls this a slippery slope and says "we [don't] want to commit to doing this in general".

> Can Y schema just delete the version Y so that it will be parsed with Cedar Y?

I assume you're referring to the second bullet of my proposal.  If so, then yes, but it wouldn't be required.

> What will happen to the X schemas with no version marker?

in the first bullet of my proposal, they would not break.  in the second bullet of my proposal, they would.

cdisselkoen: (of course, they only break if they meet the conditions described in the RFC's backwards compatibility section.  Many schemas are not affected by RFC 35)

khieta: I'm 💯 on board with @cdisselkoen's second bullet. I also like the first bullet, but based on discussion on [RFC33](https://github.com/cedar-policy/rfcs/pull/33) I've been convinced that leaving version info optional is cleanest, which suggests that a schema without a version should be interpreted as the latest version, rather than always always defaulting to an old version (e.g., 2.4).

To @mwhicks1's point on the slipper slope: In most cases, I think we'd only want to use the version info from RFC33 to print better error messages (e.g., "version X is no longer supported, please follow the upgrade instructions at < some link >"). I think we should treat the current RFC (35), where we've changed the interpretation of the same syntax between versions, as a special case and provide support for parsing the old format for a period of time (e.g., until the next major release).

mwhicks1: > I think we should treat the current RFC (35), where we've changed the interpretation of the same syntax between versions, as a special case and provide support for parsing the old format for a period of time (e.g., until the next major release).

Surely this will not be the last time that a situation like this arises. So we are setting a precedent here.

emina: > Surely this will not be the last time that a situation like this arises. So we are setting a precedent here.

I share @mwhicks1's concern:  we're setting a precedent here.

khieta: I agree that this is not the last time this will happen, and we're setting a precedent.

What I meant was that I think it's ok to provide support for old versions in cases where we have made a breaking change like this -- but that we should set the precedent that the old version is not supported indefinitely. E.g., we would only support it until the next major version.

khieta:
⚠️ Major revisions in the latest commit! ⚠️ 

Offline discussion on backwards compatibility revealed that this RFC was more backwards compatible then we originally thought: omitting the `appliesTo` field (or subfields) means that an action applies to the unspecified entity (not any entity) in all released versions of Cedar. This means that the expression `principal == User::"alice"` will be typed as `False` (not `Bool`) in cases where principal is unspecified. Thanks @mwhicks1 for pointing this out.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,90 @@
...
+| Specified `appliesTo` but omitted `principalTypes`/`resourceTypes` | Authz request with this action can only have unspecified entity as principal/resource | Same as status quo |
+| Omitted `appliesTo` | Authz request with this action can only have unspecified entity as principal/resource | Action cannot be used in authz request |
+| `appliesTo.principalTypes: []` / `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
+| `appliesTo` is defined but all sub-elements are omitted | Authz request with this action can only have unspecified entity as principal/resource; context is an empty record | ❌ Not allowed |
+
+## Motivation
+
+### Current behavior
+
+In Cedar 2.x, if an action's `appliesTo.principalTypes` or `appliesTo.resourceTypes` is not given (or if the entire `appliesTo` element is not given), then it's interpreted as an action that applies to only unspecified principals and resources.
```
mwhicks1, line 25: I don't believe this is correct. If you omit `appliesTo.principalTypes` then just the `principal` component is unspecified. You can still specify `appliesTo.resourceTypes` to require a specific `resource` entity type in a request.

The table, above, also seems to make this mistake.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,90 @@
...
+An _unspecified_ entity has a type that is distinct from any defined type.
+It is created by passing the `None` option in the principal and/or resource component of a [`Request`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Request.html).
+If an action's `appliesTo.principalTypes` or `appliesTo.resourceTypes` fields are set to the empty list, then the action is not expected to be used with any request; it should only be used as a way to group other actions.
```
mwhicks1, line 28: Worth adding: if you indicate that `principalTypes` is `[]` then it rules out using the action in requests _even if you don't specify `resourceTypes` as `[]`. This is confusing, and part of the motivation of this RFC is to address this source of confusion. (You say this below, under challenges, but it is a little strange to me to introduce the issue there for the first time. The first sentence in that part is recapping what was written above.)

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,90 @@
...
+
+The changes proposed in this RFC will _only_ affect you use the following in your schema:
+
+1. You intend for an action to apply to _no_ principals or resources, i.e.,
+
+- `appliesTo.principalTypes: []`
+- `appliesTo.resourceTypes: []`
+
+2. You intend for an action to apply to _unspecified_ principals or resources, i.e.,
+
```
mwhicks1, line 64: principals _and_ resources (not "or").

mwhicks1 (commented):
Some nits.

text/0035-schema-applies-to-unspecified.md
khieta: Revised text. Lmk what you think

text/0035-schema-applies-to-unspecified.md
khieta: Revised text. Lmk what you think

text/0035-schema-applies-to-unspecified.md
khieta: Marking this thread as resolved since the majority opinion seems to be:
> Omitting appliesTo is clearer (to my eyes) and more ergonomic than having to set both principalTypes and resourceTypes to the empty list.

cdisselkoen: What is the resolution to the issue raised in this thread about

> It seems undesirable that there is no way to write a schema with the same meaning before & after this change

> With the proposal in the original RFC, there is no way to update your schema to prepare for this change -- you have to update your schema simultaneously with the upgrade to the new version of Cedar.

These seemed to be the motivating arguments for the other side, even though I agree that the majority feel that the RFC's current proposal is clearer and more ergonomic than having to set both `principalTypes` and `resourceTypes` to the empty list.

mwhicks1 (approved):
Looks great. I'm not bothered by the backward incompatibility problem given here but if that's a showstopper then I'll wait to see what the proposed alternative might be.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,122 @@
...
+| Omitted `appliesTo` | Authz request with this action can only have unspecified entities as the principal and resource | ❌ Not allowed |
+
+These changes remove some of the confusing behavior of the current implementation (e.g., setting only one of `appliesTo.principalTypes`/`appliesTo.resourceTypes` to be the empty list), while maintaining backwards compatibility.
+Under this alternative, some schemas will produce parse errors after upgrading to Y, but users can fix these error proactively.
+For every valid schema in Y, there is some schema in X that has identical behavior.
+
+The downside of this alternative is that it still allows the empty list to mean that an action applies to no entities, which may be confusing for users.
+Specifying that an action applies to no entities also requires more characters with this alternative, possibly making it more annoying to type out.
+
+This alternative will require updates to [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) in order to maintain consistency between the two schema formats.
```
khieta, line 107: Original discussion: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1380203955

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,122 @@
...
+
+- Force Cedar Y schemas to include an explicit version marker saying version Y. Schemas with no version marker will be interpreted using the Cedar X behavior, even if you're using Cedar Y. This gives full backwards-compatibility: no one's schema changes behavior when upgrading from Cedar X to Cedar Y, unless/until you add a version marker saying version Y. However, this means the behavior proposed in this RFC won't be default, which seems undesirable.
+- Same thing, but interpret schemas with no version marker using the Cedar Y behavior. This is still a breaking change, because unmarked schemas will change behavior when upgrading from Cedar X to Cedar Y. But it solves the problem of "no way to write a schema so that validation behavior will be the same between X and Y." Users just add a "version X" marker to their schema, and then they can upgrade from Cedar X to Cedar Y without that schema changing behavior.
+
+The primary arguments against this alternative are that (1) it blocks this RFC on the acceptance of RFC 33, and (2) it commits us to maintaining multiple versions of the schema parser and sets a precedent for supporting deprecated data formats going forward.
```
khieta, line 118: Original discussion: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1381593430

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Unresolved questions
+
+- Should we provide an automated transformation for people using the existing schemas? How should we provide this transformation: as a new API in `cedar-policy`, or a standalone executable? How long should we maintain the transformation code? How can we ensure that users run the transformation at upgrade time?
```
khieta, line 122: Original discussion: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1378832191

khieta (commented):
I've added two new alternatives + an unresolved question to summarize our discussions so far 🎉 Starting fresh threads for more targeted discussion.

text/0035-schema-applies-to-unspecified.md
khieta: Started a fresh thread to discuss the question of whether we should provide translation between schema versions: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1386795521

Also note that several of my statements above turned out to be incorrect 🤦‍♀️ See https://github.com/cedar-policy/rfcs/pull/35#issuecomment-1799533721

text/0035-schema-applies-to-unspecified.md
khieta: I'm going to resolve this thread because I think it would be better addressed in #37. Will leave comments there soon.

text/0035-schema-applies-to-unspecified.md
khieta: Good points Craig. I've updated the RFC text to explicitly list the proposed alternative, and started a fresh discussion thread: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1386794236

Resolving this thread to keep the discussion organized.

text/0035-schema-applies-to-unspecified.md
khieta: Updated the RFC text to include this alternative, and started a fresh thread for discussion: https://github.com/cedar-policy/rfcs/pull/35#discussion_r1386794836

Resolving this thread to keep the discussion organized.

text/0035-schema-applies-to-unspecified.md
khieta: After writing this up, I actually like alternative A a little more that the original proposal because it alleviates backwards compatibility issues. Between these two options:
```json
"action": {}
```
```json
"action": {
  "appliesTo": {
    "principalTypes": [],
    "resourceTypes": []
  }
}
```
I think that both adequately convey that `"action"` applies to no principals or resources.

text/0035-schema-applies-to-unspecified.md
khieta: My personal preference is that we not provide this. I'm worried that we're signing up for extra code to maintain that no one will use.

emina (approved):

shaobo-he-aws (approved):

text/0035-schema-applies-to-unspecified.md
mwhicks1: I agree that both convey that `"action"` applies to no principals or resources. 

But this alternative has two downsides:
- It requires that every action group is specified more verbosely. We have at least one customer that makes heavy use of action groups, and this proposal makes schemas less readable for this customer. 
- It does match the custom schema syntax, which does allow omitting `appliesTo` (precisely so that schemas are less verbose). 

I suppose we could accept this proposal and then in the next major version to relax the prohibition on omitted `appliesTo`, so that it matches the custom syntax. This would amount to a breaking change followed by a backward compatible change, so perhaps that's not so bad.

But I would prefer we get some data about whether anyone is currently using the omitted `appliesTo` to mean an action that applies to unspecified principals and resources. I have a feeling that this is not actually used, in which case we are accepting a weaker alternative to no good purpose.

khieta:
Update: re-starting discussion on this RFC 🎉 I've just pushed a major update that switches the proposal to what was originally "Alternative A". I feel that this is a more acceptable solution for users that currently have `appliesTo.principalTypes: []` and/or `appliesTo.resourceTypes: []` in their schemas.

Swolebrain:
I don't see why this is inconsistent with RFC24. Maybe there's something I'm missing but I don't see why RFC24 needs to map exactly 1:1 to json format. There could be syntactic sugar to it. For instance, **both** of the following:

* `action ActionGroup;` 
* `action actionGroup appliesTo { principalTypes: [], resourceTypes: [], context: {} };`

Could transpile to the same thing in json:

```
"actionGroup": {
    "appliesTo": {
        "principalTypes": [],
        "resourceTypes": [],
        "context": {}
    }
},
```

could they not?

khieta:
It's true that RFC24 doesn't need to map 1:1 to the JSON format: we can easily write a translator between the two syntaxes for the current proposal.

In RFC24, `action ActionGroup appliesTo { principalTypes: [], resourceTypes: [], context: {} };` would not be valid syntax, since empty lists are not allowed.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,167 @@
...
+
+- Started: 2023-10-25
+
+## Summary
+
+| Scenario | Status Quo | Proposal |
+| --------- | -------- | ------- |
+| Specified `appliesTo` but omitted `principalTypes` and/or `resourceTypes` | Authz request with this action can only have an unspecified entity as the principal and/or resource | Same as status quo |
+| `appliesTo.principalTypes: []` AND `appliesTo.resourceTypes: []` | Action cannot be used in authz request | Same as status quo |
+| `appliesTo.principalTypes: []` OR `appliesTo.resourceTypes: []` | Action cannot be used in authz request | ❌ Not allowed |
```
cdisselkoen, line 18: Based on the comment that we're concerned about

> users that currently have appliesTo.principalTypes: [] and/or appliesTo.resourceTypes: [] in their schemas

This proposal is backwards-compatible for users that have BOTH `principalTypes` and `resourceTypes` set to `[]`, but not for users that have only ONE of those set to `[]`, for a given action.

Are we concerned about that backwards-compatibility problem?  What makes us confident that we need to address backwards-compatibility for the AND case, but not for the OR case?

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,167 @@
...
+
+The difference between omitting a field vs. setting it to the empty list, and the relationship between the two, has caused confusion for users.
+This RFC proposes to:
+
+1. Require that either both `appliesTo.principalTypes` and `appliesTo.resourceTypes` are empty lists, or neither are.
+2. Disallow a missing `appliesTo` field.
+
+These changes remove some of the confusing behavior of the current implementation while maintaining backwards compatibility.
+In particular, with this change there is only _one way_ to define an action group (both `appliesTo.principalTypes` and `appliesTo.resourceTypes` must be empty), and _one way_ to define an action where both the principal and resource are unspecified (omit the `appliesTo.principalTypes` and `appliesTo.resourceTypes` fields).
+Under this proposal, some schemas will produce parse errors when users upgrade to a new version of Cedar, but users can fix these error proactively because for every valid schema under the new proposal, there is some schema in current implementation that has identical behavior.
```
cdisselkoen, line 40:
```suggestion
Under this proposal, some schemas will produce parse errors when users upgrade to a new version of Cedar, but users can fix these errors proactively because for every valid schema under the new proposal, there is some schema in the current implementation that has identical behavior.
```

cdisselkoen, line 40: I think we can state this stronger: not just

> for every valid schema under the new proposal, there is some schema in current implementation that has identical behavior

but more strongly,

> every valid schema under the new proposal has exactly the same behavior in the current implementation

Critically, in my reading, the current text is ambiguous whether it's just "some schema has that behavior" (which might mean, a different schema from the new valid one; which would still require changing your schema when upgrading Cedar in order to maintain the same semantics).  The new text is more clear that you get the same behavior from in fact the same schema, without having to change your schema when you upgrade Cedar.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,167 @@
...
+This takes precedent over any other information in the `appliesTo`.
+For example: if `appliesTo.principalTypes` is the empty list, then the validator expects the action to never apply, regardless of whether `appliesTo.resourceTypes` is omitted or set to a particular list of entity types.
+
+The difference between omitting a field vs. setting it to the empty list, and the relationship between the two, has caused confusion for users.
+This RFC proposes to:
+
+1. Require that either both `appliesTo.principalTypes` and `appliesTo.resourceTypes` are empty lists, or neither are.
+2. Disallow a missing `appliesTo` field.
+
+These changes remove some of the confusing behavior of the current implementation while maintaining backwards compatibility.
```
cdisselkoen, line 38: Instead of "the current implementation", we could more precisely state "Cedar 2.x and 3.0" or even "Cedar 2.x and 3.x" (here and in other places in this RFC).  The tradeoff is that as this RFC text lives to posterity, we may not keep updating this text with which Cedar versions have the described behavior.  Nonetheless, the additional clarity might be worth it.

text/0035-schema-applies-to-unspecified.md
```diff
@@ -0,0 +1,167 @@
...
+In particular, with this change there is only _one way_ to define an action group (both `appliesTo.principalTypes` and `appliesTo.resourceTypes` must be empty), and _one way_ to define an action where both the principal and resource are unspecified (omit the `appliesTo.principalTypes` and `appliesTo.resourceTypes` fields).
+Under this proposal, some schemas will produce parse errors when users upgrade to a new version of Cedar, but users can fix these error proactively because for every valid schema under the new proposal, there is some schema in current implementation that has identical behavior.
+
+## Detailed design
+
+This RFC will requires updating the schema parser.
```
cdisselkoen, line 44:
```suggestion
This RFC requires updating the schema parser.
```

cdisselkoen:
> I don't see why this is inconsistent with RFC24. Maybe there's something I'm missing but I don't see why RFC24 needs to map exactly 1:1 to json format.

I think we're all in agreement -- it is "inconsistent" which might be a poor UX / confusing for any folks dealing with both the JSON and RFC24 formats.  But "inconsistent" does not mean "incompatible", and we can easily translate between the two even if they do not use precisely the same syntax for some edge cases.

text/0035-schema-applies-to-unspecified.md
khieta: Yes, your read is correct.

text/0035-schema-applies-to-unspecified.md
khieta: In the case of `appliesTo.principalTypes: []` OR `appliesTo.resourceTypes: []`, users can proactively switch to `appliesTo.principalTypes: []` AND `appliesTo.resourceTypes: []`, and know that their schema will continue to behave as expected.

In Alternative A, the sharp edge is that in some cases there is no way for a user to proactively update their schema, because there is no way to write a schema with the same behavior before & after the change.

cdisselkoen: I see, makes sense.  It's not backwards-compatible in the OR case, but at least there is a reasonable migration story.

(Maybe we should make up a term for the kind of backwards-incompatible that's in Alternative A, it's like a strictly worse kind of backwards-incompatible than the kind here for the OR case.)

shaobo-he-aws (approved):

khieta:
After much debate, we've decided to abandon this RFC for the following reasons:
1. The current proposal is a breaking change without a large benefit: it just disallows up a corner case in the schema format. We'd prefer a less invasive change, such as returning a warning rather than failing to parse previously parseable inputs. Cedar doesn't currently allow warnings during schema construction, but may be extended with this in the future.
2. This RFC only affects the JSON syntax. In general, we will recommend that users use the new schema syntax from [RFC24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md), which will be released as part of Cedar v3.1.0.
3. The breaking change that we _really_ want is [RFC55](https://github.com/cedar-policy/rfcs/pull/55), which makes the current RFC irrelevant. There is still discussion to be had on whether this new breaking change is worth it, but that can be discussed on new RFC.

PR #36, opened 2023-11-01
RFC for unbounded integers

andrewmwells-amazon:
[Rendered](https://github.com/cedar-policy/rfcs/blob/andrewmwells/unbounded-ints/text/0036-unbounded-ints.md)

### DIFF
--- a/text/0036-unbounded-ints.md
+++ b/text/0036-unbounded-ints.md
@@ -1,0 +1,95 @@
+# Unbounded Integers
+
+## Related issues and PRs
+
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR(s): Not implemented, but an implementation was started for benchmarking: https://github.com/cedar-policy/cedar/pull/376
+
+## Timeline
+
+- Started: 2023-11-01
+
+## Summary
+
+Currently, Cedar represents integers using Rust's `i64`. This RFC proposes changing the internal representation to unbounded integers using the `IBig` crate.
+
+## Basic example
+
+Currently, `1,000,000,000,000,000,000 * ({"a" : 1,000,000,000,000,000,000}.a)` will overflow and result in an error. If this RFC is accepted, it would instead evaluate to `1,000,000,000,000,000,000,000,000,000,000,000,000`.
+
+## Motivation
+
+Bounded integer semantics (either wrapping or erroring on overflow) are unintuitive [https://blog.research.google/2006/06/extra-extra-read-all-about-it-nearly.html]. Consider the following policy:
+
+```
+forbid(principal, action, resource) when {
+   principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms ||
+   !principal.is_mfa
+};
+```
+
+A casual reader of this policy may conclude that all actions are forbiden when `principal.mfa` is `false`. This is wrong. The lhs (`principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms`) is evaluated first and if it errors the entire policy is skipped.
+
+To avoid this class of mistakes, we want to change Cedar so that arithmetic never errors.
+
+## Detailed design
+
+We replace the internal representation of integers as `i64`s with an unbounded representation (see draft https://github.com/cedar-policy/cedar/pull/376 which is incomplete but was used for benchmarking). Cedar would also remove the limitation on integers recieved as inputs (both integer literals in policies and attributes recieved from JSON or directly constructed).
+
+With these limits removed, Cedar could represent the result of e.g., `i64::MAX * i64::MAX` in the residual policies returned by partial evaluation.
+
+## Drawbacks
+
+### Performance:
+Using `i64`s is obviously faster than using unbounded integers. We benchmarked several crates, and the fastest we tested (under our licensing restrictions which forbids (L)GPL) was `IBig`. We evaluated policies that should represent the worst-case runtime increase from the switch.
+
+*Summary: In the worst case, we seem to be looking at 340μs vs 530μs for purely evaluation. This is a 56% increase, but negligible in an end-to-end system.*
+
+For the worst-case effect on policy evaluation, we assume a 10k character limit and do addition / multiplication of the literal `i64::MAX` repeated up to this limit +/* `principal.a` where `principal.a == i64::MAX`. All benchmarks are run for 100 iterations:
+
+```
+IBig_addition
+mean: 368.41585148514855μs
+std_dev: 37.14383484933371μs
+IBig_multiplication
+mean: 528.5201386138613μs
+std_dev: 27.376510998230426μs
+i128_addition
+mean: 341.99384158415853μs
+std_dev: 28.056390901018368μs
+i64_mult_by_1
+mean: 327.4614752475248μs
+std_dev: 28.18572847410058μs
+```
+
+Again, this is just evaluation. These differences are in the noise for end-to-end tests:
+
+```
+i128_10kchar_addition: 2.164s for 100 runs
+IBig_10kchar_addition: 2.165s for 100 runs
+```
+
+So an avarage run takes `~21ms` and the runtime increases by `~200μs`.
+
+### Applications Using Cedar Must Limit Input:
+In this RFC, we propose lifting Cedar's limits on integers. This means applications using Cedar would have to enforce their own limits. We expect applications to sanitize any user inputs and this probably includes limiting the lengths of string inputs. We think reasonable length limits (e.g., AVP limits policies to 10k chars) will give acceptable performance, but we should test this more (see unresolved question 2).
+
+### Taking new dependency:
+If this RFC is accepted, Cedar would take a new dependency on `IBig`, increasing the attack surface.
+
+## Alternatives
+
+### Unbounded internal integers, bounded inputs
+Applications will need to limit the integers input as attributes. Cedar could choose some default limit (e.g., `i64`). The current RFC doesn't impose a specific limit and relies on applications to sanitize any user inputs as appropriate.
+
+The downside of choosing a default limit is that analysis could return counter examples that don't respect this limit. It is not clear how to fix this while keeping analysis complete and decidable.
+
+It is probably sufficient to limit the length of the string passed as input, but this should be tested (see unresolved question 2).
+
+### Wrapping integers
+Arithmetic on integers that wrap on overflow is total, but it has caused bugs in the past and so is probably not ideal for an authorization language.
+
+## Unresolved questions
+
+- How efficient is (de)serializing `IBig`s?
+- Benchmarking was done assuming inputs bounded by `[i64::MIN, i64::MAX]` and 10k character policies. How does relaxing these assumptions affect performance?


text/0036-unbounded-ints.md
```diff
@@ -0,0 +1,95 @@
...
+- Reference Issues: (fill in existing related issues, if any)
+- Implementation PR(s): Not implemented, but an implementation was started for benchmarking: https://github.com/cedar-policy/cedar/pull/376
+
+## Timeline
+
+- Started: 2023-11-01
+
+## Summary
+
+Currently, Cedar represents integers using Rust's `i64`. This RFC proposes changing the internal representation to unbounded integers using the `IBig` crate.
```
shaobo-he-aws, line 14: I think more precisely the proposal is to change the Cedar integer type from `Long` to something like `Int`, where the former is a signed 64 bit integer and the latter is a signed integer.

text/0036-unbounded-ints.md
```diff
@@ -0,0 +1,95 @@
...
+
+```
+i128_10kchar_addition: 2.164s for 100 runs
+IBig_10kchar_addition: 2.165s for 100 runs
+```
+
+So an avarage run takes `~21ms` and the runtime increases by `~200μs`.
+
+### Applications Using Cedar Must Limit Input:
+In this RFC, we propose lifting Cedar's limits on integers. This means applications using Cedar would have to enforce their own limits. We expect applications to sanitize any user inputs and this probably includes limiting the lengths of string inputs. We think reasonable length limits (e.g., AVP limits policies to 10k chars) will give acceptable performance, but we should test this more (see unresolved question 2).
```
shaobo-he-aws, line 75: "enforce their own limits" to "enforce their own limits on input character/byte size"? AVP to Amazon Verified Permissions.

text/0036-unbounded-ints.md
```diff
@@ -0,0 +1,95 @@
...
+
+It is probably sufficient to limit the length of the string passed as input, but this should be tested (see unresolved question 2).
+
+### Wrapping integers
+Arithmetic on integers that wrap on overflow is total, but it has caused bugs in the past and so is probably not ideal for an authorization language.
+
+## Unresolved questions
+
+- How efficient is (de)serializing `IBig`s?
+- Benchmarking was done assuming inputs bounded by `[i64::MIN, i64::MAX]` and 10k character policies. How does relaxing these assumptions affect performance?
```
shaobo-he-aws, line 95: Does it make sense to limit the size of input integer values? Like this answer on stackoverflow: https://stackoverflow.com/a/39681707?

text/0036-unbounded-ints.md
emina: A few comments:

(1) Regarding the performance impacts of this proposal, it’s worth emphasizing that this would change the space complexity of Cedar evaluation for the worse.  Given a policy and an input, the Cedar evaluator currently takes constant (extra) space to evaluate the policy with respect to the input.  This is because there is no operator in the language that causes the size of the intermediate computations to grow with policy or input size—all intermediate values can be represented in constant space. 

If we switch to BigInt semantics, we lose this property because addition and multiplication will cause the size of the intermediate values to grow linearly with policy and input size. 

I think that this loss of space efficiency is one of the major downsides of this proposal.  For users who aren’t careful about limiting policy or input size, this could become a non-trivial performance bottleneck.

(2) It is possible but complex and inefficient to express a BigInt semantics while limiting input size to 64-bits.  We can reduce this bounded-inputs-but-unbounded-operators semantics to the theory of Bitvectors, increasing the size of the intermediate Bitvector representations to account for overflow.  So, effectively, the analysis would mimic the space inefficiency of concrete evaluation.  It’s unlikely that this would scale in practice (even just one multiplication puts us at 128 bits), but it would remain sound, complete, and decidable.

(3) I think it’s worth giving serious consideration to the “wrapping integers” (modular arithmetic) alternative.  Yes, this semantics can lead to bugs, just like the error-on-overflow semantics.  However, it is the most efficient option of the three in terms of runtime performance; it is space-efficient; it is efficiently encodable for the purpose of analysis; and it is standard in many mainstream programming languages.   From what we have seen so far, there are very few use cases for arithmetic, and for those few use cases, policy writers can avoid bugs by guarding arithmetic operations with range constraints, e.g.: 
```
0 <= context.usedKB &&
context.usedKB <= 10000 &&
context.usedKB * 1000 < context.maxBytes
```
We can also consider providing linters to detect when policies are using arithmetic without suitable range guards.

shaobo-he-aws (approved):

text/0036-unbounded-ints.md
emina: Thinking more about the [above comment](https://github.com/cedar-policy/rfcs/pull/36#discussion_r1379480592), we can use capabilities to track _policy_ range constraints as part of validation.  Range tracking would re-use the same machinery we’re using today to track `has` constraints.  

This approach has two advantages: 

1. It lets us retain analysis completeness by avoiding putting range constraints in the schema (see [here](https://github.com/cedar-policy/rfcs/pull/13#issuecomment-1786170514) for why arbitrary subtypes can’t be supported).

2. It lets us safely keep the error-on-overflow semantics or switch to the wrapping semantics.   The wrapping semantics would still be worth considering because it has a more efficient SMT encoding than the error-on-overflow semantics. 

The main disadvantage of this approach is that errors are still possible for policies that aren’t validated.  Though that’s the case for all kinds of runtime errors, so it seems like a comparatively small disadvantage.

mwhicks1: > Range tracking would re-use the same machinery we’re using today to track `has` constraints.

Tracking `has` constraints puts tested attributes into a set. Tracking range constraints seems different: It would, to be sound, need to understand arithmetic to ensure there is no overflow. Doing so seems tantamount to the solving the problem of tracking ranges specified in schemas, but in that case you don't have to put the constraints in the policy (you could do that also, but it's not always required). If so, this proposal seems strictly worse than that one. What am I missing?

mwhicks1: I think we should dig into this concern more:

> If we switch to BigInt semantics, we lose this property [of constant (extra) space required for evaluation] because addition and multiplication will cause the size of the intermediate values to grow linearly with policy and input size. I think that this loss of space efficiency is one of the major downsides of this proposal. 

As a practical matter, many Cedar-using apps know their policies in advance, either as static policies or as templates, so they know how big can be. Thus they can bound the space they will need for evaluation. 

For services like Amazon Verified Permissions, which allow users to basically author their own policies, they can set limits on the policy size and/or the input integer size. It's probably wise for them to do so for other reasons.

Also, I note that considering/bounding policy size is already required in order to avoid evaluation failures due to stack overflows (an overflow will cause evaluation of the policy to be skipped, as with other evaluation-time errors).

mwhicks1: > (2) It is possible but complex and inefficient to express a BigInt semantics while limiting input size to 64-bits. 

See also my comment on Alternatives below -- could we do something less than perfect, but likely good enough in practice?

emina: > > Range tracking would re-use the same machinery we’re using today to track `has` constraints.
> 
> Tracking `has` constraints puts tested attributes into a set. Tracking range constraints seems different: It would, to be sound, need to understand arithmetic to ensure there is no overflow. Doing so seems tantamount to the solving the problem of tracking ranges specified in schemas, but in that case you don't have to put the constraints in the policy (you could do that also, but it's not always required). If so, this proposal seems strictly worse than that one. What am I missing?



> > Range tracking would re-use the same machinery we’re using today to track `has` constraints.
> 
> Tracking `has` constraints puts tested attributes into a set. Tracking range constraints seems different: It would, to be sound, need to understand arithmetic to ensure there is no overflow. Doing so seems tantamount to the solving the problem of tracking ranges specified in schemas, but in that case you don't have to put the constraints in the policy (you could do that also, but it's not always required). If so, this proposal seems strictly worse than that one. What am I missing?

Yes, it does need to understand arithmetic, and it is similar to the problem of tracking ranges from schemas. 

The advantage of tracking policy range constraints is that we get safety _without_ needing to put arbitrary subtypes into the schema, which would make analysis incomplete.  That is, ranges are specified directly in policies and directly on (sub)expressions that need them.  

Specifying range constraints in the schema buys some convenience but the downside is the loss of analysis completeness.  We ask users to write conditionals to guard against other errors too, such as missing attribute.  This is similar.

To me, this is the cleanest solution that balances all the design constraints we have—performance, safety, and analyzability—at the cost of losing some convenience compared to adding ranges to the schema.  I think the cost is small, especially since arithmetic is used rarely.

emina: > Also, I note that considering/bounding policy size is already required in order to avoid evaluation failures due to stack overflows (an overflow will cause evaluation of the policy to be skipped, as with other evaluation-time errors).

Stack overflow is a limitation of our current implementation that’s not fundamental to the design.

Using BitInts fundamentally changes the space complexity of Cedar evaluation.  We intentionally omitted string append from the language for the language to avoid this space complexity problem.

text/0036-unbounded-ints.md
emina: > This proposal is mentioned above but has downsides. I responded in a comment; do you agree with it?

This would, again, complicate the analysis because we have to emit extra assertions for all integer fields that may be reachable in policy.  And the design then becomes inconsistent because ranges are allowed on fields but not in sets.  These are two disadvantages are addressed by the solution above, which is to express range constraints as-needed in policies instead.

text/0036-unbounded-ints.md
mwhicks1: > Specifying range constraints in the schema buys some convenience but the downside is the loss of analysis completeness. We ask users to write conditionals to guard against other errors too, such as missing attribute. This is similar.

To make sure I understand correctly:

This proposal does not make range analysis in the validator any simpler than what's being developed in [this PR](https://github.com/cedar-policy/cedar/pull/56). If you do want to do arithmetic (e.g., directly on a number, or even on numbers in a set using `any` / `all` when we implement [RFC 21](https://github.com/cedar-policy/rfcs/blob/any-and-all-operators/text/0021-any-and-all-operators.md)), then putting the constraints in the policy requires solving the same problem you'd have to solve if they were in the schema. The difference is those constraints _only_ arise in-line, not _a priori_.

Leaving constraints out of schemas does simplify analysis in the sense that there is no need to prevent generating an out-of-range counterexample, since all numbers can always be the full range.

The latter advantage is not a user benefit. Yes, analysis would be generating "legal" counterexamples, but only because users cannot express the ranges they would like to limit them to.

text/0036-unbounded-ints.md
mwhicks1: > These are two disadvantages are addressed by the solution above, which is to express range constraints as-needed in policies instead.

I don't believe they are fully addressed. If users really do intend that inputs to fall within a range, then they cannot express this fact, and counterexamples may still be outside the actual, but impossible-to-state, range. The solution above solves the problem of preventing overflow by pushing that responsibility on to the user to always check at run-time.

emina: > I don't believe they are fully addressed. If users really do intend that inputs to fall within a range, then they cannot express this fact, and counterexamples may still be outside the actual, but impossible-to-state, range. The solution above solves the problem of preventing overflow by pushing that responsibility on to the user to always check at run-time.

Checking range constraints in a program (or,  in our case, policy) is standard.  And there will always be constraints that aren't expressible in our type system (see [above](https://github.com/cedar-policy/rfcs/pull/36#discussion_r1380343683)).  So, I don't see this as a major problem.

Another advantage of policy-range constraints is that it's a two-way door.  If we choose to adopt BigInts at some point, all policies that were valid (i.e., not overflowing) under the existing semantics remain valid  in the BigInt semantics. 

But if we adopt BigInts now, there is no going back without a serious break in compatibility.  And, we will obligate all future implementations of Cedar to implement BigInt semantics.  This might carry a larger penalty in other host languages than it does in Rust.

To be clear, from the perspective of formalization and analysis, adopting unrestricted BigInts (so, BigInt literals and inputs are allows) is the easiest solution.   

But I'm not convinced that BigInts are the right long-term solution.  I hesitate to give up on a fundamental efficiency property of the language to more easily support what seems to be a rare use case (arithmetic in policies).

text/0036-unbounded-ints.md
andrewmwells-amazon: I don't really agree with (1).
First, integers are much more compact than strings (i.e., a 10k character integer can be represented more compactly than a 10k char string). Concretely, a 10k character integer takes ~3k bits to represent vs 80k bits for the string. That's a fairly small memory increase. Considering only extra space seems unfair.

Second, evaluation doesn't currently take constant extra space, because we recurse down the expression tree. This could be fixed, but hasn't really been a problem yet. For a 10k character policy: `principal.aaaa || principal.aaab || ...`, evaluating recursively will take far more than 3k extra bits.

If using constant extra space is really a need, then I think bounding attribute values will give us the same big-O complexity we currently have.

andrewmwells-amazon: For (3), I also think wrapping semantics would be better than the current semantics. Because we skip on error, any almost arithmetic can give a surprising result, but wrapping requires arithmetic + comparison.

A problematic expression for wrapping could be:
```
curr_time_ms * 1000 > end_time_window_s
```
 So I still think unbounded integers are the best option.

text/0036-unbounded-ints.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Taking new dependency:
+If this RFC is accepted, Cedar would take a new dependency on `IBig`, increasing the attack surface.
+
+## Alternatives
+
+### Unbounded internal integers, bounded inputs
+Applications will need to limit the integers input as attributes. Cedar could choose some default limit (e.g., `i64`). The current RFC doesn't impose a specific limit and relies on applications to sanitize any user inputs as appropriate.
+
+The downside of choosing a default limit is that analysis could return counter examples that don't respect this limit. It is not clear how to fix this while keeping analysis complete and decidable.
```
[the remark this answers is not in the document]
andrewmwells-amazon: This is actually my preference but from the group discussion I thought I was in the majority so I tried to represent that. A few points:
- I think it's ok for analysis to not return counter examples matching whatever bounds are chosen by a specific application. As you point out, this is currently the case for string lengths and it is also the case for numbers of entities, numbers of ancestors and probably some other bounds.
- You can indeed add assertions for specific integers and could even iteratively add constraints for members of sets once  a counter example is returned 
- The solver shouldn't return numbers bigger than needed to form a counter example. Unfortunately, the solver could return i64::MAX+1 when -1 was also a counter example.

text/0036-unbounded-ints.md
emina: > I don't really agree with (1). First, integers are much more compact than strings (i.e., a 10k character integer can be represented more compactly than a 10k char string). Concretely, a 10k character integer takes ~3k bits to represent vs 80k bits for the string. That's a fairly small memory increase. Considering only extra space seems unfair.
> 
> Second, evaluation doesn't currently take constant extra space, because we recurse down the expression tree. This could be fixed, but hasn't really been a problem yet. For a 10k character policy: `principal.aaaa || principal.aaab || ...`, evaluating recursively will take far more than 3k extra bits.
> 
> If using constant extra space is really a need, then I think bounding attribute values will give us the same big-O complexity we currently have.

To be clear, the concern about the space is a general concern about losing a nice property of the language design, not necessarily what it means today in terms of practical memory usage.  We’re moving from a design in which it is possible, at least in principle, for the space used by the evaluator to be independent of input size, to one in which this is not possible.  In principle, we can take a Cedar expression and statically calculate the maximum amount of memory needed to evaluate that expression.  This will no longer be possible with full-on BigInts (inputs + literals) and arithmetic.  That’s a big conceptual change and needs really strong justification in my eyes. 

If we go with half-way BigInts, where we force inputs to be bounded, then we either lose analysis completeness with respect to those bounds, or the analysis becomes more complicated and expensive.  Again, I don’t think the benefits of big ints justify complicating and slowing down the analysis to enforce these bounds.  This is no different than the argument against adding enums to the schema.  It is different from strings because the Cedar language places no a priori bounds on string length.  Given sufficient memory, we can construct a valid Entity object with a string field of any length.  That is, the analysis is complete with respect to the language specification.

There is only one benefit that BigInts bring: arithmetic becomes safer.  As far as we can tell, there are very few use cases for arithmetic in Cedar.  We’d be walking through a one-way door, with non-trivial downsides, to increase the safety of a feature that’s rarely used.

Adopting BigInts also means that we’ll force all other implementations of Cedar (and there are some already) to adopt this semantics.  This semantics might be much more costly in other host languages than it is in Rust.

I acknowledge that full-on BigInts is the most mathematically simple and elegant option on the table.  Full-on BigInts is the only semantics that is both safe and adds _no_ complication to the formalization, the validator, or the analysis.  Every other option—half-way Big ints, error-on-overflow, wrap-on-overflow—complicates at least one of these components.  Of these, half-way Big ints is the least appealing in my eyes:  it carries a performance penalty, forces all future implementations to use BigInts, and makes the analysis expensive and/or incomplete.

My favorite option would be to drop arithmetic from the language, but that option is not on the table :-).  Of the remaining options, my weighing of the trade offs gives the following stack ranking:  (1) wrapping semantics, (2) error-on-overflow, (3) half-way Big Ints, and (4) full-on BigInts, with a big gap between (2) and (3).

emina: > For (3), I also think wrapping semantics would be better than the current semantics. Because we skip on error, any almost arithmetic can give a surprising result, but wrapping requires arithmetic + comparison.
> 
> A problematic expression for wrapping could be:
> 
> ```
> curr_time_ms * 1000 > end_time_window_s
> ```
> 
> So I still think unbounded integers are the best option.

Yes, wrapping semantics can definitely lead to unexpected behaviors where BigInts wouldn’t.  These behaviors can be ruled out by properly guarding arithmetic with inline range constraints, which does place extra burden on the user.  Validation of policy range constraints can help with that (though not fully remove the sharp edge, like BigInts would).

andrewmwells-amazon: > First, integers are much more compact than strings (i.e., a 10k character integer can be represented more compactly than a 10k char string). Concretely, a 10k character integer takes ~3k bits to represent vs 80k bits for the string. That's a fairly small memory increase. Considering only extra space seems unfair.

`3k` should be `30k`

andrewmwells-amazon: > To be clear, the concern about the space is a general concern about losing a nice property of the language design, not necessarily what it means today in terms of practical memory usage. We’re moving from a design in which it is possible, at least in principle, for the space used by the evaluator to be independent of input size, to one in which this is not possible. 

I still don't really agree with this characterization. In practice, responsible users will have bounds on their entity data, so there will still be a limit computable by knowing the policy and the bounds placed on entities. Should we bound the maximum number of ancestors passed in to the evaluator?

emina: Reading through the proposal and this thread again, I think that, for now, the least invasive option is to (1) retain error-on-overflow semantics, and (2) extend the validator to enforce in-policy range checks, just like it enforces in-policy attribute presence checks.

This option, while not my first choice, lets us (1) not break any existing code, (2) retain analysis completeness (at a reasonable cost), and (3) easily switch to BigInts semantics in the future, if we find that they bring additional benefits beyond simplifying the use of arithmetic. In terms of keeping the door open for a switch to BigInts, this option means that any validated policy will continue to behave correctly under BigInts.  Any policy that's not validated retains its current semantics, and it errors less under BigInts.

In terms of implementation cost, this option means that we have to change (1) the formalization, (2) the validator, and (3) the encoding to SMT.  Here, we can use Lean's dependent types for (1); it will be more complex than using plain Ints, but it's doable.  For (2), we can amortize the cost of implementing in-policy range analysis by reusing code from .  For (3), it will be a matter of adding overflow guards to the encoding.  

For users, the only consequence of this solution will be that the validator will reject some policies that were accepted before---those that used arithmetic without guards.  This can be easily fixed by adding the guards. It's not ideal that we'll introduce this break, but from what we can tell, the impact will be very limited due to the rare use of arithmetic.

As a point of comparison, to switch to full-on BigInts, which is admittedly the cleanest semantics, we would have to change (1) entity and policy parsers, (2) internal and potentially external APIs, and (3) the production evaluator.  These changes are invasive and will impact the users immediately if we have to change external APIs.  And again, adopting the full-on BigInts semantics does negatively impact the efficiency of the Cedar language design, and it places a potentially expensive requirement on future implementations.  

It's possible that the conceptual concerns about full-on BigInts don't translate to real downsides in practice.  But until we have strong evidence that this is true, I vote for adopting the conservative course of action proposed above:  error-on-overflow + in-policy range constraints.

mwhicks1: To be clear: in-range policy constraints means basically using the machinery in [this PR](https://github.com/cedar-policy/cedar/pull/56), but not requiring an extension to schemas to provide bounds. Instead, all integers are assumed to be the full 64 bits at the start, but local constraints will constrain them, e.g., when validating `context.size < 100 && context.size + 100 ...` the validator would examine expression `context.size + 100` under the assumption, due to the preceding conjunct, that `context.size` was less than 100.

I'm fine with this proposal.

emina: > To be clear: in-range policy constraints means basically using the machinery in [this PR](https://github.com/cedar-policy/cedar/pull/56), but not requiring an extension to schemas to provide bounds. Instead, all integers are assumed to be the full 64 bits at the start, but local constraints will constrain them, e.g., when validating `context.size < 100 && context.size + 100 ...` the validator would examine expression `context.size + 100` under the assumption, due to the preceding conjunct, that `context.size` was less than 100.




> To be clear: in-range policy constraints means basically using the machinery in [this PR](https://github.com/cedar-policy/cedar/pull/56), but not requiring an extension to schemas to provide bounds.

Yes, that's right.

andrewmwells-amazon: I'm also fine with this proposal.

cdisselkoen: This sounds fine to me (retain error-on-overflow, add in-policy range analysis that forces policy authors to add guards, so that validated policies can never actually overflow)

andrewmwells-amazon:
Closing in favor of https://github.com/cedar-policy/rfcs/pull/39

PR #37, opened 2023-11-02
Null JSON values in context and entity attribute JSON data

john-h-kastner-aws:
[Rendered](https://github.com/cedar-policy/rfcs/blob/jkasnter/rfc/0037/text/0037-null-context-entity-data.md)

### DIFF
--- a/text/0037-null-context-entity-data.md
+++ b/text/0037-null-context-entity-data.md
@@ -28,8 +28,8 @@ This becomes a problem when a users includes `null` in their data, expecting tha
 
 ## Detailed design
 
-This RFC proposes ignoring `null` values _only_ when they occur as thj value of a record attribute while continuing to error on `null` elsewhere.
-This is based on the observation that, while we can't handle `null` in all situations, a `null` value appearing record attribute can reasonably be treated as explicitly marking the attribute _not_ present in the record.
+This RFC proposes ignoring `null` values _only_ when they occur as the value of a record attribute while continuing to error on `null` elsewhere.
+This is based on the observation that, while we can't handle `null` in all situations, a `null` value appearing as a record attribute can reasonably be treated as explicitly marking the attribute _not_ present in the record.
 
 ### JSON Parsing Background
 
@@ -48,13 +48,13 @@ The change proposed in this RFC supports `null` where there is an obvious interp
 
 If an object contains an attribute more than once it will still be an error even if the value of some of the attributes are `null`.
 All JSON formats used by Cedar should error when encountering an attribute more than once in any JSON object.
-Even though the `null` value is marking the attribute as as absent, from the Cedar record, the entry is present in the JSON object, so it is still an error.
+Even though the `null` value is marking the attribute as absent from the Cedar record, the entry is present in the JSON object, so it is still an error.
 Moreover, explicitly marking an attribute as absent and later providing a value for that attribute is very likely to be a mistake.
 
 We use JSON objects with the reserved attributes `__expr`, `__entity`, or `__extn` as escapes for writing Cedar values that don't have JSON equivalents.
 These are JSON objects, so this RFC as written up to this point implies ignoring `null` on attributes appearing here.
 We will instead continue to treat `null` as an error in this context.
-If this were not an error and we instead ignored `null` value, `{"__entity": null}` would be treated the same as `{}`, but the `__entity` escape should parse to an entity while `{}` parses to a record.
+If this were not an error and we instead ignored `null` values, `{"__entity": null}` would be treated the same as `{}`, but the `__entity` escape should parse to an entity while `{}` parses to a record.
 We could alternatively find a way to parse `null` to an entity, perhaps the unspecified entity.
 There is currently no way to specify the unspecified entity in out JSON data format.
 If we want to extend the formats to support it, this may be a reasonable approach, but that extension is beyond the scope of this RFC, and can be added later in a backwards compatible manner if `null` remains an error here.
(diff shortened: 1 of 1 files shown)


text/0037-null-context-entity-data.md
```diff
@@ -0,0 +1,124 @@
...
+We don't want to add `null` as in Java where any reference may be `null`, but it could be added as the element of a unit type without causing too much trouble.
+This could work well for when using Cedar as a dynamic language, but its use would be limited in a statically typed context.
+
+### Error on `null`
+
+This is our current behavior.
+As stated above, it limits what data can be used to construct a context, forcing users to remove it in their own preprocessing passes. 
+Cedar should be easy and ergonomic to use, so we should hesitate to offload work users without good reason, but, if the correct interpretation of `null` is not always the same, it may be the correct choice here.
+
+If we take this alternative, we should keep in mind that many users are likely to find themselves writing the same filter to remove `null` from their data, so we should consider providing this filter ourselves, even if it is not an integral part of the Cedar data format.
```
mwhicks1, line 120: This doesn't sound that hard, actually. There are lots of libraries for easy JSON parsing/transformation, no? I think that it might be confusing that Cedar drops `null`s for attributes, but errors when they are provided in sets. Better to let users drop `null` if it's acceptable to them, wherever `null` might appear?

mwhicks1: And the drawbacks you've listed seem non-trivial ...

john-h-kastner-aws: Yes, based on the drawbacks I'm also no longer fully convinced that dropping `null` attribute only is the right solution. Dropping `null` everywhere actually seems better, but the example with `all` is very misleading.

text/0037-null-context-entity-data.md
```diff
@@ -0,0 +1,124 @@
...
+### Unexpected behavior
+
+Ignoring `null` may not be the desired behavior for all users.
+It seems reasonable that a `null` attribute value can be safely ignored, but a user may expect that `context has foo` evaluates to `true` if `foo` is present in their JSON context data.
+It will instead evaluate to `false` if `foo` is present but has a `null` value.
+
+In support of this point, observe how JavaScript treats objects containing `null`:
+```js
+Object.hasOwn({foo: null}, 'foo') // true
+```
```
khieta, line 82: This seems to be a pretty big drawback -- it's making me lean more towards disallowing `null`.

text/0037-null-context-entity-data.md
khieta: I don't think there's too much value in us providing the filter ourselves. Our filter would presumably just be a wrapper around some existing JSON parsing function. Also, the context/entity data may be coming from other languages (not Rust) and it may make more sense to do the filtering there.

text/0037-null-context-entity-data.md
khieta: What errors does a user see when they pass in a `null` value somewhere? Could we address the issue in this RFC by just making clearer error messages to explain why `null` values aren't allowed?

khieta (commented):
I'd like to see the scope of this RFC expanded to Cedar's JSON representations for schemas and policies. Although those are conceptually a little different (they're "code"/"metadata" rather than "data") I think it's useful to have the discussion all in one place.

text/0037-null-context-entity-data.md
```diff
@@ -0,0 +1,124 @@
...
+## Basic example
+
+Given a JSON object `{"foo": 1, "bar": null}`, we will ignore the `null` attribute and construct the Cedar record `{foo: 1}` instead of returning an error result.
+In the JSON object `{"foo": [null]}`, the `null` is not directly an attribute value, so current behavior will not change (we fail to construct a Cedar record).
+
+## Motivation
+
+We want to support users passing generic JSON data as a request context with minimal preprocessing requirements.
+We do this by constructing, for example, Cedar records from JSON objects and Cedar strings from JSON strings.
+Notably, however, we will not construct any request context if there is a `null` JSON value appearing _anywhere_ in the JSON data, requiring that users who may have `null` in their context preprocess their data to remove it.
```
cdisselkoen, line 26: From reading this, I'm unclear whether this is intended to be a statement of what this RFC is proposing, or what current Cedar does.  With my domain knowledge I know this is a statement of the status quo, but the text should clarify this.

Also, I feel that the Motivation section as a whole is missing maybe one more sentence explaining why the status quo is a problem and what user problems this RFC proposes to solve.

text/0037-null-context-entity-data.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Detailed design
+
+To build the attributes for a context or entity, we accept a generic [JSON value](https://www.json.org/json-en.html) with a few limitation:
+
+1. If the value is an object, it must not contain the same attribute twice. If it contains the reserved attributes `__expr`, `__entity`, or `__extn`, then it may also be rejected based on specific rules for these escapes.
+2. If the value is a number, it must be an integer. A JSON value could be a fractional (e.g, `1.0`) or exponential (e.g., `1e0`) value, but these are not permitted.
+3. The `null` value may not occur anywhere.
```
cdisselkoen, line 34: This is also unclear.  Because this is in the "Detailed design" section, I expected it to describe what the RFC is proposing.  But this RFC isn't actually proposing that "The `null` value may not occur anywhere".  We should more clearly delineate what is the status quo and what is the proposed change.

text/0037-null-context-entity-data.md
```diff
@@ -0,0 +1,124 @@
...
+In the case of `null`, Cedar does not have any `null` value, so there is no single obvious Cedar value that it should map to in all situations.
+
+In this RFC we observe that when `null` occurs as the value of a record attribute there is a natural interpretation.
+We can treat the `null` value as explicitly marking the attribute as _not_ present in the record, so we can ignore the attribute while continuing to error on `null` anywhere else.
+
+If an object contains an attribute more than once it will still be an error even if the value of some of the attributes are `null`.
+
+We use JSON objects with the reserved attributes `__expr`, `__entity`, or `__extn` as escapes for writing Cedar values that don't have JSON equivalents.
+These are JSON objects, so this RFC as written up to this point implies ignoring `null` on attributes appearing here.
+We will instead continue to treat `null` as an error in this context.
```
cdisselkoen, line 47: Curious what the reasoning is either way -- why should we accept or not accept `"__entity": null` and similar.

john-h-kastner-aws: Expanded on this bit in the text. Treating it as `{}` would implicitly make it a record instead of an entity. We could parse `null` as an entity (unspecified?), but I think that's beyond the scope of the problem this RFC wants to solve.

john-h-kastner-aws:
After discussion on this PR, I am convinced that the drawbacks as listed in the RFC outweigh the benefit of accepting more JSON values, so I am moving this RFC into the final comment period with intent to reject.

cdisselkoen:
If we do reject this RFC as proposed, do we need any documentation updates to make Cedar's behavior on `null`s clearer and suggest callers filter `null`s as appropriate for their application?

john-h-kastner-aws:
> If we do reject this RFC as proposed, do we need any documentation updates to make Cedar's behavior on `null`s clearer and suggest callers filter `null`s as appropriate for their application?

Yes. Some concrete action items we'll open as issues on relevant repositories if this is rejected: 

* Update documentation for entity and context JSON formats to explicitly call out `null` as forbidden.
* Update documentation to suggest that users with `null` in their data may want to preprocess to remove it, but also mention the issues this has around `has` and `contains`.
* Update error message on `null` values to clearly state that `null` is the problem

john-h-kastner-aws:
Closing this RFC as rejected at the end of the final comment period.

PR #38, opened 2023-11-07
Update RFC #19

khieta:
Updates to RFC #19
* Changed dates to YYYY-MM-DD format for consistency with other RFCs
* Added a "landed" date
* Removed the text about validation of unspecified entities, per discussion on #35

### DIFF
--- a/text/0019-stricter-validation.md
+++ b/text/0019-stricter-validation.md
@@ -139,4 +139,4 @@ The proposal is, we believe, the minimal change to make strict mode self-contain
 
 ## Updates
 
-* 2023-11-07: The original text said that `AnyEntity` was used to type unspecified entities. This is not the case in any released version of Cedar -- instead unspecified entities are given a special `Unspecified` type. Updated text to reflect this.
+* 2023-11-07: The original text said that the handling of unspecified entities needed to be adjusted. However, although released implementations of Cedar type unspecified entities as `AnyEntity`, `==` and `in` expressions involving unspecified entities are typed as `False` (rather than `Bool`). This provides the same behavior as using a special `Unspecified` type, as originally proposed in this RFC.
--- a/text/0019-stricter-validation.md
+++ b/text/0019-stricter-validation.md
@@ -9,7 +9,7 @@
 
 - Started: 2023-07-11
 - Entered FCP: 2023-08-18
-- Accepted: 2023-08-18
+- Accepted: 2023-09-05
 - Landed: 2023-09-08 on `main`
 
 ## Summary
--- a/text/0019-stricter-validation.md
+++ b/text/0019-stricter-validation.md
@@ -7,10 +7,10 @@
 
 ## Timeline
 
-- Start Date: 11-07-2023
-- Date Entered FCP: 18-08-2023
-- Date Accepted: 18-08-2023
-- Date Landed:
+- Start Date: 2023-07-11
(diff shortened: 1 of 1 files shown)


text/0019-stricter-validation.md
```diff
@@ -7,10 +7,10 @@
...
 ## Timeline
 
-- Start Date: 11-07-2023
-- Date Entered FCP: 18-08-2023
-- Date Accepted: 18-08-2023
-- Date Landed:
+- Start Date: 2023-07-11
+- Date Entered FCP: 2023-08-18
+- Date Accepted: 2023-08-18
+- Date Landed: 2023-09-08 on `main`
```
cdisselkoen, line 13: Suggestion to make this align better with our current template

```suggestion
- Started: 2023-07-11
- Entered FCP: 2023-08-18
- Accepted: 2023-08-18
- Landed: 2023-09-08 on `main`
```

cdisselkoen (approved):
In this PR, the RFC gets the same date for FCP and for Accepted.  Is that correct?

khieta:
> In this PR, the RFC gets the same date for FCP and for Accepted. Is that correct?

Good point -- accepted date was wrong. Will fix in next commit.

text/0019-stricter-validation.md
```diff
@@ -150,3 +136,7 @@ Mitigating this issue is that the backward incompatibility will be minimal in pr
 ## Alternatives
 
 The proposal is, we believe, the minimal change to make strict mode self-contained, but not too different from what was there before. It also should not result in pervasive changes to the existing code. Alternative designs we considered (e.g., allowing more permissive mode features in strict mode) would be more invasive.
+
+## Updates
+
+* 2023-11-07: The original text said that `AnyEntity` was used to type unspecified entities. This is not the case in any released version of Cedar -- instead unspecified entities are given a special `Unspecified` type. Updated text to reflect this.
```
mwhicks1, line 142: Even for validation in permissive mode ?

khieta: Yep, both modes. The type of an unspecified entity (i.e., a principal/resource in an environment where the action was missing `appliesTo` field(s) in the schema) is `Entity({Unspecified})` rather than `AnyEntity`. `AnyEntity` is only used for template slots and the UB of actions.

In typechecking, `<Unspecified> == <Specified Entity>` is typed as `False`.

khieta: Just double checked the code, and I was too hasty 😬 Here is the revised answer:

"Unspecified" entities are given type `AnyEntity` by the type checker, but in the rules for typing `==` and `in` are special-cased to return `False` when one element is unspecified.

khieta: This has been true since the initial release, see [`type_of_equality`](https://github.com/cedar-policy/cedar/blob/release/2.0.x/cedar-policy-validator/src/typecheck.rs#L1916) and [`typecheck_in`](https://github.com/cedar-policy/cedar/blob/release/2.0.x/cedar-policy-validator/src/typecheck.rs#L1975).

Will update the note to accurately reflect this. Sorry for the bugs... I have a lot of balls in the air this week.

khieta: Updated

mwhicks1 (approved):

andrewmwells-amazon (approved):

PR #39, opened 2023-11-08
RFC 39 for guarded arithmetic

andrewmwells-amazon:
[Rendered](https://github.com/cedar-policy/rfcs/blob/andrewmwells/guarded-arith/text/0039-guarded-arith.md)

### DIFF
--- a/text/0037-guarded-arith.md
+++ b/text/0037-guarded-arith.md
@@ -1,0 +1,82 @@
+# Guard Arithmetic Expressions within Policies
+
+## Related issues and PRs
+
+- Reference Issues: No existing issue, but supersedes https://github.com/cedar-policy/rfcs/pull/36
+- Implementation PR(s): (Existing PR for range analysis based on schema: https://github.com/cedar-policy/cedar/pull/56)
+
+## Timeline
+
+- Started: 2023-11-07
+
+## Summary
+
+We will use a range analysis (as in https://github.com/cedar-policy/cedar/pull/56) but with the bounds given in the policy rather than in the schema.
+
+## Motivation
+
+Arithmetic with bounded integers is a common source of bugs [https://blog.research.google/2006/06/extra-extra-read-all-about-it-nearly.html]. Consider the following policy:
+
+```
+forbid(principal, action, resource) when {
+   principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms ||
+   !principal.is_mfa
+};
+```
+
+A casual reader of this policy may conclude that all actions are forbiden when `principal.mfa` is `false`. This is wrong. The lhs (`principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms`) is evaluated first and if it errors the entire policy is skipped.
+
+To avoid this class of mistakes, we want to change Cedar's validator to error on arithmetic expressions that could overflow. To pass validation, arithmetic should be guarded. For example, the above policy should be written:
+
+```
+forbid(principal, action, resource) when {
+   (principal.elevated_privileges_end_time_s < i64::MAX / 1000 &&
+    principal.elevated_privileges_end_time_s > i64::MIN / 1000 &&
+    principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms) ||
+   !principal.is_mfa
+};
+```
+
+## Detailed design
+
+Every non-identity preserving arithmetic operation should be guarded by comparisons that prevent over/underflow.
+
+- We add intervals to represent the ranges of each integer.
+- Integers without explicit intervals are assumed to lie in `[i64::MIN, i64::MAX]` so almost any arithmetic operation on them will be a validation error.
+- We use interval arithmetic, which is imprecise, but because the bounds are written in the policy, the intervals can be precisely given at the point of each operation.
+
+
+## Drawbacks
+
+The policies that use arithmetic and will pass validation are ugly.
+
+We assume almost no one uses arithmetic, so we think this is an acceptable tradeoff. If there are significant uses of arithmetic, a better approach is likely desirable.
+
+## Alternatives
+
+### Unbounded Integers
+Unclear how these will affect our FFI. We can take this proposal now and switch to unbounded ints later more easily than we can switch now and reverse that decision later.
+
+### Bounds in the schema.
+It is not clear how to analyze a set with members of an EntityType that has an integer attribute with bounds while keeping analysis sound, complete and decidable.
+
+### Status quo
+See the downside in the motivation.
+
+### Less general range analysis
+Unlike https://github.com/cedar-policy/cedar/pull/56, because we're putting bounds in the policies, we could require explicit bounds for each arithmetic operation.
+
+So
+```
+(if principal.a < 9,223,372,036,854,774 && principal.a > -9,223,372,036,854,774 then principal.a * 1000 else 0) * 500 > foo
+```
+would require new bounds:
+
+```
+principal.a * 1000 + 500 < i64::MAX && (if principal.a < 9,223,372,036,854,774 && principal.a > -9,223,372,036,854,774 then principal.a * 1000 else 0) + 500 > foo
+```
+This would be easier to implement, but much more annoying to use if anyone uses chained arithmetic operations.
+
+## Unresolved questions
+
+What should the validation error messages look like? Can we suggest bounds that will look more reasonable in policies than `principal.a < 9,223,372,036,854,774 && principal.a > -9,223,372,036,854,774 && ...`?
--- a/text/0039-guarded-arith.md
+++ b/text/0039-guarded-arith.md
@@ -58,7 +58,20 @@ We assume almost no one uses arithmetic, so we think this is an acceptable trade
 Unclear how these will affect our FFI. We can take this proposal now and switch to unbounded ints later more easily than we can switch now and reverse that decision later.
 
 ### Bounds in the schema.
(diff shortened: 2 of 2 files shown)


text/0037-guarded-arith.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+A casual reader of this policy may conclude that all actions are forbiden when `principal.mfa` is `false`. This is wrong. The lhs (`principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms`) is evaluated first and if it errors the entire policy is skipped.
+
+To avoid this class of mistakes, we want to change Cedar's validator to error on arithmetic expressions that could overflow. To pass validation, arithmetic should be guarded. For example, the above policy should be written:
+
+```
+forbid(principal, action, resource) when {
+   (principal.elevated_privileges_end_time_s < i64::MAX / 1000 &&
+    principal.elevated_privileges_end_time_s > i64::MIN / 1000 &&
```
mwhicks1, line 34: Are you indirectly proposing to add `i64::MAX` and `i64::MIN` to the language? Without these things, writing inline bounds checks will be a hassle. Assuming, therefore, you _are_ proposing to add them, I suggest we not use namespace syntax, which implies `i64::MAX` is a type, not a value. Want to think of some proposals for alternative syntax?

text/0037-guarded-arith.md
```diff
@@ -0,0 +1,82 @@
...
+
+We assume almost no one uses arithmetic, so we think this is an acceptable tradeoff. If there are significant uses of arithmetic, a better approach is likely desirable.
+
+## Alternatives
+
+### Unbounded Integers
+Unclear how these will affect our FFI. We can take this proposal now and switch to unbounded ints later more easily than we can switch now and reverse that decision later.
+
+### Bounds in the schema.
+It is not clear how to analyze a set with members of an EntityType that has an integer attribute with bounds while keeping analysis sound, complete and decidable.
```
mwhicks1, line 61: It's worth elaborating on this point -- can you show an example? Would the issue only arise once/if `.any` and `.all` are in the language?

text/0037-guarded-arith.md
```diff
@@ -0,0 +1,82 @@
...
+So
+```
+(if principal.a < 9,223,372,036,854,774 && principal.a > -9,223,372,036,854,774 then principal.a * 1000 else 0) * 500 > foo
+```
+would require new bounds:
+
+```
+principal.a * 1000 + 500 < i64::MAX && (if principal.a < 9,223,372,036,854,774 && principal.a > -9,223,372,036,854,774 then principal.a * 1000 else 0) + 500 > foo
+```
+This would be easier to implement, but much more annoying to use if anyone uses chained arithmetic operations.
```
mwhicks1, line 78: Change "This would be easier" to "Requiring bounds for each operation would be easier ..." just so we are clear what your assessment is referring to.

text/0037-guarded-arith.md
emina: I don't think we should add these constants to the language.  They aren't useful for writing range constraints in that they are too lenient to allow any non-trivial arithmetic to remain within bounds.

Plus, we definitely don't want to allow division :)

text/0037-guarded-arith.md
emina: This is an issue no matter what.  For example, without being able to enforce these  bounds, the solver could return false positives for set membership constraints, e.g., `principal.someNumbersLessThanTen.contains(1000)`.

emina (approved):

text/0037-guarded-arith.md
andrewmwells-amazon: Yeah, I just wrote this to make the example easier to read, no intent to add to the language

shaobo-he-aws:
I would like to see a more elaborated example that involves a chain of arithmetic computations. For instance, what the guards should be for expressions like `principal.a + principal.b - principal.c`.

emina:
> I would like to see a more elaborated example that involves a chain of arithmetic computations. For instance, what the guards should be for expressions like `principal.a + principal.b - principal.c`.

You’d provide the same bounds inline as you would in a schema, so:

```
MIN_A <= principal.a && principal.a <= MAX_A &&
MIN_B <= principal.b && principal.b <= MAX_B &&
MIN_C <= principal.c && principal.c <= MAX_C &&
principal.a + principal.b - principal.c == principal.d
```

where `MIN_*` and `MAX_*` are placeholders for integer literals.

text/0039-guarded-arith.md
```diff
@@ -0,0 +1,95 @@
...
+    principal.elevated_privileges_end_time_s * 1000 > context.current_time_ms) ||
+   !principal.is_mfa
+};
+```
+
+## Detailed design
+
+Every non-identity preserving arithmetic operation should be guarded by comparisons that prevent over/underflow.
+
+- We add intervals to represent the ranges of each integer.
```
shaobo-he-aws, line 44: We probably want to refine what "each integer" means here. For instance, we probably don't want annotate literals with ranges.

shaobo-he-aws:
> > I would like to see a more elaborated example that involves a chain of arithmetic computations. For instance, what the guards should be for expressions like `principal.a + principal.b - principal.c`.
> 
> You’d provide the same bounds inline as you would in a schema, so:
> 
> ```
> MIN_A <= principal.a && principal.a <= MAX_A &&
> MIN_B <= principal.b && principal.b <= MAX_B &&
> MIN_C <= principal.c && principal.c <= MAX_C &&
> principal.a + principal.b - principal.c == principal.d
> ```
> 
> where `MIN_*` and `MAX_*` are placeholders for integer literals.

Yes, so we essentially need to add ranges to "atom" operands. I have a couple of further questions.

1. If a subexpression shows up in multiple expressions involving arithmetics, we should only annotate it once?
2. Do we need to distinguish lt/le/gt/ge expressions used for range annotation and those that are part of authorization logic?

text/0039-guarded-arith.md
mwhicks1: The RFC here is talking about the internal implementation. In analyzing the code, it will associate an interval with each integer-typed expression, and then refine that interval according to comparison operations. So consider this expression:
```
MIN_A <= principal.a && // A 
principal.a <= MAX_A && // B
... principal.a + 100 ... 
```
You start with an interval of `principal.a` as `[i64::MIN,i64::MAX]`. Then at position A you will have refined this interval to be `[MIN_A,i64::MAX]`. At position B you will have refined the interval to be `[MIN_A,MAX_A]`. Finally, based on this, the expression `principal.a + 100` will be checked to ensure that `MIN_A + 100` is not out of bounds (it isn't) and likewise that `MAX_A + 100` is not out of bounds.

A literal's interval will just have the min/max be the interval itself, e.g., the literal 5 will have the interval `[5,5]`. There are no annotations.

mwhicks1:
> Yes, so we essentially need to add ranges to "atom" operands. I have a couple of further questions.

Not "atoms" only -- any integer-typed expression.

> If a subexpression shows up in multiple expressions involving arithmetics, we should only annotate it once?

Each subexpression has its own associated interval when analyzing it, just like each subexpression has its own associated type.

> Do we need to distinguish lt/le/gt/ge expressions used for range annotation and those that are part of authorization logic?

As I said, there is no special annotation; we just analyze the use of inequality operators to refine associated intervals.

mwhicks1 (approved):
LGTM.

mwhicks1:
Despite the generally positive comments here, after some in-person discussion @shaobo-he-aws @emina @andrewmwells-amazon and myself feel we should reject this RFC and stick with the status quo. Here is the reasoning:

1. The Cedar semantics holds that a policy that errors is ignored when making the final authorization decision.
2. A Cedar design principle is that evaluation errors should truly be _errors_, not expected outcomes designed to ignore a policy. For example, we say you should write `resource has owner && resource.owner == principal` if `owner` is an optional field. If we embraced evaluation errors as an acceptable way of skipping policies, you could have just written `resource.owner == principal` and if `resource` didn't have `owner` then evaluation would error and the policy would be skipped.
3. A consequence of this principle is that the policy validator aims to prevent evaluation errors. Thus it validates expressions like `resource has owner && resource.owner == principal` but not `resource.owner == principal`.
4. Ideally, the validator could rule out _all_ possible evaluation errors. However, it does not: Valid policies could still error due (1) integer overflow on arithmetic, or (2) accessing the attributes or parents (in an `in` expression) of an entity not present in the entity store. Why is this Ok? Because (a) we believe such errors are rare and there are benefits to usability to ignore them at validation time, and because (b) Cedar offers applications the ability to work around the validator's incompleteness. In particular, an application can observe that an error occurred by looking at the diagnostics of an authorization response, and then take an alternative action if it wishes, such as denying the request (esp. if the erroring policy is a `forbid` policy).

This RFC aims to address (1), so that the validator rules out more errors before policies are evaluated. Why not accept it?

Basically, we feel that if this RFC is adopted that (a) policies that use integer arithmetic will be hard to read, due the required added checks, and (b) the validator code that processes those checks is complicated, adding a burden to maintenance and on proof (for the Cedar Lean specification). Another downside is that this RFC is not backward compatible: Policies that previously validated may not validate with this RFC. These downsides are not worth the upside of ruling out overflow errors, given that integer arithmetic so far is rare, and there is a way for applications to customize handling of dynamic overflow errors, if they wish. To add: the added checks required by this RFC are not needed for policy analysis -- the possibility of integer overflow is easily included in the policy logical encoding.

andrewmwells-amazon:
Moving this RFC to FCP with intent to reject.

andrewmwells-amazon:
Closing this PR as the RFC is now rejected.

PR #40, opened 2023-11-21
RFC 0040 - add trait implementations to export structs to JSON Schema

Swolebrain:
[Rendered](https://github.com/swolebrain/cedar-rfcs/blob/main/text/0040-json-schema-generation.md)


aaronjeline (approved):
Seems reasonable and useful

text/0040-json-schema-generation.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Basic example
+
+An example of what this entails is here https://github.com/josema03/cedar/compare/379a2b1330bc5f82bb2bdd9cb39db1ebd41a053b..d501acba0a878dfe898427cdc0645d8b37779562
+
+Of course, we would do all of these trait implementations (as well as the dependency addition) under feature flags so as to not affect current consumers.
+
+## Motivation
+
+Without this functionality, then structs like (AuthorizationCall)[https://github.com/cedar-policy/cedar/blob/main/cedar-policy/src/frontend/is_authorized.rs#L146] and (RecvdSlice)[https://github.com/cedar-policy/cedar/blob/main/cedar-policy/src/frontend/is_authorized.rs#L286] need to be duplicated in the calling code.
```
khieta, line 24:
```suggestion
Without this functionality, then structs like [AuthorizationCall](https://github.com/cedar-policy/cedar/blob/main/cedar-policy/src/frontend/is_authorized.rs#L146) and [RecvdSlice](https://github.com/cedar-policy/cedar/blob/main/cedar-policy/src/frontend/is_authorized.rs#L286) need to be duplicated in the calling code.
```

khieta (approved):
This sounds like a good idea to me! Could we use this to replace our current JSON interfaces in <https://github.com/cedar-policy/cedar/tree/main/cedar-policy/src/frontend>? (Currently that folder provides `json_is_authorized` and `json_validate`, which are used by [cedar-java](https://github.com/cedar-policy/cedar-java).)

khieta:
Hi @Swolebrain I discussed this with the team today, and we're all on board with deriving `JsonSchema` for more Cedar types (although we want to be careful about not accidentally exposing internal datatypes -- but we can iron this out when reviewing the PR). We also think it's a good idea to put this behind a flag so that users aren't required to take the dependency/extra compile time if they don't need it.

We also agreed that the RFC mechanism is a little heavy weight for this particular request, since it's not changing the language per se. We'd prefer using an issue to track this instead -- I've taken the liberty of porting to an issue here: https://github.com/cedar-policy/cedar/issues/492

Let me know if that sounds ok and I'll close the RFC. Thanks for submitting this!

PR #43, opened 2024-01-12
Revise any/all RFC to propose a simplified design.

emina:
[Rendered](https://github.com/cedar-policy/rfcs/blob/any-and-all-revised/text/0021-any-and-all-operators.md)

### DIFF
--- a/text/0021-any-and-all-operators.md
+++ b/text/0021-any-and-all-operators.md
@@ -1,4 +1,4 @@
-# Add `all` and `any` operators
+# Add basic `all` and `any` operators
 
 ## Related issues and PRs
 
@@ -18,14 +18,14 @@ This RFC proposes extending the language with `all` and `any` operators that all
 
 ## Basic example
 
-Consider writing a policy that allows access when all port numbers in the `context` are between 8000 and 8999.  This can't be expressed in Cedar today.
+Consider writing a policy that allows access when all port numbers in the `context` are greater than or equal to 8000.  This can't be expressed in Cedar today.
 
 With the `all` operator, we can write this policy as follows:
 
 ```
 permit(principal, action, resource)
 when {
-  context.portNumbers.all(it >= 8000 && it <= 8999)
+  context.portNumbers [all >= 8000]
 };
 ```
 
@@ -34,21 +34,21 @@ The same policy can be written using the `any` operator instead:
 ```
 permit(principal, action, resource)
 unless {
-  context.portNumbers.any(it < 8000 || it > 8999)
+  context.portNumbers [any < 8000]
 };
 ```
 
-The two operators behave analogously (though [not equivalently](#extending-the-semantics)) to `&&` and `||`. Given negation, each can be written in terms of the other. We include both for ergonomics.
+The two operators behave similarly (though [not equivalently](#extending-the-semantics)) to `&&` and `||`. Given negation, each can be written in terms of the other. We include both for ergonomics.
 
 ## Motivation
 
 The `all` and `any` operators support a common use case: iterating over a set and checking if a predicate holds for any or all members of the set.
 
-Cedar currently lacks a general way to express this use case. The only alternative is to expand the `all` or `any` checks into conjunctions or disjunctions. For example:
+Cedar currently lacks a way to express this use case. The only alternative is to expand the `all` or `any` checks into conjunctions or disjunctions. For example:
 
 ```
 // Using all
-["a", "ab", "abc"].all(it like "a*")
+["a", "ab", "abc"] [all like "a*"]
 
 // Using &&
 ("a" like "a*") &&
@@ -62,7 +62,7 @@ The new operators provide a built-in way to concisely express these universal an
 
 ## Detailed design
 
-The Cedar syntax, semantics, and type system are all extended to support the new operators. These extensions are backward compatible, and existing policies won't be affected. In particular, we will keep the `.containsAny` and `.containsAll` operators on sets, even though they can be expressed in terms of `any` and `all`.
+The Cedar syntax, semantics, and type system are all extended to support the new operators. These extensions are backward compatible, and existing policies won't be affected. The design of the operators is inspired by the [ForAllValues and ForAnyValue operators in the IAM language]().
 
 ### Extending the syntax
 
@@ -70,39 +70,48 @@ This RFC proposes extending the Cedar grammar as follows:
 
 ```
 Access    :=
-  '.' QUANTIFIER '(' Predicate ')'
-  | '.' IDENT ['(' [ExprList] ')']
+  '.' IDENT ['(' [ExprList] ')']
   | '[' STR ']'
+  | '[' QUANTIFIER Predicate ']'
 QUANTIFIER := 'all' | 'any'
-Predicate  := ...
+Predicate  :=
+  INTCMP Expr |
+  'like' PAT |
+  IDENT '(' [ExprList] ')'
+INTCMP := '<' | '<=' | '>' | '>='
 ```
 
-The `Predicate` expression differs from `Expr` in two ways:
+The extended grammar supports applying functions, integer comparisons, and the `like` operator to sets of values.
 
-1. Predicates can use the keyword `it`. A `Predicate` defines the body of function with a single parameter, and we use `it` as the parameter name. This is similar to [Kotlin's `it` keyword](https://kotlinlang.org/docs/lambdas.html#it-implicit-name-of-a-single-parameter).
-
-2. Predicates cannot contain nested `any` or `all` expressions. We disallow nested quantifiers for both performance and analyzability reasons. Nested quantifiers allow writing expressions that run in $O(2^n)$ time, where $n$ is the number of nested quantifiers. For example:
+For example:
 
 ```
-[0,1].all(
-  [0,1].all(
-    ...
-      [0,1].all(true)...))
+context.ips [any isLoopBack()]
+context.ips [all isInRange(ip("192.168.0.1/24"))]
+resource.tags [any like "private*"]
+resource.scores [all greaterThan(principal.preference)]
 ```
 
-Because nested quantifiers are prohibited, there is no need to support predicates with arbitrary parameter names. The `it` keyword sufficiently extends the grammar.
+The grammar doesn't support ooperating on sets of entities (e.g., using `in`) or sets of records (e.g., using `has`).  It also doesn't support combining multiple operators in a single `any`/`all` expression. Boolean combinations need to be expressed separately, outside of `any`/`all`.
+
+For example, this expression checks if all port numbers are between 8000 and 8999:
+
+```
+context.portNumbers [all >= 8000] &&
+context.portNumbers [all <= 8999]
+```
 
 ### Extending the semantics
 
-The semantics of `all` and `any` are straightforward when there are no errors:
-- `E.all(P)` evaluates to true when `P[e/it]` is true for every element `e` in `E`.
-- `E.any(P)` evaluates to true when `P[e/it]` is true for some element `e` in `E`.
+The semantics of `all` and `any` are straightforward when there are no errors.  For example:
+- `E [all INTCMP E']` evaluates to true when `e INTCMP E'` is true for every element `e` in `E`.
+- `E [any INTCMP E']` evaluates to true when `e INTCMP E'` is true for some element `e` in `E`.
 
-But what happens if evaluating `P` errors on some element `e`? In this case, the evaluation of `all` or `any` terminates abruptly with a distinguished `QuantifierError`.
+But what happens if evaluating `e INTCMP E'` errors on some element `e`? In this case, the evaluation of `all` or `any` terminates abruptly with a distinguished `QuantifierError`.
 
 The `QuantifierError` error may include additional diagnostic information, as long as this information is bounded in size and computed deterministically. For example, the `QuantifierError` error could specify the source location, or a fixed-length description of the smallest erroring value in the underlying set, according to some fixed total order on Cedar values. This error handling approach ensures that evaluating  `all` and `any` _deterministic_: the result is the same regardless of iteration order. It also ensures that evaluation is _space efficient_: the size of the evaluator output remains constant in policy and input size.
 
-For example, consider the expression `[1, true].all(it like "foo")`. Regardless of the order in which the predicate is applied, the expression returns the same `QuantifierError`.
+For example, consider the expression `[1, true] [all like "foo"]`. Regardless of the order in which the predicate is applied, the expression returns the same `QuantifierError`.
 
 Compare this to the [alternative semantics](#non-deterministic-error-behavior) that just propagates the error discovered, when it is discovered. Then evaluating this expression could lead to the error that "a Long was found where a String was expected" or it could lead to the error "A Boolean was found where a String was expected." Which one depends on evaluation order, but no clear order exists because set elements are specifically unordered.
 
@@ -111,7 +120,7 @@ The proposed semantics lets us think about `all` and `any` in terms of their exp
 For example:
 ```
 // Throws QuantifierError
-[1, true].all(it like "foo")
+[1, true] [all like "foo"]
 
 // Throws a TypeError: "like" can't be applied to an integer
 1 like "foo" && true like "foo"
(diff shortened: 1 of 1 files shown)


jszmajda:
I appreciate the need to add `any` and `all` to the language, and I support it. The general specification on how they work makes sense to me. The only issue I have is the use of square brackets to surround these keywords. These brackets break from the current language design in a fairly significant way. I understand the bracketing to allow for subsequent operators modifying `any` and `all` (e.g. `[any (operator) (value)]`. Would it break parsing too significantly to remove the need for the square brackets? For example, `context.portNumbers all >= 8000`?  I appreciated how the diff shows `any` and `all` implemented as functions on all types.. I agree with removing the `it` keyword, but `context.portNumbers.all(>= 8000)` is also more regular and preferred IMO. My preference is founded on the idea that a language with a smaller total symbology is easier to understand and use, and that that simplified understanding and use leads to simplified teaching, which is a net benefit.

text/0021-any-and-all-operators.md
```diff
@@ -18,14 +18,14 @@ This RFC proposes extending the language with `all` and `any` operators that all
...
 
 With the `all` operator, we can write this policy as follows:
 
 ```
 when {
-  context.portNumbers.all(it >= 8000 && it <= 8999)
+  context.portNumbers [all >= 8000]
 };
 ```
```
cdisselkoen, line 30: I'm a little sad that in this revision, we lost the convenient/nice syntax for examples like number-in-range.  Instead of a single `all` operation expressing that all the numbers are in a given range, we have to use two separate `all` operations, as discussed farther below in this RFC.  I realize that no expressivity is gained just by adding back `&&` and `||`, but I wonder if it would still be more ergonomic to support `&&` and `||` for use cases such as this?  Under the hood, we could even just desugar it into multiple `any` or `all` -- for instance,
```
context.portNumbers [all (it >= 8000 && it <= 8999)]
```
could literally desugar into
```
context.portNumbers [all >= 8000] &&
context.portNumbers [all <= 8999]
```
Thoughts on this?

text/0021-any-and-all-operators.md
```diff
@@ -62,47 +62,56 @@ The new operators provide a built-in way to concisely express these universal an
...
-    ...
-      [0,1].all(true)...))
+context.ips [any isLoopBack()]
+context.ips [all isInRange(ip("192.168.0.1/24"))]
+resource.tags [any like "private*"]
+resource.scores [all greaterThan(principal.preference)]
 ```
 
-Because nested quantifiers are prohibited, there is no need to support predicates with arbitrary parameter names. The `it` keyword sufficiently extends the grammar.
+The grammar doesn't support ooperating on sets of entities (e.g., using `in`) or sets of records (e.g., using `has`).  It also doesn't support combining multiple operators in a single `any`/`all` expression. Boolean combinations need to be expressed separately, outside of `any`/`all`.
```
cdisselkoen, line 95: Can we clarify the comment
 > doesn't support operating on sets of entities (e.g., using `in`)
 
 Are sets of entities unsupported entirely, or is it only `in` that is unsupported?  For instance, if there were an extension function that took an entity as an input, could we use that with `any`/`all`?

emina:
>  I agree with removing the `it` keyword, but `context.portNumbers.all(>= 8000)` is also more regular and preferred IMO.  

I prefer this version too and that was the original proposal we discussed :).  

But there was a concern that people would be confused by this, because it would be legal to write the following two expressions in Cedar, which are the same except for parentheses but mean totally different things:

```
context.numbers.all < 5   // using "all" as an attribute on an integer
context.numbers.all(< 5) // using "all" to iterate on a set of integers 
```

Do you think that's a concern?  Any ideas for how we could find out if others would find this potentially confusing?

text/0021-any-and-all-operators.md
```diff
@@ -62,47 +62,56 @@ The new operators provide a built-in way to concisely express these universal an
...
+  INTCMP Expr |
+  'like' PAT |
+  IDENT '(' [ExprList] ')'
+INTCMP := '<' | '<=' | '>' | '>='
 ```
 
-The `Predicate` expression differs from `Expr` in two ways:
-
-1. Predicates can use the keyword `it`. A `Predicate` defines the body of function with a single parameter, and we use `it` as the parameter name. This is similar to [Kotlin's `it` keyword](https://kotlinlang.org/docs/lambdas.html#it-implicit-name-of-a-single-parameter).
+The extended grammar supports applying functions, integer comparisons, and the `like` operator to sets of values.
```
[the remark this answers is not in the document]
emina: There is a cost to including equality (see [here](https://github.com/cedar-policy/rfcs/pull/43/files#diff-a95b565a911d6c60f2ef956742cd2798be3f159ff4716d1f80f51ecdb15649ddR238)).

And there is a significant cost (highly complex analysis) to including `in` (see [here](https://github.com/cedar-policy/rfcs/pull/43/files#diff-a95b565a911d6c60f2ef956742cd2798be3f159ff4716d1f80f51ecdb15649ddR239)).

We could add `is` (though I don't see much value) but I would strongly vote against providing any iteration over values that are themselves collections (sets and records).  I think it's simpler to say `any`/`all` apply to values that are atomic in Cedar: primitives and extension types (ip addresses and decimals for now).

text/0021-any-and-all-operators.md
jszmajda: I prefer removing `it` - it's an implicit-context keyword (e.g. if you know what "it" is from other languages, you grok it, but if you don't then it's another thing to explain). I think removing `it` and allowing for composition with `&&` and `||` allows for a more consistent overall language and reduced overall conceptual overhead.

jszmajda:
> But there was a concern that people would be confused by this, because it would be legal to write the following two expressions in Cedar, which are the same except for parentheses but mean totally different things:
> 
> ```
> context.numbers.all < 5   // using "all" as an attribute on an integer
> context.numbers.all(< 5) // using "all" to iterate on a set of integers 
> ```
> 
> Do you think that's a concern? Any ideas for how we could find out if others would find this potentially confusing?

I see the concern on `context.numbers.all < 5` and agree that that's not ideal. The parenthetical clears that up if it's required, but it may be lost. I still prefer not introducing square brackets. I understand that parsing the syntax of `context.numbers all < 5` involves detecting a keyword that isn't simple (e.g. as an infix operator), but I imagine the parsing wouldn't be that much harder to detect `<entity> (any|all) (<operator> <entity>|<function>)`?

mwhicks1 (approved):

max2me:
I share Jess' sentiment on syntax. I find using square brackets less intuitive and ergonomic, compared to the earlier approaches.

If I had a magic wand, I'd follow predicates syntax from JS/Typescript (both due to developers being familiar with it and also for future expansions it allows):
`context.numbers.all(it => it < 5)` where `it` can be a fixed (and the only allowed) name.

I am not concerned about possible confusion of whether `all` is name of an attribute or a name of the operator. As long as we have clear error messages when it's misused and documented trumping order, it will work well.

jszmajda:
> If I had a magic wand, I'd follow predicates syntax from JS/Typescript (both due to developers being familiar with it and also for future expansions it allows): `context.numbers.all(it => it < 5)` where `it` can be a fixed (and the only allowed) name.

I don't prefer this because it implies that Cedar might support lambdas, or would evaluate the expression by binding a variable and using it. There's no other variable binding in the language today, so I think this would open the door to a slippery slope.

text/0021-any-and-all-operators.md
emina: Yeah, I'm a bit sad about this one too.  

Another idea (and the syntax of this needs finessing) is to add an operator to the language that lets you express range constraints nicely.  @mwhicks1 says Ruby has an operator that does this nicely; we could take inspiration from their design.  This operator would be useful for Cedar anyway, so that we can write something like `context.number inRange [1..10]` versus having to write `context.number >= 1 && context.number <= 10`.

text/0021-any-and-all-operators.md
emina: Will clarify: `in` is the only problematic operator we have right now.  Others, such as `is`, are fine.

emina:
> I see the concern on `context.numbers.all < 5` and agree that that's not ideal. The parenthetical clears that up if it's required, but it may be lost. I still prefer not introducing square brackets. I understand that parsing the syntax of `context.numbers all < 5` involves detecting a keyword that isn't simple (e.g. as an infix operator), but I imagine the parsing wouldn't be that much harder to detect `<entity> (any|all) (<operator> <entity>|<function>)`?

@mwhicks1 What are your thoughts on this one?  Should we consider going back to `context.numbers.all (< 5)`? 

What about the alternative that Jess is proposing?  Do we think this is parseable?

For what it's worth, `context.numbers.all (< 5)` still looks most natural to me (like curried notation), due to my prior programming biases, which I agree may not be widely shared :)

hakanson:
I'm also not concerned about property or attribute confusion, because the parser handles that today on set based operations such as `contains()` - for example (in my schema) `resource.labels.contains("contest")` is valid but `resource.labels.contains == "contest"` generates this error (from 2.4.3):

> Unexpected type. Expected {"type":"Record","attributes":{},"additionalAttributes":true}, or {"type":"Entity"} but saw {"type":"Set","element":{"type":"String"}}

The `.all()` syntax feels more natural but not having used Kotlin, I wouldn't have though to use the `it` keyword. No significantly better ideas; could use `?` or `_` as placeholder instead (`context.numbers.all(? < 5)` or `context.numbers.all(_ < 5)`).

I also think `context.numbers.all(< 5)` is readable, but `context.numbers.all (< 5)` (with ` `) makes me think I had extra whitespace.

hakanson:
Another thought - given `.all()` / `.any()` look like functions on sets, would different naming similar to `.containsAll()` / `.containsAny()` be clearer (and leave future options for `any` or `all` keywords)?  Something like `exprAll()` or `evalAll()` that hints the parameter is an expression to be evaluated?

emina:
> Another thought - given `.all()` / `.any()` look like functions on sets, would different naming similar to `.containsAll()` / `.containsAny()` be clearer (and leave future options for `any` or `all` keywords)? Something like `exprAll()` or `evalAll()` that hints the parameter is an expression to be evaluated?

I have a bias for short names (`any`/`all`) myself but am curious to hear what others think.

jszmajda:
> I have a bias for short names (`any`/`all`) myself but am curious to hear what others think.

I also have a slight preference for shorter naming, but @hakanson does have a good point that `.containsAny()` and `.containsAll()` clarifies that these are functions on sets. So I suppose if we do not go forward with the operator syntax (`x any oper y`), then I'd slightly prefer `.containsAny()` / `.containsAll()`.

jszmajda:
> So if we want a longer name, we'd need some alternatives, like `.exprAll` and `.exprAny` suggested by Kevin.

ahh.. could `containsAll` etc be polymorphic? I think that would be less surprising and keep the language small

max2me:
I'd prefer then be more explicit as in `.allMatch()` / `anyMatch()` to differentiate from `containsAny/containsAll` while still maintaining continuity.

> I don't prefer this because it implies that Cedar might support lambdas
@jszmajda can you please help me understand this concern? Are you worried folks will try to declare reusable functions or..?

jszmajda:
> > I don't prefer this because it implies that Cedar might support lambdas
> 
> @jszmajda can you please help me understand this concern? Are you worried folks will try to declare reusable functions or..?

It implies that Cedar might add support for functions, or at least it drives a mental model of using a function, both of which are in conflict with (at least from what I understand) the goals of the language

emina:
> It implies that Cedar might add support for functions, or at least it drives a mental model of using a function, both of which are in conflict with (at least from what I understand) the goals of the language 

100%. that’s my thinking too. As a programmer, I’d also be disappointed if you let me a name local variable but then not use it freely, as proposed in the costly generalized design we want to avoid.

text/0021-any-and-all-operators.md
mwhicks1: Ruby has a "range" object: https://ruby-doc.org/core-2.5.1/Range.html. Instead of literally making such an object we could just steal its syntax as Emina has proposed, e.g., `context.portNum inRange (1..10)` or `context.portNum inRange (1...10)`, where `..` is "inclusive" (1 up to and including 10) and `...` is exclusive (1 up to but excluding 10).

mwhicks1:
> I understand that parsing the syntax of `context.numbers all < 5` involves detecting a keyword that isn't simple (e.g. as an infix operator), but I imagine the parsing wouldn't be that much harder to detect `<entity> (any|all) (<operator> <entity>|<function>)`?

I like Jess's proposed syntax if we are looking for a compromise, e.g., `context.portNumbers any >= 1000` rather than `context.portNumbers.any (>= 1000)` or `context.portNumbers [any >= 1000]`. I think this syntax is especially good when the LHS is a literal, e.g., `[1,2] any >= 1` as opposed to `[1,2] [any >= 1]` (even though this is not likely to be a common case as I mentioned in a previous comment).

From a parsing perspective I _think_ this should be Ok, but we might want to prototype it to make sure. I think we want to make sure it still leaves open the possibility of using `it` later, if we want/need to, e.g., `context.portNumbers any (it >= 1000 && it <= 2000)`.

To make clearer that this expression is boolean, I wonder if we can tweak the syntax even a bit more? 

For example:  `context.portNumbers ?any >= 1000` or `context.portNumbers any? >= 1000` ? A benefit of this approach is that we can make `?any` and `?all` keywords safely, without clashing with potential attribute names. For that matter, we could do `context.portNumbers.any? >= 1000` and then not require parentheses ...

emina:
> To make clearer that this expression is boolean, I wonder if we can tweak the syntax even a bit more?
> 
> For example: `context.portNumbers ?any >= 1000` or `context.portNumbers any? >= 1000` ? A benefit of this approach is that we can make `?any` and `?all` keywords safely, without clashing with potential attribute names. For that matter, we could do `context.portNumbers.any? >= 1000` and then not require parentheses ...

Of the proposed operator-like alternatives, I like `context.portNumbers.any? >= 1000` the best.  It looks like function application in an ML-like language :)

I’m not a fan of `?any` and `?all` because they look like template placeholders (`?principal` and `?resource`).

I’m also less enthusiastic about `context.portNumbers any >= 1000` because I’m having trouble pattern-matching it against constructs I know from other languages.  I can’t think of another infix operator that has multiple arguments on one of the sides (in this case, on the right).

emina:
If we’re still looking for a longer name for a method-call-like syntax, we could consider taking inspiration from XACML and go for `anyOf` and `allOf`.  So something like `context.portNumbers.anyOf(>= 1000)` or `context.ips.allOf(isLoopback())`. 

But I prefer the ML-like syntax that Mike suggested above:

```
context.portNumbers.any? >= 1000
context.ips.all? isLoopback()
```

max2me:
Sharing another option for completeness: we could adopt [CSS pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) like approach that would look like this:

```
context.portNumbers:any(>= 1000)
context.ips:all(isLoopback())
```

jszmajda:
I'm a fan of `object.any? <oper> <expr>` - I think that bridges the gap between challenging tradeoffs :)

text/0021-any-and-all-operators.md
jszmajda: I will say that Ruby's range operators are an incessant source of bugs and confusion—is it an inclusive or exclusive range? Is it a Range object or translated to an Array immediately? I've written so many bugs because of the syntax 😞  I found [this listing of various range syntaxes](https://rigaux.org/language-study/syntax-across-languages.html#VrsDatTpsRng) and honestly none of them are great though, so if Cedar does use a Range syntax (and I agree it's helpful), I'd argue for using either inclusive or exclusive but not both.

Stepping back from numeric ranges though, I imagine there there are other any/all set membership conditions that might want to be expressed that aren't numeric. I don't think `it` syntax really solves for those (and it also implies that Cedar would allow for variable binding..), but continued application of `&&`, `||` , and `!` certainly will. I'd prefer to avoid the pitfalls of ranges and also keep the doors open to more complex sets by relying entirely on setting ranges with logical operators. I also don't believe they're any harder to grok. IMO `element.all? >= 4 && element.all? < 8` is much clearer than `element.all? inRange [4..8]` (or is that `[4..9]`? or `[3..8]`? 😅)

mwhicks1:
> I'm a fan of `object.any? <oper> <expr>` - I think that bridges the gap between challenging tradeoffs :)

Me, too -- this has grown on me! As a possible next step can we prototype this syntax in the Cedar parser to explore its impact? Worries are always that (a) ambiguity, and (b) error messages.

emina:
> I'm a fan of `object.any? <oper> <expr>` - I think that bridges the gap between challenging tradeoffs :)

+1.  I like Mike's proposal to prototype the syntax.  I can change the text of this RFC to use the new syntax. 

Any strong objections to the `expr.any? <oper> <expr>` syntax?  

Some examples of what it looks like, replicated here for convenience:

```
context.portNumbers.any? >= 1000
context.ips.all? isLoopback()
```

@max2me @hakanson @cdisselkoen and anyone else following this thread :), let us know what you think.

hakanson:
I like it for the readablilty by that "policy reviewer" who hadn't read the language spec; they could intuit what it means.  

From RFC, `resource.tags.any? like "private*"` would be allowed.  Does `like` require a `*` wildcard or would `resource.tags.any? like "private"` be valid?  Why not `resource.tags.any? == "private"` ?  There is a note in the RFC that `==` isn't supported because of "O(n) vs O(1)" reasons, but that could be clearer on how `contains()` can be O(1) but an `any? ==` wouldn't be just as fast.

cdisselkoen:
Expanding on @hakanson's point, why couldn't we have `any? ==` just as fast as `contains()`?  For instance, why couldn't we just desugar `any? ==` to `.contains()` when creating the AST?

hakanson:
I tested this in 2.4.3
```
permit (principal, action, resource)
when { resource.if.in.has.then.else.like.is.true.false };
```

and got an error message like `poorly formed: This identifier is reserved and cannot be used: if` for all the identifiers except `is` which wasn't until 3.0?

**Question:** when `is` was added to Cedar, was it also added to this "identifier is reserved" list and does that impact any new identifiers (like `it`, 'all', 'any' from the original RFC?  `all? and `any?` are nice because they couldn't be identifiers.  Is that worth capturing in the RFC?

emina:
> I like it for the readablilty by that "policy reviewer" who hadn't read the language spec; they could intuit what it means.
> 
> From RFC, `resource.tags.any? like "private*"` would be allowed. Does `like` require a `*` wildcard or would `resource.tags.any? like "private"` be valid? Why not `resource.tags.any? == "private"` ? There is a note in the RFC that `==` isn't supported because of "O(n) vs O(1)" reasons, but that could be clearer on how `contains()` can be O(1) but an `any? ==` wouldn't be just as fast.

Yes, `resource.tags.any? like "private*"` would be allowed, and `like` behaves in this context in the same way as it does when applied to a single string.  So, `context.tag like "private"` is syntactically well-formed, but it's equivalent to `context.tag = "private"`.  In the case of sets,  `resource.tags.any? like "private"` is an inefficient way of writing `resources.tags.contains("private")`.  The former is O(n); the latter is O(1).

`contains()` is O(1) because it's a specialized operation on a set---a hash set lookup.  (Actually, in Cedar itself, it might be `O(log n)` if we're using a tree set.). But `.any? ==` is linear time because it applies the `==` operator to every element of the set individually, comparing it to the RHS.

We could implement rewrite rules (like @cdisselkoen suggests) that would turn `.any? ==` into `.contains`, but I would vote against that.  It adds cognitive overhead (when / why should I choose one way to say X versus another?).  And these rewrites are another potential source of bugs.

I prefer the solution @mwhicks1 proposed above:  
* When we see `E1.any? == E2`, throw an error that says "Use `E1.contains(E2)`".
* When we see `E1.all? == E2`, throw an error that says "Use `E1 == [E2]`.
* And, to your point about `like`, we could also throw an error or issue a warning if you write `E1.any? like pat` where pat has no wildcards.

emina:
> `all? and `any?` are nice because they couldn't be identifiers. Is that worth capturing in the RFC?

Yes!

mwhicks1:
One other point of discussion:

@cdisselkoen proposed `context.portNums any? >= 500` rather than `context.portNums.any? >= 500` on the basis that dropping the `.` makes `any?` more like other Cedar operators, like `like`. 

I prefer the proposal in the current RFC (with `.` notation), and here is my long-winded reasoning about that.

First, I observe that while `like` and `+` are binary operators that don't use `.` notation, set operators are different, e.g., we write `context.portNums.contains(1000)` or `context.portNums.containsAll([1000,1001])`. Thus `context.portNums.any?` is consistent with this syntax.

However, these set operators are basically being presented a method calls, where parentheses contain the single argument to the function. `any?` and `all?` have _two_ arguments, which are the operator (like `>=`) and the value for its RHS (like `1000`). For greater consistency, then, we could consider method-call syntax for `any?`/`all?`. But method calls normally comma-separate their arguments and `context.portNums.any?(>=,1000)` looks weird. We considered a slight variation earlier, `context.portNums.any?(>= 1000)` (no comma), but this also looks weird -- the `>= 1000` part looks like a single "argument" to the call, but then it is not a proper expression.

So, we have landed on `context.portNums.any? >= 1000` where `>=` and `1000` are the two arguments to the call, employing the style of curried function calls as in functional programming languages. That is, we can read `context.portNums.any?` as a function of type `(long -> long -> bool) -> long -> bool`.

How might we interpret the alternative without the `.`, i.e., `context.portNums any? >= 1000`? This would be treating `any?` as three-argument operator (of type `'a set -> ('a -> 'a -> bool) -> 'a -> bool`) but written infix, rather than prefix. The only example I can recall of a three-argument infix(ish) operator in an existing language is the turnstyle in C, `e1 ? e2 : e3`, which also includes a colon, and we don't want that. (Our adding of the `?` on `any`/`all` sort of evokes the turnstyle, though, now that I think of it.) It would be more conventional to write `any?  context.portNums >= 1000` for a three-argument function. But this syntax would be inconsistent with the other Cedar set operators by having the operator precede the set, rather than follow it, and I think it would surprise users.

So all of this is to say that I think we have landed on the right compromise:

1. Using the dot notation makes `any?` and `all?` consistent with other set operators
2. Avoiding parentheses and treating them as curried functions avoids the need for commas, which improves readability
3. Using `any?` and `all?` and not `any` and `all` avoids conflicts with normal identifiers which might be used as attributes, and also evokes that something unusual is going on, e.g., the use of currying

Now I am wishing that we had defined `contains?` and `containsAll?` rather than `contains` and `containsAll`, and that we didn't require parentheses around their arguments. Oh well. 😄

emina:
> Huge +1 to the currying analysis. My bias is that I like ML languages so.. take it with a grain of salt, but totally with you.
> 
> For `contains?` and `containsAll?` - could we also add those, and not remove the old ones? It broadens the surface area of the language, but in the name of least surprise.

I vote for keeping things small:  just adding `.any?` and `.all?` and keeping everything else as-is.

In addition to keeping things small, this approach also makes it clear that `.any?` and `.all?` are special:  they aren't like the other set operators (e.g., `contains`).  The other set operators are all first order.  And `.any?` and `.all?` are half-way to higher order functions :)  They are also more expensive to analyze.

I think it's helpful that they stand out visually too.

andrewmwells-amazon (approved):

apandays:
I believe the suggestion was to not support the "in" operator with "any" for a set. I think the usecase mentioned in this quip can be easily solved with relation-based policies if we support "in".

https://quip-amazon.com/h7lSAKMNM5h3/Cedar-Templates-as-Groups-Vice-or-Virtue

PR #45, opened 2024-01-29
add apply keyword RFC

philhassey:
[Rendered](https://github.com/strongdm/cedar-rfcs-apply/blob/main/text/0045-apply-keyword.md)


text/0045-apply-keyword.md
```diff
@@ -0,0 +1,119 @@
...
+outcome of the authorization event.  By taking into consideration the
+annotation, I am able to reverse the decision to `deny` if all matching policies
+are annotated with `@apply("after")`.
+
+This alternative does not work with the `forbid` case because any matching
+forbid statements will cause the entire request to be denied, regardless of if
+the principal should be permitted.
+
+This alternative does not support the general case of applying regardless of
+permit / forbid.
```
khieta, line 115: Hi @philhassey! Would this workaround work if the `is_authorized` function returned more information? Currently the `Response` will store the list of satisfied `permit` policies (for an Allow decision) or the list of satisfied `forbid` polcies (for a Deny decision). What if the `Response` instead specified all satisfied policies, regardless of decision?

text/0045-apply-keyword.md
```diff
@@ -0,0 +1,119 @@
...
+## Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- this does add the `apply` and `on` keyword to the grammar, thus increasing the
+complexity of the Cedar language
+- this may have some impact on the ability to short-circuit policy decisions,
+since all apply statements need to be checked to be added to the diagnostics
+
+## Alternatives
```
khieta, line 102: Can you also add some suggestions for alternative names? I'm personally partial to `log`, `info`, or `advise`, since we use "apply" in other contexts (e.g., to say that a policy "applies" for a particular request).

text/0045-apply-keyword.md
philhassey: Yes this could work, I would just throw away the given decision from `is_authorized` and come up with my own decision based on the annotations.

text/0045-apply-keyword.md
philhassey: Thanks, I've added a list of alternative name possibilities to the Alternatives section.

text/0045-apply-keyword.md
```diff
@@ -0,0 +1,134 @@
...
+1. Apply on permit
+
+```cedar
+@mfa("true")
+apply on permit (
+    principal,
+    action == Postgres::Action::"select",
+    resource == Postgres::Table::"example.com:5432/db"
+);
+```
```
cdisselkoen, line 56: For this example in particular, I'm not convinced this is a good motivating example.  The goal stated above was "I want to ... require MFA anytime someone accesses a certain table."  I think this should instead be done with something like `when { context.mfa == true }` on the `permit` policy(ies) in question; or if that is undesirable, then by adding a `forbid` policy along the lines of 
```
forbid ( 
    principal, 
    action == Postgres::Action::"select", 
    resource == Postgres::Table::"example.com:5432/db" 
) when { 
    context.mfa == false 
};
```

This expresses the authorization-relevant logic directly in the Cedar policy(ies), rather than requiring the application to correctly interpret policy annotations.  In general it is a best-practice to have all of the logic relevant to an authorization decision expressed in Cedar itself, not expressed in application logic surrounding the call to Cedar.

cdisselkoen (commented):
Thanks for this well-written and well-thought-out RFC!  One comment for now; I'll give a more full review later (perhaps after seeing more comments from others).

text/0045-apply-keyword.md
philhassey: Yeah, if the user was previously authenticated with MFA, I'd be able to include the MFA information in the context.

The case I'm describing is when you want to trigger an MFA after the cedar is evaluated, so I can have the provider pop up a window on your phone with the action "allow this query".  The result of the MFA then modifies the final decision.

philhassey:
I've added a few clarifications to the PR as well as added another variation of a possible non-language-changing workaround.

text/0045-apply-keyword.md
cdisselkoen: Cedar partial evaluation is designed for this use-case.  We can leave `context.mfa` unknown, and if the overall authentication decision ends up depending on `context.mfa`, partial evaluation will report an unknown with a residual like `context.mfa == true`.  Your application could take this unknown, realize it needs the user to authenticate with MFA, and then when appropriate retry the `is_authorized` call.

You could also do this without partial evaluation.  You could add the `forbid` policy I wrote above, but annotate it with something that tells you that if this `forbid` actually controls the decision, you should ask the user to authenticate with MFA.  Once they do, you can retry the `is_authorized` call.

philhassey: In the first case, when you say "your application could take this unknown" is that meaning, by reviewing the AST we'd be able to know what fields are unknown?  I hadn't read about the partial evaluation before, so that is interesting.  It seems like that could work, though I'd also want some annotation somewhere to indicate what the MFA prompt should be.  I'm imagining:

```
@mfa("query a table")
permit (
    principal, 
    action == Postgres::Action::"select", 
    resource == Postgres::Table::"example.com:5432/db" 
) when { 
    context.mfa == true 
};
```

In the second case are you saying the cedar would look like:

```
@mfa("query a table")
forbid ( 
    principal, 
    action == Postgres::Action::"select", 
    resource == Postgres::Table::"example.com:5432/db" 
) when { 
    context.mfa == false 
};

permit (
    principal, 
    action == Postgres::Action::"select", 
    resource == Postgres::Table::"example.com:5432/db" 
) when { 
    context.mfa == true 
};
```

So the forbid would give me the hint to gather the MFA, then to re-run the authorization afterwards.

If I'm capturing the concepts correctly, I can add them to the alternatives.

cdisselkoen: Yes, that's right.  In the second case the `permit` doesn't even need the `when` clause, because the `forbid` overrides it.  (The `when` clause doesn't hurt, of course.)

philhassey: I've added both of these alternatives to the RFC.  Thanks!

cdisselkoen:
The final comment period (FCP) for this RFC is starting now, with intent to reject the RFC.  Following the discussion above and additional offline discussion with the RFC author and the Cedar maintainers, we arrived at the conclusion that this functionality doesn't seem to belong as part of core Cedar, and the various use-cases can be better served by the alternatives described in the RFC text.

The FCP will end 2024-03-15 at noon PT / 3pm ET / 7pm UTC. Please add comments, and especially any objections, if you have any. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

PR #47, opened 2024-02-05
make `==` and `in` on `Unspecified` to be errors

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/unspecified-semantics/text/0047-unspecified-semantics.md)

### DIFF
--- a/text/0047-unspecified-semantics.md
+++ b/text/0047-unspecified-semantics.md
@@ -1,0 +1,178 @@
+# Make `==` and `in` on `Unspecified` to be errors
+
+## Related issues and PRs
+
+- Reference Issues: [cedar#217](https://github.com/cedar-policy/cedar/issues/217)
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2024-02-05
+
+## Summary
+
+The `==` and `in` operations will always produce errors when applied to
+`Unspecified` entities. This replaces the current behavior, which produces
+errors in some cases, `true` in other cases, and `false` in other cases.
+
+## Basic example
+
+Cedar's API allows you to leave the principal, action, or resource of a request
+unspecified, indicating that the principal, action, or resource "shouldn't
+matter" for any policy evaluation.
+Suppose `principal` is left unspecified in a request.
+
+1. If some policy contains the expression `principal == User::"alice"` or
+`User::"alice" == principal`, those expressions currently (Cedar 2.x and 3.x)
+evaluate to `false` -- unspecified entities are defined to not be equal to any
+entity in the hierarchy.
+2. If some policy contains the expression `principal in Group::"foo"` or
+`User::"alice" in principal`, those expressions currently (Cedar 2.x and 3.x)
+evaluate to `false` -- unspecified entities are defined to not be `in` any
+entity in the hierarchy.
+3. If some policy contains the expression `principal == principal` or
+`principal in principal`, those expressions currently (Cedar 2.x and 3.x)
+evaluate to `true` (even when `principal` is unspecified in the request).
+Unspecified entities are defined to be equal to themselves.
+4. If some policy contains the expression `principal has foo`, that expression
+currently (Cedar 2.x and 3.x) evaluates to `false` -- unspecified entities are
+defined to have no attributes.
+4. If some policy contains the expression `principal.foo`, that expression
+currently throws an error.
+
+With this RFC, we change the behavior to be consistent: _All_ of the above
+examples will throw an error.
+
+## Motivation
+
+The behavior of unspecified entities is complicated, unintuitive, and has to
+be special-cased frequently in the validator.  Even we, as Cedar developers, are
+often confused or tripped up by the semantics of `Unspecified`.
+The distinction between unspecified entities and partial-evaluation unknowns is
+also subtle and confusing and has caused bugs.
+
+The proposed semantics is more consistent for Cedar users, as outlined above in
+"Basic example", and hopefully also more intuitive for users (reduce surprises).
+The original intended use-case for `Unspecified` was that a caller who leaves a
+variable (say `principal`) unspecified, is implicitly guaranteeing that
+`principal` does not matter to the authorization outcome (e.g., because no
+policies check `principal`).
+This RFC's proposed semantics is more true to this intended use-case than the
+current semantics.
+
+Additionally, the proposed semantics will allow us to simplify Cedar's
+implementation by unifying the behavior of `Unspecified` and `Unknown` in many
+ways, allowing us to share code for these two features.
+
+## Detailed design
+
+### Semantics
+
+The detailed semantics is clear and simple: if `principal` (resp. `action` or
+`resource`) is Unspecified in the request, it is an evaluation error for
+`principal` (resp. `action` or `resource`) to appear in any (evaluated)
+expression in the policy.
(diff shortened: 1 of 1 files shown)


text/0047-unspecified-semantics.md
```diff
@@ -0,0 +1,178 @@
...
+evaluate to `true` (even when `principal` is unspecified in the request).
+Unspecified entities are defined to be equal to themselves.
+4. If some policy contains the expression `principal has foo`, that expression
+currently (Cedar 2.x and 3.x) evaluates to `false` -- unspecified entities are
+defined to have no attributes.
+4. If some policy contains the expression `principal.foo`, that expression
+currently throws an error.
+
+With this RFC, we change the behavior to be consistent: _All_ of the above
+examples will throw an error.
```
mwhicks1, line 44: I presume the use of `is` on an unspecified entity also results in an error. Any other operators to mention?

text/0047-unspecified-semantics.md
```diff
@@ -0,0 +1,178 @@
...
+`resource`) is Unspecified in the request, it is an evaluation error for
+`principal` (resp. `action` or `resource`) to appear in any (evaluated)
+expression in the policy.
+(The scope constraints `principal,` `action,` and `resource,` are always still
+allowed of course.)
+
+From the evalautor perspective, we must still allow unspecified variables to
+appear in unevaluated/dead code, because the evaluator simply does not evaluate
+those expressions due to short-circuiting; but despite this, validator can still
+soundly prohibit these variables in all expressions in the policy.
```
mwhicks1, line 81: I don't understand what this means. I think you mean that if we evaluate `if true then true else principal has owner` and `principal` is Unspecified, this will still evaluate to `true`. I feel like this goes without saying?

text/0047-unspecified-semantics.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Timeline
+
+- Started: 2024-02-05
+
+## Summary
+
+The `==` and `in` operations will always produce errors when applied to
+`Unspecified` entities. This replaces the current behavior, which produces
+errors in some cases, `true` in other cases, and `false` in other cases.
```
mwhicks1, line 16: Then: It's now possible for expressions in the scope to have an error. I imagine this affects how we do policy slicing. Have you thought that through?

text/0047-unspecified-semantics.md
cdisselkoen: Today, `is` always returns `false` for unspecified entities.  This RFC proposes an error.  I will adjust the RFC text, good catch

text/0047-unspecified-semantics.md
cdisselkoen: You are correct that's what it means.  Perhaps it goes without saying, but because of the language "appears in the policy" in the previous policy, I thought it was safer to clarify.  This is a Detailed Design section after all

text/0047-unspecified-semantics.md
cdisselkoen: Good catch, I have not thought this through

emina: The soundness of slicing hinges on the fact that scope constraints cannot error.  We should not violate this invariant without a very good reason.

The current behavior of `Unspecified` is consistent with what you would get if you declared a special type with no attributes that is distinct from all other types, and assigned your principal to a dummy value of this special type.  I don't think the semantics of this is confusing or hard to explain.

If we're going to make a breaking change, then I would vote for completely dropping the support for `Unspecified`.

text/0047-unspecified-semantics.md
cdisselkoen: It's a little less clear whether your snippet should still validate.  My text points out that we could write the validator to produce an error for your snippet.  We could also write the validator to validate your snippet.  Both are sound

text/0047-unspecified-semantics.md
emina: A naive question:  why do we need to have any special handling of `Unspecified`?  Can we drop it from validation, PE, etc., and just leave runtime APIs that that implement `Unspecified` as literally a value of type with no attributes, with the value generated by the Rust code? 

If we choose a type name that is not a valid Cedar identifier, that type is guaranteed to be distinct from all other types.

emina: Basically, can we sort of have our cake (runtime support for `Unspecified`) and eat it too (no special handling in the tooling)?

emina: Admittedly, it's disappointing to have runtime support but not tooling support.  So maybe the cleanest course of action is to drop `Unspecified`.

cdisselkoen: We originally implemented `Unspecified` basically that way.  However, rather than an illegal entity-type identifier, we thought it much clearer and error-proof to use an `Option` type.  It was thought that forcing users of `EntityType` to explicitly consider the `Unspecified` case was a virtue and would lead to fewer bugs, particularly in validation/PE/etc.

cdisselkoen: Consider that the validator has to consider the possibility that values exist of type `Unspecified`, even though the `Unspecified` type is not declared in the schema.  With the `Option` or explicitly-enumerated type, the validator basically cannot forget this.  If `Unspecified` was simply implemented as a value with illegal entity type, the validator could easily forget to handle the case for that entity type.

emina: Do we have a way to estimate the impact of dropping `Unspecified`?

emina: > Consider that the validator has to consider the possibility that values exist of type `Unspecified`, even though the `Unspecified` type is not declared in the schema. With the `Option` or explicitly-enumerated type, the validator basically cannot forget this. If `Unspecified` was simply implemented as a value with illegal entity type, the validator could easily forget to handle the case for that entity type.

My naive thought is that we would automatically extend _every_ user schema with the special `Unspecified` type(s) and then just run normal validation against this extended schema.

Where does this naive approach break down?

mwhicks1: Why do we need to do that? Why shouldn't users define their own "unspecified" types when they want to make up a value to pass it in? Then they can include this type in the schema, which will make clear it's a member of nothing and has no attributes?

Maybe this conversation belongs in a new RFC ...

khieta: I think the proposals in this thread (switch to treating unspecified entities as entities with a unique type for evaluation and validation) are viable. This is originally how we did it, but we then switched to encoding this information at the type level because we thought this would be less error prone (hah!).

I worry that making this change in the validator will cause us to change precision in some cases, which could lead to breaking changes. But I'd have to look at the code more carefully to be sure.

emina: Right, the above proposal is a mechanical way to simulate exactly that:  users defining their own "unspecified" types in the schema and having actions refer to those unspecified types.

I also think it's a cleaner design for users to do this.

The proposal is a way to avoid making a breaking change if we decide that getting rid of `Unspecified` (preferred) is not an option at this time.

john-h-kastner-aws: I don't immediately see a problem with handling unspecified as an implicit type in the schema if the semantics actually match that.

cdisselkoen:
Closing this due to discussion above.  In particular, this RFC would violate our design constraint that policy scopes cannot error.

PR #48, opened 2024-02-05
Cedar Schema Annotations

aaronjeline:
[Rendered](https://github.com/cedar-policy/rfcs/blob/schema-annotations/text/0048-schema-annotations.md)


text/0048-schema-annotations.md
```diff
@@ -0,0 +1,140 @@
...
+```
+With an arbitrary number of them being able to prepend to a top level declaration in a schema.
+
+Thus the full schema syntax becomes:
+```
+Schema      := {Namespace}
+Namespace   := ('namespace' Path '{' {Decl} '}') | {Decl}
+Decl        := Entity | Action | TypeDecl
+Entity      := Annotations 'entity' Idents ['in' EntOrTyps] [['='] RecType] ';'
+Action      := Annotations 'action' Names ['in' (Name | '[' [Names] ']')] [AppliesTo] [ActAttrs]';'
```
john-h-kastner-aws, line 75: Annotations on actions could be a bit confusing when annotations aren't reflected in any way at run time. It kind of feels like a second set of pseudo-attributes.

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,140 @@
...
+This lets us standardize certain annotations, like `doc`. 
+This probably can't happen unless we also do this for policies, which we've said we don't want to do.
+### Doc Strings as comments
+Instead of annotations, we could add "doc-strings" as a first class feature.
+Could look like this:
+```
+/# Stop users from accessing a high security document unless:
+/#  A) The principal and user are at the same location
+/#  B) The principal has a job level greater than  4
+forbid(principal, action, resource) when {
```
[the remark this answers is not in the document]
aaronjeline: No, the docs case was what was asked for by someone in the slack, and I was seeing if people preferred the more general approach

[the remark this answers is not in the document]
john-h-kastner-aws: With the customer schema syntax I do think we'll see more requests for annotations in the schema eventually. It would feel natural to use the same annotation mechanism in policies and schema.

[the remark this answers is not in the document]
aaronjeline: No reason we can't do both annotations and doc comments

nakedible-p:
Comments below. All *my* opinions only! I don't want to say IMHO in every step, so I'm just writing them in assertive form.

- Schema is 50% validation, 50% documentation. Simply allowing a single line `@doc("This is what this is for")` is woefully inadequate to what is actually needed. If we look at GraphQL, OpenAPI, JSON Schema, Rustdoc, etc. then a documentation is usually of the form (regardless of actual syntax):

```
# Main explanation of the item

More descriptive explanation of the item, with much more detail in a multiline and multi paragraph format.

## Examples

How policies using this usually look.

## Warnings

Here's what to keep in mind when using the item.
```

- All items need documentation. Your syntax is meant to be compact in defining a set of actions so you don't need to repeat groups and context and attributes on every action, but you still need documentation for each item. Each attribute needs documentation. Each context item needs documentation.
- Think about formatting as well. Schema shouldn't depend on whitespace and you try to make it look nice, but do think how it will actually look when fully formatted out. Right now, you don't allow nearly enough extra commas, grouping parenthesis, etc. to make it natural to write something that isn't just a single line. For example, this is from a WIP schema using comments to document that I've tried to format as convenient and clear as possible:
```
  // Bootstrap actions that are allowed for both authenticated and enrolling terminals.
  action
    // Can establish an RPC connection
    //
    // This action merely gives permission to establish an RPC connection, but all
    // further access is checked separately.
    ptRpcConnect,

    // Can ask for new authentication tokens
    //
    // Enrolling terminals use this to ask for proper authentication tokens. Access
    // control for this is limited to a shared secret, where as the granted tokens
    // allow full access.
    ptTokenGrant
  in [
    ptBootstrapActions
  ] appliesTo {
    principal: [PaymentApplication, BootstrapApplication]
  } attributes {
    // Terminal hardware this application is running on
    hardware: Hardware,
  };
```
- Consider if it's better to have the documentation precede the item or to be after the item in question. Both approaches are seen. Python docstrings are after, as are OpenAPI and JSON Schema description fields, where as GraphQL, Rustdoc, etc. have documentation preceding.
- Whatever way is chosen, assume tools need to be able to generate HTML documentation from the schema, so strict syntax and programmatic way to parse that is crucial. Also consider that there is need for a top-level documentation - perhaps top level for a schema file, or just top-level for a certain namespace.

Perhaps a good example on how Cedar schemas with documentation should look like is to look at the GraphQL schema of GitHub: https://docs.github.com/public/schema.docs.graphql

Swolebrain:
@nakedible-p
What would you say to the idea of adding cedar functionality to graphql/openAPI doc formats as opposed to re-creating what they have? This is all high-level in my mind but i imagine a world where im defining a graphql schema and i can specify something like:

```

query GetBooks @cedarAction('GetBooksWithAuthor', resource=Book) {
  books {
    title
    author {
      name
    }
  }
}
```

That way, as I define my queries and mutations i can define the actions and resources that those queries map to.

nakedible-p:
You already have this "embed authentication in graphql schema" thing with AWS Amplify: https://docs.amplify.aws/javascript/build-a-backend/graphqlapi/customize-authorization-rules/ It's not Cedar, but a very similar idea. Probably one day it can be Cedar as well.

nakedible-p:
Here is a basic JSON schema with descriptions: https://json-schema.org/learn/miscellaneous-examples#basic

max2me:
Am I reading it correctly that this proposal would not support annotation individual attributes?

text/0048-schema-annotations.md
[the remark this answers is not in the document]
cdisselkoen: In Rust, doc comments are actually just sugar for Rust's equivalent of annotations.  (Specifically, sugar for the [`#[doc]` attribute](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html).)  We could consider doing something similar.  What that means for this RFC is that we could support just doc comments now, and later if we introduce general annotations we could have doc comments desugar into a special reserved annotation.  Or, we could support general annotations now, and later we could introduce doc comments as a sugar on top.  We don't have to worry about closing off one or the other path.

text/0048-schema-annotations.md
cdisselkoen: I agree that I think annotations on entity types and common types are more clear than annotations on actions.

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+```
+@doc("Stop users from accessing a high security documents unless they are in the same building")
+forbid(principal, action, resource) when {
+    resource.security_level == "HIGH"
+unless {
+    resource.location == principal.location
+};
+```
```
mwhicks1, line 37: This is a policy not a schema. Why is it here? I'd delete to avoid confusion.

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,139 @@
...
+Like Cedar policies, users may want to associate arbitrary, machine readable metadata with Schema objects.
+We solved this problem in Cedar Policies by allowing for *annotations*: arbitrary key/value pairs that are attachable to policies.
+This could be extended to Cedar Schemas, allowing users to attach attributes an entity type, common type, and action definitions.
+
+
+## Basic example
+
+1. Doc comments
+```
+@doc("This entity defines our central user type")
```
mwhicks1, line 23: You have a number here as if this is the first of several examples, but it's the only one.

For doc comments, it makes more sense to me to define an opinionated literate format. This is what Rust, Java, etc. all do, and it's nicer to read and perhaps more natural to write (per your Alternative "Doc strings as comments"). But we could still _also_ have arbitrary annotations to handle other use-cases (which it would be nice to see, here).

text/0048-schema-annotations.md
mwhicks1: For the doc-comments use-case, you want annotations pretty much everywhere, e.g., to describe what actions do, right?

mwhicks1 (commented):
I'd like to see this fleshed out to consider more use-cases. If doc strings are the only use-case, then a comment-based approach seems best to me. But we surely have other use-cases in mind, too, but we need to see them to evaluate possible designs.

max2me:
Can we please consider supporting annotation on individual attributes as well? This would allow integrating tools and services to embed data that helps authoring policies easier.

For example:

```
@doc("This entity defines our central user type")
entity User { 
    manager : User,

    @doc("Which team user belongs to")    
    @docLink("https://schemaDocs.example.com/User/team")
    team : String
};
```

text/0048-schema-annotations.md
cdisselkoen: That is true.  As noted in a different thread, I think we want to allow annotations (particularly doc comments) on individual attributes of entity/common types as well.

text/0048-schema-annotations.md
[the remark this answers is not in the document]
hakanson: For a doc comment syntax, I think `//#` is preferable to `/#` since backwards compatible with valid line comments, and similar to other syntaxes for overloading comments.   An example is `//!` , the esbuild [legal comment](https://esbuild.github.io/api/#legal-comments) syntax to preserve in output files .

```
//! Copyright Cedar Contributors
//! SPDX-License-Identifier: Apache-2.0
```

khieta:
I think #63, which proposes to support doc comments in schemas and policies, gives a better solution for the `@doc(...)` case -- do we still want schema annotations even if we already have schema doc comments?

emina:
> I'd like to see this fleshed out to consider more use-cases. If doc strings are the only use-case, then a comment-based approach seems best to me. But we surely have other use-cases in mind, too, but we need to see them to evaluate possible designs.

One (real customer) use case is to share the same schema among several organizations, while allowing UI builders to display only parts of the schema that are relevant to a specific organization.  For example, entity types A and B should show up in UI builders only for members of the organization X.  And the UI builder extracts this association from the annotation `@VisibleTo("X")` on entity types A and B.

nakedible-p:
Anothere example from AWS Amplify land on GraphQL:
```graphql
type Customer @model @auth(rules: [{ allow: public }]) {
  id: ID!
  name: String! @index(name: "byNameAndPhoneNumber", sortKeyFields: ["phoneNumber"], queryField: "customerByNameAndPhone")
  phoneNumber: String
  accountRepresentativeID: ID! @index
}
```
Even though such things are not directly applicable, it doesn't take much imagination to figure out similar things applied to a Cedar schema.

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,131 @@
...
+```
+Annotation := '@' IDENT '(' STR ')'
+Annotations := {Annotations}
+```
+With an arbitrary number of them being able to prepend to a top level declaration in a schema.
+
+Thus the full schema syntax becomes:
+```
+Schema      := {Namespace}
+Namespace   := ('namespace' Path '{' {Decl} '}') | {Decl}
```
john-h-kastner-aws, line 64: Based on some recent discussion, we might want to support annotations on the namespace. I don't see a reason not to.

text/0048-schema-annotations.md
john-h-kastner-aws: We probably do want annotations basically everywhere. It's _possibly_ confusing on actions, but leaving it out is guaranteed to annoy users.

shaobo-he-aws:
> Can we please consider supporting annotation on individual attributes as well? This would allow integrating tools and services to embed data that helps authoring policies easier.
> 
> For example:
> 
> ```
> @doc("This entity defines our central user type")
> entity User { 
>     manager : User,
> 
>     @doc("Which team user belongs to")    
>     @docLink("https://schemaDocs.example.com/User/team")
>     team : String
> };
> ```

Sure, we can. Just to double check. Are we looking into supporting entity attributes only? Note that attributes can also be found in `context` declarations.

max2me:
I support this proposal as it will streamline the development of improved policy authoring tools while avoiding the need to  separate configurations for schema meta data.

text/0048-schema-annotations.md
```diff
@@ -122,6 +123,92 @@ The same fact also applies to entity attribute declarations.
...
+A corresponding JSON schema for the above example is as follows.
+```JSON
+{
+    "": {
+    "annotations": {
+        "doc": "this is the namespace"
+    },
+    "commonTypes": {
+        "Task": {
+        "annotations": {
```
john-h-kastner-aws, line 135: I assume this matches how we do annotations in the EST?

john-h-kastner-aws (approved):
Looks good. I support this extension

text/0048-schema-annotations.md
shaobo-he-aws: We only allow annotations on policy level in ESTs. So they will on the same level as, e.g., `"effect"`.

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,242 @@
...
+PRIMTYPE    := 'Long' | 'String' | 'Bool'
+WHITESPC    := Unicode whitespace
+COMMENT     := '//' ~NEWLINE* NEWLINE
+```
+
+### JSON Syntax
+None of the three top-level constructs (EntityTypes, Actions, CommonTypes) in JSON schemas allow for arbitrary key/value pairs.
+This means a new key can be safely added while preserving backwards compatibility.
+The same fact also applies to entity attribute declarations.
+This proposal reserves the `annotations` key at the top level of each of those constructs, which contains an Object, containing each annotation key as an Object key, associated with the annotation string.
```
andrewmwells-amazon, line 123: Maybe we could use `__annotations` if this is a concern

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,140 @@
...
+    resource.security_level == "HIGH"
+unless {
+    (resource.location == principal.location) || (principal.job_level > 4 )
+};
+```
+This has nice and easy multi-line syntax, but is special cased and not as general.
+
+## Unresolved questions
+
+* If this RFC is accepted, do we want to patch [RFC24 (Schema Syntax](https://github.com/cedar-policy/rfcs/issues/24") to contain the full new grammar so it's in one place?
```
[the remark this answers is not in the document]
cdisselkoen: Bumping this, let's remove this from `Unresolved questions` since it seems to be resolved

cdisselkoen (approved):

text/0048-schema-annotations.md
shaobo-he-aws: Maybe we can also use reserved `__cedar::schema::annotations` to be 100% secure.

andrewmwells-amazon (approved):

emina (commented):
I support this proposal.

text/0048-schema-annotations.md
```diff
@@ -0,0 +1,238 @@
...
+
+The Cedar spec takes no opinion or stance on the interpretation of annotations.
+The interpretation is entirely up to users of Cedar.
+
+### Cedar Schema Syntax
+Attributes in Cedar Schemas will mirror the syntax used for attributes in a policy: informally that's `@<key>("value")`.
+Formally the following rule is added to the Cedar grammar:
+```
+Annotation := '@' IDENT '(' STR ')'
+Annotations := {Annotations}
```
micahhausler, line 82: Should this be
```
Annotations := {Annotation}
```

shaobo-he-aws: Good catch. Will update it.

PR #49, opened 2024-02-06
Small change to schema syntax

aaronjeline:
Small change to how actions look, removes the `appliesTo` keyword
[Rendered](https://github.com/cedar-policy/rfcs/blob/0024-change/text/0024-schema-syntax.md)


text/0024-schema-syntax.md
```diff
@@ -39,11 +39,15 @@ entity List in [Application] {
...
+action CreateList, GetLists { 
+    principal: [User], 
+    resource: [Application] 
+};
 
-action GetList, UpdateList, DeleteList, CreateTask, UpdateTask, DeleteTask, EditShares
-    appliesTo { principal: [User], resource:[List] };
+action GetList, UpdateList, DeleteList, CreateTask, UpdateTask, DeleteTask, EditShares { 
+    principal: [User], 
+    resource:[List]
```
cdisselkoen, line 49:
```suggestion
    resource: [List] 
```

text/0024-schema-syntax.md
```diff
@@ -183,7 +191,7 @@ action ViewDocument in [ReadActions]
   attributes {
     version: 1,
     group: "meta"
-  };
+  }
```
cdisselkoen, line 194: why this? don't we still need the closing `;`?

cdisselkoen (approved):
Looks good

text/0024-schema-syntax.md
khieta: Also, I think the braces are mismatched -- `attributes` should be at the same nest level as `principal` etc.

aaronjeline: No, I don't think it can be, otherwise how to specify an "empty" body.

cdisselkoen:
```
action View {
  attributes: {
    readOnly: true
  }
};
```
?  

ie it just omits `principal`, `resource`, and `context`

cdisselkoen: the alternative `action View {} { attributes: ... };` is also fine with me though

cdisselkoen: er I mean `action View {} attributes { ... };`

andrewmwells-amazon (approved):

aaronjeline:
Decided against this small tweak.

PR #50, opened 2024-02-08
RFC 50: Refine the parsing rule of multiply-by-constant operation

shaobo-he-aws:
[Rendered](https://github.com/cedar-policy/rfcs/blob/multiply-by-const/text/0049-parse-multiply-by-constant.md)

### DIFF
--- a/text/0049-parse-multiply-by-constant.md
+++ b/text/0049-parse-multiply-by-constant.md
@@ -1,0 +1,60 @@
+# Refine the Parsing Rule of Multiply-By-Constant Operation
+
+## Related issues and PRs
+
+- Reference Issues: https://github.com/cedar-policy/cedar/issues/619
+- Implementation PR(s): (leave this empty)
+
+## Timeline
+
+- Started: 2024-02-07
+
+## Summary
+
+This PR proposes that the parser checks validity of multiplication operations after left-associativity is explicitly applied on the CST level. And the order of resulting AST should be the same as that of CST.
+
+## Basic example
+
+`2*1*context.value` should be conceptually converted into `((2*1)*context.value)` before determining if it's a valid multiply-by-constant operation. And the resulting AST should be `(mul (mul 2 1) context.value)`.
+
+## Motivation
+
+Let's first revisit the status quo: The parser separates the operands of a CST representing a (possibly chained) multiplication operation into a non-integer-constant and a list of integer constants, preserving the order of the constants, and then *left-fold* the constant list with the non-constant. For instance, `2*1*content.value` is parsed into `(mul (mul content.value 2) 1)`.
+
+Note that the parser changes the order of operands, which does not alter the evaluation result since Cedar's multiplication operation is associative. However, changing the evaluation order can make error message confusing: For instance, evaluating `2*6*context.value` when `context.value` is `i64::MAX/10` produces an error message like ```integer overflow while attempting to multiply `1844674407370955160` by `6```, contradicting the intuitive left-to-right evaluation order.
+
+The parser's rule to determine if multiplication operations are valid is also confusing. For instance, `2*1*context.value` is valid whereas its variant `(2*1)*context.value` is not. The inconsistency further causes ambiguity in EST to AST conversion: ESTs of the aforementioned example are the same but only the former can be parsed to an AST.
+
+## Detailed design
+
+To resolve the issues, this PR proposes a design to explicitly apply Cedar's left-associativity rule on the CST level and determine validity afterwards. The CSTs, if deemed valid, should be converted to ASTs that have the same order. Note that there will be only two operands for the CST representation of a multiplication operation, as opposed to a list of operands per status quo.
+
+The validity checking algorithm is simple: Count the non-constant-integer operands of a CST representing a multiplication operation recursively and error if it's greater than 1. An example algorithm is as follows.
+
+```Rust
+fn count_non_constant_operands(lhs, rhs) -> bool {
+    match (lhs, rhs) {
+        (cst::Mul(ll, lr), cst::Mul(rl, rr)) => count_non_constant_operands(ll, lr) + count_non_constant_operands(rl, rr),
+        (cst::Mul(ll, lr), Lit(Int(_))) => count_non_constant_operands(ll, lr),
+        (cst::Mul(ll, lr), _) => count_non_constant_operands(ll, lr) + 1,
+        (Lit(Int(_)), cst::Mul(rl, rr)) => count_non_constant_operands(rl, rr),
+        (_, cst::Mul(rl, rr)) => count_non_constant_operands(rl, rr) + 1,
+        (Lit(Int(_)), Lit(Int(_))) => 0,
+        (Lit(Int(_)), _) => 1,
+        (_, Lit(Int(_))) => 1,
+        (_, _) => 2,
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


text/0049-parse-multiply-by-constant.md
```diff
@@ -0,0 +1,60 @@
...
+
+This recursive algorithm validates expressions derived from arbitrary operand associations of a chained multiplication operation. For instance, expressions `1*2*context*3`, `1*(2*context*3)`, and `1*(2*context)*3` are all accepted.
+
+## Drawbacks
+
+The major drawback of this approach is that it induces a breaking change.
+
+## Alternatives
+
+An alternative is to let the ESTs representing multiplication operations to a list of operands, just like what CSTs do.
```
emina, line 58: Isn’t the alternative also a breaking change?  Plus, the alternative could also make the EST API harder to use.  For example, which way are the clients of the EST going to associate the list of multiplication elements?

Are there any fixes to #619 that don’t introduce breaking changes?  For example, is it possible to change the CST -> EST converter to produce an EST that associates multiplication in exactly the same way as the CST -> AST converter?  That way, CST -> EST -> AST conversion would fail only if CST -> AST conversion would for the same CST.

text/0049-parse-multiply-by-constant.md
```diff
@@ -0,0 +1,60 @@
...
+        (_, cst::Mul(rl, rr)) => count_non_constant_operands(rl, rr) + 1,
+        (Lit(Int(_)), Lit(Int(_))) => 0,
+        (Lit(Int(_)), _) => 1,
+        (_, Lit(Int(_))) => 1,
+        (_, _) => 2,
+    }
+}
+```
+
+This recursive algorithm validates expressions derived from arbitrary operand associations of a chained multiplication operation. For instance, expressions `1*2*context*3`, `1*(2*context*3)`, and `1*(2*context)*3` are all accepted.
```
john-h-kastner-aws, line 50: The pseudo-code here doesn't quite line up with our actual CST and you've not listed this specific exmaple, so it's not clear to me if the proposed algorithm is intended to count `(2*1)*context.value` as having 1 non-constant operand. Presumably it should?

text/0049-parse-multiply-by-constant.md
john-h-kastner-aws: We could likely do this as an extension to the EST grammar without a breaking change. I do agree that having both options for multiplication makes the EST API more complicated and somewhat harder to used.

My understanding is that changing CST->EST alone wouldn't be enough. The expression `(1 * 2)` in an EST is treated as non-constant, so we would also need to update the EST to AST step (but in an obvious way).

text/0049-parse-multiply-by-constant.md
shaobo-he-aws: Good catch. It seems we had to peel the CST of `(2*1)` all the way down to a `cst::Mul`.

text/0049-parse-multiply-by-constant.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Drawbacks
+
+The major drawback of this approach is that it induces a breaking change.
```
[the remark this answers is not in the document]
shaobo-he-aws: @mwhicks1 had the same observation. I'll update the RFC.

text/0049-parse-multiply-by-constant.md
shaobo-he-aws: To expand John's comment: `(1*2)*true`'s CST to AST conversion fails whereas `1*2*true`'s doesn't. But both of them have the same EST because we do not distinguish `(1*2)` and `1*2` on the EST level. If we want to ensure that if CST->AST fails, then CST->EST->AST also does, we may need to check validity (i.e., if there's more than one non-constant) at CST to EST conversion and also operator association. We still need to also check validity during EST to AST conversion because otherwise the users could create something like `principal.value * context.value`. And we want the two checks consistent.

text/0049-parse-multiply-by-constant.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Started: 2024-02-07
+
+## Summary
+
+This PR proposes that the parser checks validity of multiplication operations after left-associativity is explicitly applied on the CST level. And the order of resulting AST should be the same as that of CST.
+
+## Basic example
+
+`2*1*context.value` should be conceptually converted into `((2*1)*context.value)` before determining if it's a valid multiply-by-constant operation. And the resulting AST should be `(mul (mul 2 1) context.value)`.
```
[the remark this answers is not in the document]
shaobo-he-aws: Done.

text/0049-parse-multiply-by-constant.md
shaobo-he-aws: Please see the revised algorithm.

text/0049-parse-multiply-by-constant.md
john-h-kastner-aws, line 1: nit: this is #50, not #49

text/0049-parse-multiply-by-constant.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Basic example
+
+`2*1*context.value` should be conceptually converted into `((2*1)*context.value)` before checking if it's a valid multiply-by-constant operation. The checking should pass and the resulting AST should be `(mul (mul 2 1) context.value)`.
+
+## Motivation
+
+Let's first revisit the status quo: The parser separates the operands of a CST representing a (possibly chained) multiplication operation into a non-integer-constant and a list of integer constants, preserving the order of the constants, and then *left-fold* the constant list with the non-constant. For instance, `2*1*content.value` is parsed into `(mul (mul content.value 2) 1)`.
+
+Note that the parser changes the order of operands, which does not alter the evaluation result since Cedar's multiplication operation is associative. However, changing the evaluation order can make error message confusing: For instance, evaluating `2*6*context.value` when `context.value` is `i64::MAX/10` produces an error message like ```integer overflow while attempting to multiply `1844674407370955160` by `6```, contradicting the intuitive left-to-right evaluation order.
```
andrewmwells-amazon, line 24: This isn't true. Consider the policy:
```
permit(principal, action, resource) when
{
    6*0*principal.foo >= 0
};
```
Where `User::"Bob".foo` = 1844674407370955160.
Authorization will deny due to overflow. If we change the policy to:
```
permit(principal, action, resource) when
{
    0*6*principal.foo >= 0
};
```
Authorization will allow.

andrewmwells-amazon (changes requested):

cdisselkoen (approved):
This RFC will make our parsing, and any possible overflow errors during multiplication, more intuitive for users.

One thing not explicitly called out in the RFC is that in order to support ASTs like `(2*1)*context.value`, we will need to change the AST structure, as currently each AST `mul` node contains one expression and one constant, but in that example the two operands to the second mul are `context.value` and `(2*1)` -- an AST GetAttr node and an AST mul node.  I believe this change to the AST can be made, however, and the AST constructors can enforce this RFC's recursive condition, that at least one of the operands to each mul is either a constant integer or a product of constant integers.

text/0049-parse-multiply-by-constant.md
cdisselkoen: @shaobo-he-aws can you change the filename here to be `0050-`?

shaobo-he-aws:
Closed in favor of #57

PR #51, opened 2024-02-09
Changes `IDENT` -> `PATH

aaronjeline:
Small change to the grammar:

The rule for types used to be this:
```
Type      := PRIMTYPE | IDENT | SetType | RecType
```
Where `IDENT` is a pure non-namespaced identifier.
I believe we want `Type` to instead refer to a `Path` which is non-empty list of identifiers separated by `::`

[Rendered](https://github.com/cedar-policy/rfcs/blob/grammar_fixes/text/0024-schema-syntax.md)


john-h-kastner-aws (approved):
Looks good. A reference to a type can be qualified to reference types declared in another namespace. `TypeDecl` still uses Ident since those can't be qualified.

andrewmwells-amazon (approved):

andrewmwells-amazon:
LGTM.

aaronjeline:
> Looks good. A reference to a type can be qualified to reference types declared in another namespace. `TypeDecl` still uses Ident since those can't be qualified.

Right, because that `Ident` is always implicitly in the current namespace

PR #52, opened 2024-02-15
Reserve all namespaces starting with `__cedar`

john-h-kastner-aws:
[Rendered](
https://github.com/cedar-policy/rfcs/blob/jkastner/reserved_ns/text/0052-reserved-namespaces.md)


text/0052-reserved-namespaces.md
```diff
@@ -0,0 +1,102 @@
...
+  ...
+}
+```
+
+and in entity data
+
+```json
+[
+    {
+        "uid": { "__entity": { "type": "__cedar::User", "id": "alice"} },
```
mwhicks1, line 41: Not sure if this was your intention, but you are highlighting that `__entity` is something that starts with two underscores. It is part of entity formatting, not entity data, but it's a little confusing whether you mean that it's ruled out, too. You might point out that `__entity` illustrates the use of `__` as a prefix to avoid conflict with user-defined names, and this is a precedent set by C/C++ (for example): https://devblogs.microsoft.com/oldnewthing/20230109-00/?p=107685

mwhicks1 (approved):
In general I like this change but before making the final decision we should get some data to convince ourselves that this will not affect a lot of customers.

khieta (approved):
Overall lgtm, but I think I prefer the "larger breaking change" alternative. This is more consistent with our current use of the reserved `__entity` attribute.

text/0052-reserved-namespaces.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+In this entity data example, the `__entity` JSON key does not demonstrate the new reserved namespace mechanism.
+It is a place were we already assign special significance to some identifiers starting with `__` in entity data.
+
+## Motivation
+
+In designing [RFC24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) we found that we needed to disambiguate user defined identifiers from built-in primitive and extension types.
+We accomplished this by reserving only the `__cedar` namespace inside of schema,
+but we anticipate that we may need more reserved namespaces in more contexts in the future.
```
cdisselkoen, line 58: I'm not totally convinced by this motivation.  If we have `__cedar` as a reserved namespace, what would cause us to need other `__` namespaces?  If we do need other namespaces, we can use sub-namespaces of `__cedar` like  `__cedar::A`, `__cedar::B`, etc, which are already reserved as part of the `__cedar` namespace.

text/0052-reserved-namespaces.md
```diff
@@ -0,0 +1,108 @@
...
+### A larger breaking change
+
+We could make a larger breaking change by reserving all _identifiers_ starting with `__` instead of just namespaces.
+This would include inner namespace elements and attribute names on entities and records.
+For example, we would additional forbid referencing an entity `namespace::__User` or writing a record literal `{__attr: 1}`.
+We don't currently see an immediate need for this more aggressive change, but making it now provides even more flexibility in future changes at the cost of the breaking change affecting more users.
+
+### A smaller breaking change
+
+We can make a smaller breaking change by reserving only the `__cedar` namespace.
```
cdisselkoen, line 98: I prefer this alternative.  I envision this would be sufficient for any purposes envisioned by this RFC.  Or to rephrase, this RFC hasn't convinced me that we need more than this.

text/0052-reserved-namespaces.md
```diff
@@ -0,0 +1,108 @@
...
+Only identifiers whose path starts with `__cedar` would become invalid identifiers, leaving users free to use an identifier like `__namespace::User`, but not `__cedar::User` or `__cedar::inner::User`.
+This still provides names we can use in future features while only affecting users who specifically use this exact namespace.
+
+### No breaking change
+
+We can avoid a breaking change to Cedar by updating the Grammar to allow for a new identifier that is currently invalid, and then reserve that identifier.
+For example, we could reserve `$cedar` or `cedar$` (other options for characters include `%` or `!`).
+It would would be an error to use this identifier in any way that isn't explicitly permitted by the Cedar spec.
```
cdisselkoen, line 106: I also think this would be a completely reasonable alternative, if we can settle on a suitable identifier.  (There is some concern that `$cedar` looks like a macro invocation or template parameter in other languages.)

cdisselkoen (commented):
Not currently convinced by this RFC's main proposal, and at the moment I think some of the alternatives are better.  See comments inline.

text/0052-reserved-namespaces.md
shaobo-he-aws: I agree with Craig on this alternative. But we should rephrase it as "namespaces starting with `__cedar`".

text/0052-reserved-namespaces.md
shaobo-he-aws: This implementation of this alternative is hard to get right as we need to make exceptions on the grammar level.

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

shaobo-he-aws (approved):

PR #53, opened 2024-02-20
Enumerated Entity Types RFC

mwhicks1:
Proposing an extension to schemas and validation to support enumerations of entity types.

[Rendered](https://github.com/cedar-policy/rfcs/blob/enum-entities/text/0053-enum-entities.md)

### DIFF
--- a/text/0053-enum-entities.md
+++ b/text/0053-enum-entities.md
@@ -1,0 +1,135 @@
+# Enumerated Entity Types
+
+## Timeline
+
+- Started: 2024-02-20
+
+## Summary
+
+Extend schemas to support declared enumerations of entity-typed values, analogous to how schemas can currently be used to enumerate a finite list of `Action`-typed values.
+
+## Basic example
+
+An enumerated entity type is declared as a normal entity type, but includes the keyword `enum` followed by the list of legal entity UIDs. Here is a simple example:
+```
+entity User;
+entity Color enum ["red", "blue", "green"];
+entity Task {
+    owner: User,
+    name: String,
+    status: Color
+};
+action UpdateTask
+    appliesTo { principal: [User], resource: [Task] };
+```
+These data definitions could be used in policies such as 
+```
+permit(
+    principal,
+    action == Action::"UpdateTask",
+    resource)
+when {
+    principal == resource.owner &&
+    resource.status != Color::"red"
+};
+```
+
+## Motivation
+
+Enumerated types are useful when you have a fixed set of possible values, and the only thing you want to do with the values is compare them for equality. While you could effectively treat an entity type as an enumeration now, without declaring it in a schema, you gain some benefits by declaring it:
+
+- The validator can error on uses of illegal enumerated values, e.g., flagging the typo `resource.status != Color::"Red"` in the `when` clause in the basic example. 
+- When using a policy analyzer, it can always generate request and entity store instances where the enumerated entity type has declared-valid values, rather than random UIDs.
+- When using an IDE or web-based policy builder, the enumeration can inform auto-completion suggestions. For the basic example above, in an auto-completing IDE writing `resource.status != ` ... would pop up the three options.
+
+## Detailed design
+
+An enumerated entity `Foo` is declared by writing
+```
+entity Foo enum [ ... ];
+```
+where `...` is the list of allowed values. These values can be strings or identifiers.
+
+You can use an enumerated entity type anywhere you can use a normal entity type. Since an enumerated entity cannot have attributes, nor can it have ancestors in the entity hierarchy, all you can do with it is test it for equality (e.g., with `==` or `contains`).
+
+The policy validator confirms that any enumerated entity literal in a policy is valid. The request validator does likewise. The entity store validator confirms that enumerated entities _do not_ appear in the store, or if they do, they have no attributes or ancestors. It also confirms that the declared enumerated entity type has no invalid values, and references to entities of the enumerated type are valid. The schema-based entity parser likewise confirms that parsed-in enumerated entity values are valid.
+
+As another example, consider the following.
+```
+entity Application enum [ "TinyTodo" ];
+entity User in [ Application ];
+action CreateList
+    appliesTo { principal: [User], resource: [Application] };
+```
+This is a generalization of our TinyTodo example from RFC 24, where we can refine the definition of `Application` to indicate that it has a single value, `Application::"TinyTodo"`. This allows the validator to catch typos in policies, such as the following.
+```
+permit(
+    principal,
+    action in [Action::"CreateList"],
+    resource == Application::"TinyTODO"
+);
+```
+Likewise the request validator would flag a request with `Application::"TinyTODO"` as its resource, and it would flag the passed-in entity store if it contained such an illegal value.
+
+## Drawbacks
+
+One reason not to do this is that it's not particularly full featured---you cannot do anything useful with an enumerated entity value in a policy other than compare it for equality. We consider more full-featured extensions in the alternatives below, but these have drawbacks of their own. The functionality could be easily extended later, depending on how things play out.
+
+## Alternatives
+
+### Alternative A: Enumerated primitive values
+
+We previously proposed, in [RFC 13](https://github.com/cedar-policy/rfcs/blob/enums/text/0013-schema-enums.md), declaring a finite set of primitive values (strings, numbers, etc.) as an enumerated type. The [killer objection to the RFC](https://github.com/cedar-policy/rfcs/pull/13#issuecomment-1786170514) is that it introduces subtyping (you want to use the enumerated type of string as both the enumeration and the string, e.g., with `like`), which is a significant complication for policy analysis. The present proposal is not problematic for analysis as enumerated entities can be encoded like any entity, but with added ground constraints limiting what form their UIDs can take on.
+
+### Alternative B: Enumeration as a distinct concept
+
+Rather than specify a particular entity type as an enumeration, we could define a new concept of enumeration as another kind of primitive type. Here is a notional schema:
+```
+enum type Color = red | blue | green;
+entity Task {
+  name: String,
+  status: Color
+};
+```
+Here is a notional policy:
+```
+permit(
+    principal,
+    action == Action::"UpdateTask",
+    resource)
+when {
+    resource.status != Color.red
+};
+```
+This syntax is similar to what's provided for Java `enum`s. 
+
+The benefit of this approach is that it may feel a little more natural than representing a concept, like a color, as a set of legal entity values. It would also be easy to encode this approach in a policy analysis.
+
+The main drawback of this approach is that it introduces a new primitive type to the language that does not make much sense without schemas. Recall that for Cedar, validation with a schema is _optional_. We'd allow users to include any random enumerated identifiers (like Scheme-style _symbols_) in policies which, without validation, would fail equality checks at run-time.
+
+The proposed approach that lists particular entity values as an enumeration makes no change to the language, leveraging the existing entity type concept. So policies without schemas still make sense. With the added schema, there is additional benefit when validating and constructing reasonable policies.
+
+### Alternative C: Enumerated entities with hierarchy
+
+Enumerated entities as proposed are limited in their functionality and specificity. We could extend them. For example:
+```
+entity Application enum [ "TinyTodo" ];
+entity RequestEntity enum [ "principal", "resource" ] in [ Application ];
+entity User in [ Application ];
+action CreateList
+    appliesTo { principal: [User], resource: [Application] };
+action GetLists
+    appliesTo { principal: [User, RequestEntity::"principal"],
+                resource: [Application]};
+```
+This differs from some earlier examples in the following ways:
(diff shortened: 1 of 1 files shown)


andrewmwells-amazon (approved):
I'd be happy with the proposal or alternative C, but starting with the simple version seems reasonable to me.

max2me:
This functionality will help eliminate class of errors that could otherwise occur during development.

text/0053-enum-entities.md
```diff
@@ -0,0 +1,146 @@
...
+- When using a policy analyzer, it can always generate request and entity store instances where the enumerated entity type has declared-valid values, rather than random UIDs.
+- When using an IDE or web-based policy builder, the enumeration can inform auto-completion suggestions. For the basic example above, in an auto-completing IDE writing `resource.status != ` ... would pop up the three options.
+
+## Detailed design
+
+An enumerated entity `Foo` is declared by writing
+```
+entity Foo enum [ … ];
+```
+where `[` … `]` is the list of allowed values. These values can be strings or identifiers.
```
cdisselkoen, line 51: I don't think I like throwing in the "or identifiers" on the end.  Also, none of the existing examples in this writeup take advantage of the "or identifiers" part: in all of the examples in the RFC, the values are strings, not identifiers. 

Here's an example of what the "or identifiers" allows, but I think we should disallow:
```
entity Color enum [red, blue, green];
```
vs how it appears in all the examples currently:
```
entity Color enum ["red", "blue", "green"];
```
This latter form seems more intuitive, since you'll later have to reference the values as `Color::"red"`, `Color::"blue"`, etc.

cdisselkoen (approved):
The main proposal seems like a sensible MVP (minimum viable product).  Happy to see this added to the language.

max2me (approved):
Ship it!

text/0053-enum-entities.md
mwhicks1: That's fair. We can drop the ID part and potentially add it later if there's interest in that.

mwhicks1:
I thought of another thing that might be useful to add: A built-in attribute `name` that contains the name of the enumerated type as a `String`. It will always match the UID. So for the example in the proposal, you could write
```
permit(
    principal,
    action == Action::"UpdateTask",
    resource)
when {
    principal == resource.owner &&
    resource.status.name == "red"
};
```
instead of
```
permit(
    principal,
    action == Action::"UpdateTask",
    resource)
when {
    principal == resource.owner &&
    resource.status != Color::"red"
};
```
Importantly, this would let you compute on enumerated type names, e.g., with `like`, which is something you can do in Java's `enum`.

mwhicks1:
> A built-in attribute name that contains the name of the enumerated type as a String

After some discussion off-line, it seems like this functionality could be useful for entities generally. It could be implemented as a special attribute (say, `$id`) with which you can access an entity's UID. For some entities this attribute won't be particularly useful (e.g., when the UID is a random immutable string acting as a unique identifier), but for others it will, including `enum` entities and action entities. I'll write up a separate RFC to consider it.

mwhicks1:
Added [RFC 54](https://github.com/cedar-policy/rfcs/pull/54) for `$id` pseudo-attribute.

andrewmwells-amazon:
Re-reading this, I see we use both "red" and "TinyTodo" in our examples. While I don't think the language should enforce a capitalization system, I think we should decide on, encourage and consistently use one in our examples. Based on C++ conventions, I'd vote for `ALLCAPS`, but I'm less opinionated on what we choose than that we should choose something.

text/0053-enum-entities.md
```diff
@@ -0,0 +1,146 @@
...
+2. Enumerated entities can appear as _singleton types_, e.g., as `RequestEntity::"principal"` in the definition of action `GetLists`.
+
+Both of these extensions are similar to what's available for `Action`s right now, but generalized to arbitrary entity types. You could also imagine enumerated entity types having attributes, as has been anticipated for `Action`s.
+
+One drawback of this Alternative is that it creates added complication for both the validator (e.g., to handle singleton types) and the analyzer (e.g., to deal with the more general hierarchy constraints).
+
+Another drawback is that it adds complication to the entity store: Once you add hierarchy constraints and attributes, you need to create actual entities to include with policy requests (or extract them from the schema at request-time). The RFC as proposed does not require that.
+
+The good news is that this Alternative is a strict generalization of the RFC as proposed, which means if we agree to the current proposal we can later upgrade to some or all of this Alternative without incurring a breaking change.
```
john-h-kastner-aws, line 146: I wouldn't be surprised if we re-thought Actions in schema while implementing this extension, so we may end up _wanting_ to make a breaking change, but that could come up regardless of when we implement this part.

john-h-kastner-aws (approved):
Main proposal looks great. 

The extensions in alternative C seem reasonable, but I don't want to accept those immediately.

text/0053-enum-entities.md
```diff
@@ -0,0 +1,146 @@
...
+## Summary
+
+Extend schemas to support declared enumerations of entity-typed values, analogous to how schemas can currently be used to enumerate a finite list of `Action`-typed values.
+
+## Basic example
+
+An enumerated entity type is declared as a normal entity type, but includes the keyword `enum` followed by the list of legal entity UIDs. Here is a simple example:
+```
+entity User;
+entity Color enum ["red", "blue", "green"];
```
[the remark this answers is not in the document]
mwhicks1: To me it felt natural to declare the entity type, first, and then to refine what UIDs it can take on, second. Migrating from TinyTodo without enums, we had
```
entity Application;
entity User in [Application];
```
and then within enums added it becomes
```
entity Application enum ["TinyTodo"];
entity User in [Application];
```
Thus users can become comfortable always writing `entity NAME` and then following it up, and not sometimes writing `enum` before the name.

Also, it looks a little weird to me with the order reversed; I want to see punctuation between the entity name and the enumerations (beyond `[`). 
```
entity enum Application ["TinyTodo"];
entity enum Color ["red", "blue", "green"];
```
You could add `::` there as in `entity enum Application :: ["TinyTodo"]` but then seems like it might be confusing too.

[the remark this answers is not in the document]
khieta: I also prefer `entity enum Color`. I feel like this helps distinguish enums from other types of entities, which can have attributes & ancestors in the hierarchy.

khieta, line 16: Attaching @andrewmwells-amazon's comment to the text for easier discussion threading.
> Re-reading this, I see we use both "red" and "TinyTodo" in our examples. While I don't think the language should enforce a capitalization system, I think we should decide on, encourage and consistently use one in our examples. Based on C++ conventions, I'd vote for ALLCAPS, but I'm less opinionated on what we choose than that we should choose something.

khieta: I also would prefer consistency here. To match "TinyTodo", maybe switch the example to use "Red", "Blue", "Green"?

khieta (approved):
I really like this proposal! I also think it will be a big benefit to the change proposed in RFC #55.

text/0053-enum-entities.md
emina: Adding ancestors and attributes is okay for analysis as long as they are all fully known and specified in the schema.  What would be difficult to handle is if we wanted to have some `enum` instance (or, for that matter, some action) with unknown or partially known ancestors or attribute.   

I don’t see a good use case for partially specified actions and enums but it would be good to clarify in the RFC that everything about these special entities (that are fixed in the schema) must be fully specified.

emina:
> I thought of another thing that might be useful to add: A built-in attribute `name` that contains the name of the enumerated type as a `String`. It will always match the UID. So for the example in the proposal, you could write
> 
> ```
> permit(
>     principal,
>     action == Action::"UpdateTask",
>     resource)
> when {
>     principal == resource.owner &&
>     resource.status.name == "red"
> };
> ```
> 
> instead of
> 
> ```
> permit(
>     principal,
>     action == Action::"UpdateTask",
>     resource)
> when {
>     principal == resource.owner &&
>     resource.status != Color::"red"
> };
> ```
> 
> Importantly, this would let you compute on enumerated type names, e.g., with `like`, which is something you can do in Java's `enum`.

I vote against adding this because it will be expensive to analyze for large enums.  Effectively, any attempt to dereference the “name” of an unknown (symbolic) enum entity will expand into a chained if-then-else expression that compares the unknown value to every instance of the type. 

We can reconsider if there is a strong demand for it.  But for the MVP, I’d go with the proposal as written—without any ancestors or attributes.

text/0053-enum-entities.md
emina: To be clear, this is okay in the sense that it doesn’t affect soundness, completeness, or decidability.  But it will have a negative effect on performance for large enums (see [this comment](https://github.com/cedar-policy/rfcs/pull/53#issuecomment-1974227827)).

text/0053-enum-entities.md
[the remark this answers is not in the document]
aaronjeline: Also prefer `entity enum`

aaronjeline (approved):
Agree w/ most of what's said in the thread, esp that we should go with the minimum option (no magic `id` attr and no ancestors) and expand later if need be.

Making the note that the entityuids essentially already are Schema style symbols, just with an extra de reference operator. I don't like duplicating the functionality to add symbols.

aaronjeline:
Talked about the some offline, any reason to not disallow empty enums? We can always add it back later if we find a use for empty types, but for now seems like possible to cause confusion.

text/0053-enum-entities.md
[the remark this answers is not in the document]
mwhicks1: I'm the odd person out, but I strongly prefer `entity Application enum [...]` over `entity enum Application [...]` for uniformity reasons.

When we write `entity` we are declaring a type, first and foremost. Then we refine that declaration with an enumeration of elements of that type. Putting the `entity enum ...` makes it seem like `enum` is the type.

Consider OCaml: You write `type id = string` or `type list = Cons of int * list | Nil`, i.e., you declare the type name first and then say what's in it. You don't prefix the type name with information about what kind of type it is. 

I'll add that other refinements/defs go after the declaration of the entity type, now. `entity Application enum ["TinyTodo"]` is similar to `entity List in [Application]` -- we are postfixing the type declaration with information about it. In the future we are leaving open the possibility of combining these things, e.g., `entity Application in [Application] enum ["TinyTodo"]` (or swapping their order) works better in terms of phrases of declaration.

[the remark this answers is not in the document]
cdisselkoen: This incidentally raises another good point, that `enum` is a valid entity type, so `entity enum in [Group]` and the like are valid today -- potentially a parsing problem (or at least confusion) for the `entity enum Application` option

emina:
> Talked about the some offline, any reason to not disallow empty enums? We can always add it back later if we find a use for empty types, but for now seems like possible to cause confusion.

Coming back to this comment, I realized that for the analysis to faithfully capture the semantics of Cedar, these types _must_ be inhabited, or we need to add extra code to handle the case when they are not.  This extra code would amount to saying that uninhabited (empty) types can’t be used with any operator other than `is`.  But `expr is Foo`, where `Foo` is uninhabited is always “false”, so it’s unclear why this would be a useful thing to write (as opposed to just writing `false`).   

 @mwhicks1 Can we amend the RFC to explicitly forbid empty enum types?

emina (approved):

nakedible-p:
Consider RFC48 as well – enum values may need documentation attributes, so you can specify what each value means and what it is used for.

hakanson:
Clarifying question about the validation of entities passed in to the PARC request when it contains a value not in the enumeration.

> The policy validator confirms that any enumerated entity literal in a policy is valid. The request validator does likewise.

Given this "favorite color" schema

```
entity Color enum ["Red", "Blue", "Green"];
entity User {
    favorite: Color
};
action MatchColor
    appliesTo { principal: [User], resource: [User] };
```

and these Cedar policies

```cedar
// ALLOW Users with matching favorite colors
permit (
    principal,
    action == Action::"MatchColor",
    resource
) when {
    principal.favorite == resource.favorite
};

// Blue is always ALLOWed :) 
permit (principal, action, resource)
when {
    principal.favorite == Color::"Blue"
};
```

Does validation happen when I make these PARC requests?  If not, would the first get an ALLOW?

```yaml
P: User::"Gandalf the Grey"
A: Action::"MatchColor"
R: User::"Gandalf the Grey"
C: {}

P: User::"Gandalf the Grey"
A: Action::"MatchColor"
R: User::"Gandalf the White"
C: {}
```

and these entities

```json
[
  {
      "uid": { "type": "User", "id": "Gandalf the Grey" },
      "attrs": {
          "favorite": { "type": "Color", "id": "Grey" }
      },
      "parents": []
  },
  {
      "uid": { "type": "User", "id": "Gandalf the White" },
      "attrs": {
          "favorite": { "type": "Color", "id": "White" }
      },
      "parents": []
  }
]
```

mwhicks1:
> Clarifying question about the validation

We have validators for Cedar policies and optionally for Cedar requests and entities (with schema-based parsing). In your example, schema-based parsing of the entities file would flag the entities as non-compliant with the schema.

mwhicks1:
> Consider RFC48 as well – enum values may need documentation attributes, so you can specify what each value means and what it is used for.

Yes, any annotation support for `entity` declarations in schemas would also apply to `entity` declarations that are `enum`s.

PR #54, opened 2024-02-23
eid as attribute

mwhicks1:
Proposes to make an entity reference's EID a pseudo-attribute `$id`.

[Rendered](https://github.com/cedar-policy/rfcs/blob/eid-attribute/text/0054-eid-attribute.md)

### DIFF
--- a/text/0054-eid-attribute.md
+++ b/text/0054-eid-attribute.md
@@ -1,0 +1,61 @@
+# Include entity ID as a special attribute
+
+## Timeline
+
+- Started: 2024-02-22
+
+## Summary
+
+Add a pseudo-attribute `$id` that when applied to an entity reference returns its entity identifier.
+
+## Basic example
+
+`Action::"readFile".$id == "readFile"`
+
+## Motivation
+
+When an entity's EID is meaningful (and not a random, unique identifier), it could be useful for logic expressed in policies. For example, in RFC 53 we propose enumerated entity types, such as the following:
+```
+entity Course enum ["CMSC330", "CMSC351", "MATH241", "ENGL101"];
+```
+This schema definition indicates four `Course`-typed entities, i.e., `Course::"CMSC330"`, `Course::"CMSC351"`, `Course::"MATH241"`, and `Course::"ENGL101"`. You might like to write a policy that refers to the EIDs of these values, e.g.,
+```
+permit(
+    principal in Group::"students",
+    action == Action::"addCourse",
+    resource is Course
+) when {
+    !(resource.$id like "CMSC*") && context.today < context.addDeadline ||
+    context.today < (context.addDeadline-1)
+}
+```
+This policy states that a student can add a non computer-science course any time up to the add-deadline, but computer science courses must be added a day before that.
+
+You could similarly want to examine `Action` entities' EIDs:
+```
+permit(principal, action, resource)
+when {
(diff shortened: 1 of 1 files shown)


hakanson:
Is there a potential drawback / anti-pattern where Cedar policies will use `$id` and String values where before they used Entities?  For example, this policy has `owner` as an Entity

```
permit (principal is PhotoApp::User, action, resource is PhotoApp::Photo)
when { resource.owner == principal };
```

example slice that would be passed in using `__entity`

```
  {
    "uid": { "type": "PhotoApp::Photo", "id": "JaneDoe.jpg" },
    "parents": [],
    "attrs": {
      "owner": {
        "__entity": { "type": "PhotoApp::User", "id": "JaneDoe" }
      }
    }
  }
```

This RFC would enable policies where `owner` is an String matching the id

```
permit (principal is PhotoApp::User, action, resource is PhotoApp::Photo)
when { resource.owner == principal.$id };
```

```
  {
    "uid": { "type": "PhotoApp::Photo", "id": "JaneDoe.jpg" },
    "parents": [],
    "attrs": {
      "owner": "JaneDoe"
    }
  }
```

The `owner` attribute as String may seem more obvious to early policy authors, but using an Entity in the schema enables future conditions to use other attributes of that Entity beyond the id.  I'm not apposed to the RFC, just thinking about future best practices documentation, as this might affect backwards compatible Cedar schema authoring.

mwhicks1:
> Cedar policies will use $id and String values where before they used Entities?

Yes, I could imagine this happening. Put another way, we are setting `resource.owner` to be an entity's EID and not the entity itself, which means that different-typed "entities" (one of which is stored as just its string EID) could be deemed equal, e.g., `User::"a9edd19b-46f3-486b-887d-4c378aced880".$id == Admin::"a9edd19b-46f3-486b-887d-4c378aced880".$id` while `User::"a9edd19b-46f3-486b-887d-4c378aced880" != Admin::"a9edd19b-46f3-486b-887d-4c378aced880"`.

I'm not sure how tempting this would be to do, though, if users follow our advice of using unique IDs for EIDs of entities like users, resources, etc. The EID is going to be non-sensical (like `a9edd19b-46f3-486b-887d-4c378aced880` above), so you won't be tempted to store it in an attribute directly. The `$id` attribute is useful only when the EID is meaningful, which should be the case for `enum`s and `Action` entities.

andrewmwells-amazon:
I feel like this would encourage bad practices. For Actions, I think policies should check `action in MyActionGroup` instead of `action.$id like "prefix*"`, so in my understanding this RFC would only be useful for enumerated entities. 

If enough people ask for a feature like this, I'd rather implement alternative C of https://github.com/cedar-policy/rfcs/pull/53 than add this.

khieta:
Agree with @andrewmwells-amazon's comment above. Also see @D-McAdams's post on why not to allow wildcard matching for Entity Ids: <https://cedarland.blog/design/why-no-entity-wildcards/content.html>.

emina:
> Agree with @andrewmwells-amazon's comment above. Also see @D-McAdams's post on why not to allow wildcard matching for Entity Ids: https://cedarland.blog/design/why-no-entity-wildcards/content.html.

Agree with the above (as well as Andrew’s and Kevin’s points) and would vote against adding this feature.

PR #55, opened 2024-02-27
Explicit Unspecified Entities

khieta:
[Rendered](https://github.com/cedar-policy/rfcs/blob/remove-unspecified/text/0055-remove-unspecified.md)

### DIFF
--- a/text/0055-remove-unspecified.md
+++ b/text/0055-remove-unspecified.md
@@ -61,11 +61,8 @@ We expect that this concept will only become more confusing once we stabilize th
 Unspecified entities also lead to two sharp corners in our public API:
 
 1. Omitting a `appliesTo` field in the schema (or setting it to null) means that an action applies only to unspecified entities, while using an empty list means that it applies to no entities. There is no way to say that an action applies to both unspecified entities and other types of entities. For more discussion, see [RFC35](https://github.com/cedar-policy/rfcs/pull/35). (Note: RFC35 has been closed in favor of the current RFC.)
-2. It is unclear how we should create `Request`s that use both unspecified and unknown entities. This has not been an issue so far since partial evaluation and its APIs are experimental, but we are looking to stabilize them soon. Currently the best (although not ideal) proposal is to use a builder pattern for `Request`s (e.g., `request.principal(Some(User::"alice"))` sets the principal field of a request). To set a field to be “unspecified” users should pass in `None` instead of `Some(_)`; to set a field to be unknown users should not call the field constructor at all.
-
-### Unspecified vs. unknown
-
-Unspecified and unknown entities are both pre-defined "dummy" entities that act like any other entity, as long as they are unused or only used in a trivial way (e.g., `principal == principal`).
+2. It is unclear how we should create `Request`s that use both unspecified and unknown entities. This has not been an issue so far since partial evaluation and its APIs are experimental, but we are looking to stabilize them soon. Currently the best (although not ideal) proposal is to use a builder pattern for `Request`s (e.g., `request.principal(Some(User::"alice"))` sets the principal field of a request). To set a field to be “unspecified” users could pass in `None` instead of `Some(_)`; to set a field to be unknown users would not call the field constructor at all.  
+_NB:_ Unspecified and unknown entities are both pre-defined "dummy" entities that act like any other entity, as long as they are unused or only used in a trivial way (e.g., `principal == principal`).
 If a policy uses an _unknown entity_ in a non-trivial way, then evaluation will produce a residual.
 If a policy uses an _unspecified entity_ in a non-trivial way, then evaluation will produce an error.
 
--- a/text/0055-remove-unspecified.md
+++ b/text/0055-remove-unspecified.md
@@ -15,7 +15,8 @@ Cedar currently supports _unspecified entities_, which are entities of a special
 
 The current implementation of unspecified entities results in several corner cases in our code (especially the validator), which have been difficult to maintain (see discussion on [this recent PR](https://github.com/cedar-policy/cedar/pull/603)). The concept of unspecified entities is (in our opinion) confusing for users, and will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities.
 
-This RFC proposes to drop Cedar support for unspecified entities, instead recommending that users create their own application-specific entity types that act like unspecified entities.
+This RFC proposes to change the terminology surrounding unspecified entities, instead calling them "default" entities.
+It also proposes to give an explicit name to the unspecified entity type (`__cedar::Default`) and to require this name to be used in requests and schemas that use this type.
 
 ## Basic example
 
@@ -29,12 +30,12 @@ permit(principal is User, action == Action::"createFile", resource);
 
 When making an authorization request for this action, the principal should be a `User`, and the context can store information about the file to create, but it is unclear what the resource should be.
 In fact, given the policy above, the resource _doesn't matter_.
-In this case, a Cedar user may want to make the authorization request with a “dummy” resource; this is what Cedar’s unspecified entities are for.
+In this case, a Cedar user may want to make the authorization request with a “dummy” resource; as of Cedar version 3.1, this is what unspecified entities are for.
 
-As of Cedar v3.1, the only way to “construct” an unspecified entity is by passing `None` as a component when building a [`Request`](https://docs.rs/cedar-policy/3.0.1/cedar_policy/struct.Request.html).
+The only way to “construct” an unspecified entity is by passing `None` as a component when building a [`Request`](https://docs.rs/cedar-policy/3.0.1/cedar_policy/struct.Request.html).
 So a request for this action might look like \<`Some(User::"alice")`, `Some(Action::"createFile")`, `None`, ...\> (using pseudo-syntax).
 
-In the schema, omitting the relevant `appliesTo` field says that an action applies to an unspecified entity.
+In the schema, the only way to say that an action applies to an unspecified entity is by omitting the relevant `appliesTo` field.
 For example, the following schemas specify that the "createFile" action applies (only) to an unspecified resource.
 
 Cedar schema syntax ([RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md)):
@@ -54,35 +55,8 @@ Cedar schema JSON syntax:
 }
 ```
 
-Now consider a "createAccount" action.
-It's unclear what principal _or_ resource should be used with this action.
-The approach above can be extended to this case too: make the action apply to an unspecified principal _and_ resource.
-But now "unspecified" entities are being used to represent multiple distinct concepts: a file store, (unidentified) users creating an account, and a system that manages accounts.
-
-This RFC proposes that users should instead explicitly create these "dummy" entities in an application-specific manner.
-For this example, the user should create entities of three types: `FileSystem`, `UnauthenticatedUser`, and `AccountManager`.
-The first will act as the principal for a "createFile" request, and the second and third will act as the principal and resource for a "createAccount" request.
-
-This is exactly the approach we used in the [TinyTodo application](https://github.com/cedar-policy/cedar-examples/tree/release/3.0.x/tinytodo): we made a special `Application::“TinyTodo”` entity to act as the resource for the “CreateList” and “GetLists” actions.
-This proposal is also in line with [our previous suggestion](https://cedar-policy.slack.com/archives/C0547KH7R19/p1706656288284189) to use an application-specific `Unauthenticated` type to represent "unauthenticated" users, rather than using unspecified entities.
-
-## Motivation
-
-Unspecified entities are a confusing concept, as evidenced by messy special casing in our own code and multiple RFCs to refine the behavior ([RFC35](https://github.com/cedar-policy/rfcs/pull/35), [RFC47](https://github.com/cedar-policy/rfcs/pull/47)).
-We expect that this concept will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities.
-Unspecified and unknown entities are both pre-defined "dummy" entities that act like any other entity, as long as they are unused or only used in a trivial way (e.g., `principal == principal`).
-If a policy uses an _unknown entity_ in a non-trivial way, then evaluation will produce a residual.
-If a policy uses an _unspecified entity_ in a non-trivial way, then evaluation will produce an error.
-
-Unspecified entities also lead to two sharp corners in our public API:
-
-1. Omitting a `appliesTo` field in the schema (or setting it to null) means that an action applies only to unspecified entities, while using an empty list means that it applies to no entities. There is no way to say that an action applies to both unspecified entities and other types of entities. For more discussion, see [RFC35](https://github.com/cedar-policy/rfcs/pull/35). (Note: RFC35 will be closed in favor of the current RFC, if accepted.)
-2. It is unclear how we should create `Request`s that use both unspecified and unknown entities. This has not been an issue so far since partial evaluation and its APIs are experimental, but we are looking to stabilize them soon. Currently the best (although not ideal) proposal is to use a builder pattern for `Request`s (e.g., `request.principal(Some(User::"alice"))` sets the principal field of a request). To set a field to be “unspecified” users should pass in `None` instead of `Some(_)`; to set a field to be unknown users should not call the field constructor at all.
-
-## Detailed design
-
-This RFC proposes to drop unspecified entities from the Cedar language.
-This means that the type of the `Request` constructor will change from:
+This RFC proposes to instead refer to these special entities as **default entities** since you should use them when you want _some default value_ to pass into an authorization request, but you don't care what the exact value is because it doesn't impact authorization.
+This RFC also proposes to introduce a new special type (`__cedar::Default`) that allows users to refer to default entities _explicitly_ instead of _implicitly_, and change the type of the `Request` constructor from:
 
 ```rust
 pub fn new(
@@ -98,6 +72,130 @@ pub fn new(
 to:
 
 ```rust
+pub enum RequestEntity {
+    Default,
+    Unknown,
+    EntityUid(EntityUid)
+}
+
+pub fn new(
+        principal: RequestEntity,
+        action: RequestEntity,
+        resource: RequestEntity,
+        context: Context,
+        schema: Option<&Schema>,
+    ) -> Result<Self, RequestValidationError>
+...
+```
+
+(Note: the `Unknown` variant may be hidden behind the "partial-eval" experimental flag, depending on when this RFC is implemented.)
+
+Under this proposal, the request above would become \<`EntityUid(User::"alice")`, `EntityUid(Action::"createFile")`, `Default`, ...\>.
+
+Finally, schemas will need to explicitly say that an action applies to the default type, rather than omitting an `appliesTo` field.
+For example, the schemas above would need to be modified as follows.
+
+Cedar schema syntax (RFC 24):
+
+```
+action createFile
+  appliesTo { principal: [User], resource: [__cedar::Default] };
+```
+
+Cedar schema JSON syntax:
+
+```json
+"createFile": {
+    "appliesTo": {
+        "principal": ["User"],
+        "resource": ["__cedar::Default"]
+    }
+}
+```
+
+Users will not be allowed to include entities of type `__cedar::Default` in the store, which ensures that these entities will never have attributes or ancestors/descendants, which are properties we will rely on during validation.
+Users will not be allowed to reference the type `__cedar::Default` in their policies (so neither `principal == __cedar::Default::"principal"` nor `principal is __cedar::Default` are allowed).
+Users do not need to add `__cedar::Default` to their schemas (and it will be an error if they do) because `__cedar::Default` is defined automatically.
+
+## Motivation
+
+Unspecified entities are a confusing concept, as evidenced by messy special casing in our own code and multiple RFCs to refine their behavior ([RFC35](https://github.com/cedar-policy/rfcs/pull/35), [RFC47](https://github.com/cedar-policy/rfcs/pull/47)).
+We expect that this concept will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities, despite sharing similar names.
+Unspecified and unknown entities are both pre-defined "dummy" entities that act like any other entity, as long as they are unused or only used in a trivial way (e.g., `principal == principal`).
+If a policy uses an _unknown entity_ in a non-trivial way, then evaluation will produce a residual.
+If a policy uses an _unspecified entity_ in a non-trivial way, then evaluation will produce an error.
+
+Unspecified entities also lead to two sharp corners in our public API:
+
+1. Omitting a `appliesTo` field in the schema (or setting it to null) means that an action applies only to unspecified entities, while using an empty list means that it applies to no entities. There is no way to say that an action applies to both unspecified entities and other types of entities. For more discussion, see [RFC35](https://github.com/cedar-policy/rfcs/pull/35). (Note: RFC35 has been closed in favor of the current RFC.)
+2. It is unclear how we should create `Request`s that use both unspecified and unknown entities. This has not been an issue so far since partial evaluation and its APIs are experimental, but we are looking to stabilize them soon. Currently the best (although not ideal) proposal is to use a builder pattern for `Request`s (e.g., `request.principal(Some(User::"alice"))` sets the principal field of a request). To set a field to be “unspecified” users should pass in `None` instead of `Some(_)`; to set a field to be unknown users should not call the field constructor at all.
+
+## Detailed design
+
+### Naming specifics
+
+The names `RequestEntity` and `__cedar::Default` are both subject to change.
+
+For the latter, we have agreed that we want to use the `__cedar` prefix (reserved as part of [cedar#557](https://github.com/cedar-policy/cedar/pull/557) in v3.1), but the exact type name is up for debate.
+Options we've considered include:
+
+- `None`
+- `Null`
+- `Ghost`
+- `Dummy`
+- `Empty`
+- `Irrelevant`
+- `Generic`
+
+Here are some options we’ve ruled out:
+
+- `Any`: possibility to confuse the scope condition `principal` with `principal is __cedar::Any`. The first applies to any principal, while the latter applies to only the “unspecified” entity.
+- `Arbitrary`: same issue as “any”
+- `Unspecified`: potential for confusion with partial evaluation “unknown”
+- `Mock`: implies that this entity type should be used for debugging/testing
+
+### `Default` vs. `EntityUid(__cedar::Default::"eid")`
+
+Under the current proposal, there are two ways to pass an unspecified entity into a `Request`: `RequestEntity::Default` and `RequestEntity::EntityUid(__cedar::Default::"eid")`.
+Both are equivalent, although the latter allows the user to assign an particular name ("eid") to the entity, which could be helpful when debugging error messages.
+When using the `Default` variant, Cedar will assign the entity id "principal", "action", or "resource" depending on where the entity is used in the request.
+
+### Upgrade strategy
+
+This RFC proposes a **breaking change** for both the API and the language.
+It breaks the API because it requires modifying the `Request` constructor, and it breaks the language because it changes the schema format.
+Fortunately, this proposal allows for a straightforward period of deprecation.
+
+To start, we will leave the current APIs and schema format as-is, while also adding the new `__cedar::Default` type and support for using it in schemas and requests.
+Passing `None` into the `Request` constructor will become equivalent to passing in `Some(__cedar::Default::"eid")`.
+Omitting an `appliesTo` field in the schema will become equivalent to having `[__cedar::Default]`, but we will emit a warning to recommend updating to the new syntax.
+
+Then, in a subsequent major version, we can implement the breaking changes described above.
+
+### Documentation
+
+As pointed out in reviews of this RFC, a key reason why unspecified entities are confusing is that we have almost no documentation for them.
+Thus, an important part of this RFC is developing useful documentation.
+We expect this to include a blog post with examples of applications where you may want an unspecified entity, and recommendations for best practices.
+
+## Drawbacks
+
+Breaking changes are painful for users.
+We need to make sure that we provide enough value in the major version with this change for the upgrade to be worth it.
+
+## Alternatives
+
+### Alternative A - Get rid of the unspecified entity type
+
+The current proposal can be viewed as a re-framing of the current unspecified type.
+A more radical option (as originally proposed by this RFC) is to completely drop Cedar support for unspecified entities, instead recommending that users create their own application-specific entity types that act like unspecified entities.
+
+This is (arguably) the approach we used in the [TinyTodo application](https://github.com/cedar-policy/cedar-examples/tree/release/3.0.x/tinytodo): we made a special `Application::“TinyTodo”` entity to act as the resource for the “CreateList” and “GetLists” actions.
+This alternative is also in line with [our previous suggestion](https://cedar-policy.slack.com/archives/C0547KH7R19/p1706656288284189) to use an application-specific `Unauthenticated` type to represent "unauthenticated" users, rather than using unspecified entities.
+
+Under this alternative, the type of the `Request` constructor would become:
+
+```rust
 pub fn new(
         principal: EntityUid,
         action: EntityUid,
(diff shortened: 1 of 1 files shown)


andrewmwells-amazon (approved):
LGTM

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,194 @@
...
+The current implementation of unspecified entities results in several corner cases in our code (especially the validator), which have been difficult to maintain (see discussion on [this recent PR](https://github.com/cedar-policy/cedar/pull/603)). The concept of unspecified entities is (in our opinion) confusing for users, and will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities.
+
+This RFC proposes to drop Cedar support for unspecified entities, instead recommending that users create their own application-specific entity types that act like unspecified entities.
+
+## Basic example
+
+Cedar uses a Principal, Action, Resource, Context \<P,A,R,C\> model for requests.
+This model can feel awkward in cases where there is not a clear principal and/or resource for an action.
+For example, consider a "getFile" action.
+The principal here should be a `User`, and the context can store information about the file to get, but it is unclear what the resource should be.
```
cdisselkoen, line 25: I'm not following -- wouldn't the resource be the file in this case?

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,194 @@
...
+action getFile
+  appliesTo { principal: [User] };
+```
+
+Cedar schema JSON syntax:
+
+```json
+"getFile": {
+    "appliesTo": {
+        "principal": [User]
```
cdisselkoen, line 47:
```suggestion
        "principal": ["User"]
```

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+Now consider a "createAccount" action.
+It's unclear what principal _or_ resource should be used with this action.
+The approach above can be extended to this case too: make the action apply to an unspecified principal _and_ resource.
+But now "unspecified" entities are being used to represent multiple distinct concepts: a file store, (unidentified) users creating an account, and a system that manages accounts.
+
+This RFC proposes that users should instead explicitly create these "dummy" entities in an application-specific manner.
+For this example, the user should create entities of three types: `FileSystem`, `UnauthenticatedUser`, and `AccountManager`.
+The first will act as the principal for a "getFile" request, and the second and third will act as the principal and resource for a "createAccount" request.
```
cdisselkoen, line 59: as the _resource_ for a "getFile" request?

Based on the discussion it seems like the action is more of create-file than get-file? or maybe it's more like list-files?

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,194 @@
...
+
+Unspecified entities also lead to two sharp corners in our public API:
+
+1. Omitting a `appliesTo` field in the schema (or setting it to null) means that an action applies only to unspecified entities, while using an empty list means that it applies to no entities. There is no way to say that an action applies to both unspecified entities and other types of entities. For more discussion, see [RFC35](https://github.com/cedar-policy/rfcs/pull/35). (Note: RFC35 will be closed in favor of the current RFC, if accepted.)
+2. It is unclear how we should create `Request`s that use both unspecified and unknown entities. This has not been an issue so far since partial evaluation and its APIs are experimental, but we are looking to stabilize them soon. Currently the best (although not ideal) proposal is to use a builder pattern for `Request`s (e.g., `request.principal(Some(User::"alice"))` sets the principal field of a request). To set a field to be “unspecified” users should pass in `None` instead of `Some(_)`; to set a field to be unknown users should not call the field constructor at all.
```
cdisselkoen, line 74: Is this the most intuitive?  Have we considered whether doing it the other way around (None = unknown, not calling = unspecified) would be more intuitive for users?

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,194 @@
...
+  appliesTo { principal: [User], resource: [__cedar::Ghost] };
+```
+
+Cedar schema JSON syntax:
+
+```json
+"getFile": {
+    "appliesTo": {
+        "principal": [User],
+        "resource": [__cedar::Ghost]
```
cdisselkoen, line 154:
```suggestion
        "principal": ["User"],
        "resource": ["__cedar::Ghost"]
```

cdisselkoen (commented):
I'm in favor of this idea as a Cedar maintainer; it would make the Cedar implementation simpler and remove a source of bugs.  But I'm not entirely convinced that the benefits are worth the pain of a (language-)breaking change.  I'd like to see more feedback from potentially-affected users before I approve this -- do users agree that this is a good change?  Do the suggested fixes/migrations (defining their own "unspecified" entities) make sense in their applications, and make sense to them?

We currently have no breaking changes slated for 4.0 that would break existing _policies_ or _schemas_ (only breaks to our API surface); it would be very cool if on the 4.0 release we could say "All of your 3.x policies / schemas will continue to work exactly the same, only some APIs have been changed".  This RFC would spoil that by introducing a breaking change to the schema format for 4.0.

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,194 @@
...
+
+The current implementation of unspecified entities results in several corner cases in our code (especially the validator), which have been difficult to maintain (see discussion on [this recent PR](https://github.com/cedar-policy/cedar/pull/603)). The concept of unspecified entities is (in our opinion) confusing for users, and will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities.
+
+This RFC proposes to drop Cedar support for unspecified entities, instead recommending that users create their own application-specific entity types that act like unspecified entities.
+
+## Basic example
+
+Cedar uses a Principal, Action, Resource, Context \<P,A,R,C\> model for requests.
+This model can feel awkward in cases where there is not a clear principal and/or resource for an action.
+For example, consider a "getFile" action.
```
andrewmwells-amazon, line 24: I think this would be cleaner with  "CreateList" from TinyTodo.  IIUC, use `Application::"TinyTodo"` as the only `Application` is essentially our current recommendation for such a scenario. 

Maybe say something like: "The principal for `CreateList` is a User, but what is the resource? The list doesn't exist yet. One option is to leave the resource unspecified. Our current recommendation is to create a type to represent \"Anything in this application.\" In our TinyTodo example, we use `Application` and `Application::TinyTodo`."

I think it may be helpful to add policies for both examples.

Also, it is may be worth highlighting that this RFC is much more attractive once we have https://github.com/cedar-policy/rfcs/pull/53  (at least in my opinion)

andrewmwells-amazon (approved):

andrewmwells-amazon:
> We currently have no breaking changes slated for 4.0 that would break existing _policies_ or _schemas_ (only breaks to our API surface); it would be very cool if on the 4.0 release we could say "All of your 3.x policies / schemas will continue to work exactly the same, only some APIs have been changed". This RFC would spoil that by introducing a breaking change to the schema format for 4.0.

If this is a concern, I think we could do alternative A and provide a script to automatically migrate schemas.

shaobo-he-aws (approved):

text/0055-remove-unspecified.md
khieta: Fair question. I think if we keep our current `Request::new` api in addition to the builder pattern, then we need to stick with `None` == unspecified, since this is already the precedent. But if we get rid of the current constructor, then I'm open to swapping.

text/0055-remove-unspecified.md
khieta: Yeah, that example didn't end up making as much sense as I thought. I'll switch back to "createFile"

text/0055-remove-unspecified.md
khieta: Updated the the text & added a policy for "createList". Lmk what you think!

aaronjeline:
I personally like alt A, but I agree w/ craig about getting some feedback from users.

text/0055-remove-unspecified.md
andrewmwells-amazon: Looks good, resolving thread.

text/0055-remove-unspecified.md
andrewmwells-amazon: > if we get rid of the current constructor, then I'm open to swapping.

Even we get rid of the current constructor and think swapping would be more intuitive from a blank slate, I still worry it would be confusing for people who already used `None` for unspecified.

nakedible-p:
**I am in favor of this proposal.** But there's some things to be said.

First of all, originally, I thought it was a good idea that unspecified entities is a special case. The downside of explicity unspecified entities is that it is easy to write a policy like: 
```
permit(principal == Unauthenticated::"unspecified", action == Action::"SomeAction", resource == Res::"SomeResource")
```
The problem with this is that it's easy to write a policy that gives *less* access if you are authenticated than it gives if you are authenticated. Where as if the only way to match an unspecified entity is leaving it unspecified in the policy, then there's no risk of giving less access by authentication:
```
permit(principal, action == Action::"SomeAction", resource == Res::"SomeResource")
```

The other upside is for people coming from RDF background. An unspecified entity is very much like an RDF "[blank node](https://en.wikipedia.org/wiki/Blank_node)". But it is also a downside, as it isn't exactly like a blank node (a free variable).

However, I have since changed my opinion on the who matter. This is a very slight distinction, and it's not reasonable to expect people will think about this in enough detail to catch the distinction. Having unspecified entities behave exactly as other entities will simplify the brain process, simplify the API and force people to be more explicit about what is the same entity and what is not.

If "Alternative A" is chosen, possibly for just transitional use that will be deprecated, I would suggest `__cedar::None`. In my opinion, it's better than `Null`. An additional possibility would be `__cedar::Blank`, but I think that should be left for the possibility that Cedar at some point implements real blank node support.

And finally – I have been bitten by this problem myself. But it wouldn't have been any problem as long as Cedar would've documented the best practices for dealing with unspecified entities. In the end, the most important piece to ensure that you give guidance on the way to solve such issues.

khieta:
Thanks for the comments @nakedible-p! I'll add `None` to the list of candidates for Alternative A. Also thanks for the call out about documentation -- that's definitely something we need regardless of whether we accept this RFC or not. Hopefully coming in an upcoming blog post!

mwhicks1:
I'm strongly in favor of Alternative A. Here is my reasoning:

- I believe that doing nothing (not accepting this RFC) will create (or sustain) confusion for customers. There are several issues, well argued in the main body of the proposal.
  1. Trying to understand the `Request` API is difficult -- is `None` for an unspecified entity, or an unknown one? Adding a new `Request` builder doesn't fix the problem for the existing API. Since partial evaluation is a signature feature in demand by many customers, we should strive to make its APIs clean and understandable.
  2. Implicit unspecified entities were already confusing, as per the long discussion on `appliesTo` for schemas, and what it means to omit `principal` or set it to `principal: []`. This RFC does away with that confusion.
- I believe that leaving it to customers to design their own notion of an unspecified entity which intends to represent an irrelevant input, as proposed in the main body of this RFC, also risks confusion. Suppose a customer decides `AnyEntity::"Unspecified"` should represent the "don't care" principal in a request. They can easily defeat their own intent by writing a policy with the phrase `principal == AnyEntity::"Unspecified"` in it rather than leaving `principal` unconstrained (similar to what [nakedible-p](https://github.com/nakedible-p) says in his comment). Alternative A stops this footgun by using the validator to prevent `__cedar::Ghost::"..."` from appearing in a policy.
- Another benefit of Alternative A is that it provides an extremely clear migration path from 3.2 to 4.0. There are very simple steps you have to take to fix your code, with the validator to help. A simple tool can be used to transform schemas and Rust code invoking `isAuthorized` automatically, since it would know what "unspecified" entity ID to use.
- At the same time, customers can be encouraged to define their own notions of "placeholder" entity, to . We might even consider extending the enumerated entities feature with a keyword `ghost` that can (a) require the enumerated entity to be a singleton, and (b) signal to the validator that it should never appear in policies.

I would add that I'd prefer something other than `Ghost`. I'd propose either `None` or `Null`.

One other thing: I'm not convinced that `Application::"TinyTodo"` is the same thing as the unspecified entity. Per Alternative A, the unspecified entity should never appear in policies. But `Application::"TinyTodo"` explicitly appears, for a reason. The unspecified entity is useful when principal, resource, etc. are _irrelevant_, and never need to be "looked at" in a policy.

khieta:
Thanks for the feedback! After reading the replies from @nakedible-p and @mwhicks1, I'm starting to see why not being able to explicitly reference unspecified entities might be a good thing, since it encourages `principal` rather than `principal == ...` for a "made-up" entity type.

I also appreciate the comment from @cdisselkoen on the weight of a language-breaking change.

Based on the comments, I will be revising this RFC to propose Alternative A. Hopefully I'll have up a revised version later today!

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,199 @@
...
+But now "unspecified" entities are being used to represent multiple distinct concepts: a file store, (unidentified) users creating an account, and a system that manages accounts.
+
+This RFC proposes that users should instead explicitly create these "dummy" entities in an application-specific manner.
+For this example, the user should create entities of three types: `FileSystem`, `UnauthenticatedUser`, and `AccountManager`.
+The first will act as the principal for a "createFile" request, and the second and third will act as the principal and resource for a "createAccount" request.
+
+This is exactly the approach we used in the [TinyTodo application](https://github.com/cedar-policy/cedar-examples/tree/release/3.0.x/tinytodo): we made a special `Application::“TinyTodo”` entity to act as the resource for the “CreateList” and “GetLists” actions.
+This proposal is also in line with [our previous suggestion](https://cedar-policy.slack.com/archives/C0547KH7R19/p1706656288284189) to use an application-specific `Unauthenticated` type to represent "unauthenticated" users, rather than using unspecified entities.
```
khieta, line 67: Re @mwhicks1's comment above (pinning to the text to enable threaded discussion):

> I'm not convinced that `Application::"TinyTodo"` is the same thing as the unspecified entity. Per Alternative A, the unspecified entity should never appear in policies. But `Application::"TinyTodo"` explicitly appears, for a reason. The unspecified entity is useful when principal, resource, etc. are _irrelevant_, and never need to be "looked at" in a policy.

I'm still convinced that it is 🙂 `Application::"TinyTodo"` is an artificial entity with no attributes that is never used in policies conditions. In it's current form, it could have just as well been implemented using an unspecified entity.

mwhicks1: The reason I think they are different is the motivation for Alternative A: You should not be able to reference the unspecified entity in policies. But that's exactly what we do for `Application::"TinyTodo"`:
```
permit(principal, action in [Action::"CreateList", Action::"GetLists"], resource == Application::"TinyTodo");
```
To use the unspecified entity, we'd have to write the policy thus:
```
permit(principal, action in [Action::"CreateList", Action::"GetLists"], resource);
```
Listing the resource explicitly has at least two advantages:
- It's [useful for indexing](https://docs.cedarpolicy.com/bestpractices/bp-populate-policy-scope.html), which relies on scope constraints.
- In general, we think of `Application::"TinyTodo"` as a list "container", which [is useful to name explicitly in the policy](https://docs.aws.amazon.com/verifiedpermissions/latest/userguide/design-resources-in-containers.html). The reason is, in part, to avoid inadvertently authorizing a request. In this example, a client that mistakenly passes in a resource like `List::"mylist"` would still satisfy the policy, e.g., perhaps thinking that `Action::"GetLists"` is a way to show a list's contents.

khieta: A followup question: in my (limited) imagination, the only time a user would use an unspecified entity is when they are calling an action with a generic principal (e.g., an unauthenticated user) or an generic resource (e.g., an application, a file system, etc). If you think the resource case is better handled by a custom entity type (like `Application`), then do you still think that unspecified entities are really that useful?

mwhicks1: > To your second point, using `List::"mylist"` with `Action::"GetLists"` would result in a request validation error,

This is a good point. Of course, you will only get an error if you turn on request validation. Having the `== Application::"TinyTodo"` in the policy makes the situation evident in the policy, rather than in the schema, which I think is nicer. (Indeed, that's one reason we chose to include an `is` operator -- writing `resource is List` makes the policy clearer than simply saying that a particular action only applies to `List` objects in the schema.)

> do you still think that unspecified entities are really that useful?

I don't think they are that useful. :) To me, their best use is for actions that simply do not need a principal and/or resource at all, but the structure of Cedar requests and policies obligates a client to provide one. The "unspecified" resource acts as that stand-in.

khieta:
⚠️ Major revision posted ⚠️ 

This RFC has been updated to propose what was originally Alternative A -- i.e., it no longer proposes removing unspecified entities from the Cedar language.

----

While revising this RFC, I took some time to think carefully about what bothers me about unspecified entities, and the list I came up with was: 1) unclear terminology, 2) confusion with "unknown" entities when constructing a `Request`, and 3) awkward syntax in the schema.

I then reworked the proposal with these criteria in mind. Here’s the TLDR of the new proposal, numbered based on which concern is addressed:

1. Update the terminology surrounding unspecified entities, instead calling them **default entities**. Default entities will have type `__cedar::Default`.
2. Define a new `RequestEntity` type to use as input to the `Request` constructor. The enum has variants `EntityUid`, `Unknown`, and `Default`.
3. In order to specify that an action applies to a default entity, the schema `appliesTo` list must contain `__cedar::Default`.

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,225 @@
...
+- Started: 2024-02-27
+
+## Summary
+
+Cedar currently supports _unspecified entities_, which are entities of a special, unique type that have no attributes, and are not ancestors or descendants of any other entity in the store. Unspecified entities are intended to act as placeholders for entities that don’t influence authorization (see examples below).
+
+The current implementation of unspecified entities results in several corner cases in our code (especially the validator), which have been difficult to maintain (see discussion on [this recent PR](https://github.com/cedar-policy/cedar/pull/603)). The concept of unspecified entities is (in our opinion) confusing for users, and will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities.
+
+This RFC proposes to change the terminology surrounding unspecified entities, instead calling them "default" entities.
+It also proposes to give an explicit name to the unspecified entity type (`__cedar::Default`) and to require this name to be used in requests and schemas that use this type.
```
khieta, line 19: We had previously discussed `None` and `Null`, but I think "default" better represents how the entity should be used: you should use it when you want to make an authorization request, but the principal/resource don't matter, so you just want _some default value_.

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,225 @@
...
+- `Arbitrary`: same issue as “any”
+- `Unspecified`: potential for confusion with partial evaluation “unknown”
+- `Mock`: implies that this entity type should be used for debugging/testing
+
+### `Default` vs. `EntityUid(__cedar::Default::"eid")`
+
+Under the current proposal, there are two ways to pass an unspecified entity into a `Request`: `RequestEntity::Default` and `RequestEntity::EntityUid(__cedar::Default::"eid")`.
+Both are equivalent, although the latter allows the user to assign an particular name ("eid") to the entity, which could be helpful when debugging error messages.
+When using the `Default` variant, Cedar will assign the entity id "principal", "action", or "resource" depending on where the entity is used in the request.
+
```
emina, line 162: In general, I like this option, but I'm concerned about having two distinct ways to do the same thing.  If I understand correctly, users can construct an _arbitrary_ entity of the `__cedar::Default` and pass it in?  So, I could construct a request that binds both the `principal` and `resource` variables to, say, `__cedar::Default::"foo bar baz"`.  Is that right (or maybe I'm misreading)?

This seems potentially dangerous and confusing.  My inclination would be to restrict `__cedar::Default` to be an enum over three values (`__cedar::Default::"principal"`, `__cedar::Default::"action"`, `__cedar::Default::"resource"`), and to restrict the API accordingly, to make it impossible to construct requests that mix up the bindings of Cedar variables to these values.

khieta: Good point. I also don't think we want users using the same default entity for both principal and resource, so I'll update the proposal to just disallow the second case. (It had questionable usefulness anyways.)

nakedible-p:
I think there's one distinction that is missed - and that's that I don't think in general `__cedar::Default::"principal" == __cedar::Default::"principal"` should be true. And unspecified principal should refer to some principal that isn't given in, but which should be distinct from all other principals, including other unspecified principals. If something stores the current principal as the `owner` of some object, and the current principal is unspecified, as is common in unauthenticated requests, then that doesn't mean that every unspecified request should own the object. Like mentioned before, there are equivalences to blank nodes in RDF. An unspecified principal should be a "free variable", like a blank node is – not something that is necessarily the same instance. But, of course, in some cases it does make sense to have have a distinct principal that is just one instance - so usages will differ.

Personally I will still use an explicit custom principal type for "unauthenticated" requests, and I will have the principal ID generated as ULID, so that it will never be equal to any previously generated principal. That means that personally for me it will not really matter how the semantics of the default principal will be set in Cedar. I would still suggest that Cedar will recommend people to always explicitly create their "default" principal types by themselves, and not use the `__cedar::Default` types except for compatibility with older stuff.

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,220 @@
...
+Cedar currently supports _unspecified entities_, which are entities of a special, unique type that have no attributes, and are not ancestors or descendants of any other entity in the store. Unspecified entities are intended to act as placeholders for entities that don’t influence authorization (see examples below).
+
+The current implementation of unspecified entities results in several corner cases in our code (especially the validator), which have been difficult to maintain (see discussion on [this recent PR](https://github.com/cedar-policy/cedar/pull/603)). The concept of unspecified entities is (in our opinion) confusing for users, and will only become more confusing once we stabilize the partial evaluation experimental feature, which allows “unknown” entities that are subtly different from “unspecified” entities.
+
+This RFC proposes to change the terminology surrounding unspecified entities, instead calling them "default" entities.
+It also proposes to give an explicit name to the unspecified entity type (`__cedar::Default`) and to require this name to be used in requests and schemas that use this type.
+
+## Basic example
+
```
mwhicks1, line 22: Nit: This section is basically the full design, not just a basic example. It might be easier to read/understand this RFC if this section were much shorter, and the details around the design were move further below.

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,220 @@
...
+Here are some options we’ve ruled out:
+
+- `Any`: possibility to confuse the scope condition `principal` with `principal is __cedar::Any`. The first applies to any principal, while the latter applies to only the “unspecified” entity.
+- `Arbitrary`: same issue as “any”
+- `Unspecified`: potential for confusion with partial evaluation “unknown”
+- `Mock`: implies that this entity type should be used for debugging/testing
+
+### Upgrade strategy
+
```
mwhicks1, line 159: Also mention the changes to `appliesTo` etc. in schemas as part of upgrades?

mwhicks1 (approved):

text/0055-remove-unspecified.md
khieta: Yes it got inflated over time 🙂 Will clean up tomorrow.

khieta:
Re @nakedible-p:

> I don't think in general `__cedar::Default::"principal" == __cedar::Default::"principal"` should be true. An unspecified principal should ... be distinct from all other principals, including other unspecified principals.

That's a good point. I agree that unspecified principals in different requests should refer to conceptually different entities.

Nevertheless, I think it is ok that all unspecified principals will end up being `__cedar::Default::"principal"` (under the current design) because there is no way for a principal from one request to interact with a principal from another request. We won't support storing anything with type `__cedar::Default` in the entity store or context, so it will not be possible to, for example, store an unspecified principal as the owner of another entity. Unspecified entities can only exist in a `Request`'s principal/action/resource.

nakedible-p:
If it's limited to only requests, then I also see no problems in having a single value like `__cedar::Default::"principal"`. As a user, I'd expect anything starting with `__cedar` to be a bit special anyway. Thanks for thinking about this.

text/0055-remove-unspecified.md
andrewmwells-amazon: Agree `Default` make sense.

text/0055-remove-unspecified.md
andrewmwells-amazon: I also think `Application::"TinyTodo"` should be viewed as an unspecified entity. My reasoning is that a policy store should only contain policies for the application in question, so saying 'permit ... when I'm in Application::"TinyTodo"' doesn't add information.

I don't find the objection that users may apply `Action::"GetLists"` to a `List` convincing. When writing code, I'd expect the author to refer to the schema. Assuming they misunderstand `Action::"GetLists"`, they should reach some point like:
```
fn GetLists(user: User, list: List) -> Vec<ListItem> {
  let req = build_req_get_lists(user, list);
  if authorizer.is_authorized(req) {
    return Error::Unauthorized();
  }
  ...
```
The author needs to build a request given `user` and `list`. To decide how they will almost certainly look at the schema, and it will quickly become clear they misunderstand what `Action::"GetLists"` does.
I suspect adding `Application::"TinyTodo"` will actually make this more confusing. They need to understand why this entity was created and that there is no corresponding argument for them to pass. Using unspecified, it becomes clear that there's no `List` argument needed in the above code.

I do think `Action::"GetLists"` is poorly named, but that's a separate problem.

text/0055-remove-unspecified.md
cdisselkoen: I like `Default`

cdisselkoen (approved):
This version of the RFC makes a lot of sense to me, and I'm in favor.  I particularly like how it enables a deprecation period and smooth transition across the breaking change.

shaobo-he-aws (approved):

emina (approved):

max2me:
I will be a dissenting voice on this one as I strongly prefer Alternative A - getting rid of the unspecified entity type altogether in favor of enumerated entity types introduced in RFC53.

Both the main proposal and Alternative A are breaking changes, so I'd rather go a step forward and require customers to update schema, code, and policies once for the benefit of a brighter future (the main proposal would require customers to update just schema and code, yet all the same testing will have to occur).

My concerns with the main proposal:

* Lack of symmetry -- you omit something in the policy and yet explicitly mention it in the authorization code and schema. This leads to a fragmented experience of working with Cedar as a technology.
* Surprise factor -- I suspect that most customers who leave principal/resource unscoped in policies will not realize that in addition to all the entities they reference in their code, there is a hidden default type.

In my opinion, occasional duplication of policies and the need to define your own default type is a small cost to pay for the overall simplicity of the system.

andrewmwells-amazon (approved):

text/0055-remove-unspecified.md
aaronjeline: `Default` also ties in nicely with rust naming conventions

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,248 @@
...
+        "principal": ["User"],
+        "resource": ["__cedar::Default"]
+    }
+}
+```
+
+Users will not be allowed to include entities of type `__cedar::Default` in the store, which ensures that these entities will never have attributes or ancestors/descendants, which are properties we will rely on during validation.
+Users will not be allowed to reference the type `__cedar::Default` in their policies (so neither `principal == __cedar::Default::"principal"` nor `principal is __cedar::Default` are allowed).
+Users do not need to add `__cedar::Default` to their schemas (and it will be an error if they do) because `__cedar::Default` is defined automatically.
+Finally, users cannot create entity uids with type `__cedar::Default`, which prevents using default entities with the `RequestEntity::EntityUid` constructor.
```
mwhicks1, line 163: What is the goal of these restrictions? Is there a property we can state and prove about them?

I think that the goal is to prevent users from hijacking that implementation mechanism of default entities which would now be explicitly exposed. When the mechanism is fully hidden, as today, we don't have to worry about a user trying to make a new kind of default entity, or giving it ancestors, or referring to it in policy expressions. But now that we've given it an explicit name, there's mischief involving that name we have to prevent.

I originally thought the goal of the restrictions was different: To ensure that policies that might use the default entity _ignore_ it, since the point is that the entity _doesn't matter_. But these restrictions don't do that. You can still write a policy with a clause like `principal == User::"Alice"` and if you pass in the default entity for the principal then this expression will _not_ be true, affecting policy satisfaction. You could also write `principal has manager && principal.manager == User::"Bob"` and again this will evaluate just fine with the default entity. This feels weird. Further up we write

> If a policy uses an _unknown entity_ in a non-trivial way, then evaluation will produce a residual.
> If a policy uses an _unspecified entity_ in a non-trivial way, then evaluation will produce an error.

The caveat "non-trivial" seems strange. What makes behavior trivial? If the use of a default entity affects the policy evaluation outcome in _any_ way, that seems "non trivial".

It would make more sense to me to restrict _all_ uses of default entities. For example, if `Action::"createFile"` was specified to as having a default resource entity (the schema would say `createFile` applies to `__cedar::Default`), then a policy that could use `Action::"createFile"` as its action should always ignore the `resource` component. If it doesn't, that's an error. In other words, if you wrote the following, you should get an evaluation and validation error for the expression `resource == File::"Root"`:
```
permit(principal, action == Action::"createFile", resource == File::"Root")
```
This is a more invasive change, so probably we don't want to do it. But at the same time I think that sheds some light on the confusion around unspecified entities today. This RFC is not proposing to change the semantics of unspecified entities make them less confusing, it is just changing the implicit specification we have now to be a (equally confusing) explicit one.

text/0055-remove-unspecified.md
```diff
@@ -0,0 +1,248 @@
...
+We need to make sure that we provide enough value in the major version with this change for the upgrade to be worth it.
+
+## Alternatives
+
+### Alternative A - Get rid of the unspecified entity type
+
+The current proposal can be viewed as a re-framing of the current unspecified type.
+A more radical option (as originally proposed by this RFC) is to completely drop Cedar support for unspecified entities, instead recommending that users create their own application-specific entity types that act like unspecified entities.
+
+This is (arguably) the approach we used in the [TinyTodo application](https://github.com/cedar-policy/cedar-examples/tree/release/3.0.x/tinytodo): we made a special `Application::“TinyTodo”` entity to act as the resource for the “CreateList” and “GetLists” actions.
```
mwhicks1, line 216: This RFC specifically forbids mentioning the default entity in policies, but TinyTodo policies do mention `Application::"TinyTodo"` explicitly:
```
permit(principal, action == Action::"CreateList", resource == Application::"TinyTodo");
```
If we think these things are the same, then we should allow the default entity to be mentioned in policies. Per my comment above, I think that's a modification in the wrong direction: We should be even _more_ restrictive on the use of default entities, not less.

mwhicks1 (changes requested):
I'm going to reverse my position and agree with Max that alternative A (the original proposal) is the one to use. Sorry!

Per the comment I left in-line, this proposal is helpful in making the use of unspecified/default entities explicit in schemas, rather than implicit. Doing so exposes a choice to users they may not have appreciated, so they think about it more, and it will simplify the partial evaluation API. Giving explicit default entities the same semantics as today's implicit ones eases the upgrade path.

But retaining the semantics of unspecified entities is disappointing, because it's confusing: A default entity is not truly ignored, as we might expect/hope, it is just a special, but second-class, kind of entity. Exposing this truth will further confuse users who are already confused: Why am I not allowed to write the default entity in policies? Why can't I create my own entities with that type? This relates to the lack of symmetry Max was talking about.

Alternative A "solves" these problems by forcing users to think in terms of default entities they define themselves. I think the simplicity is a win. Its main downside is a potentially harder upgrade path. I wonder if we can dig into that a little more. If users are using default entities as intended, then their policies will be ignoring them, and the upgrade path is easy. It is only if users are misusing default entities today that upgrades might be harder; e.g., they are using them in `==` checks etc. Could we detect these situations with a tool, to provide help?

Note that we could also consider an Alternative B that goes with the core proposal but adds _more_ restrictions on default entities' use, per my inline comment. That would ensure default entities are truly ignored, and might be a good thing. It would address @nakedible-p's concern 

> I don't think in general __cedar::Default::"principal" == __cedar::Default::"principal" should be true. 

We don't need to do this now; a future improvement to this effect would be non-breaking.

text/0055-remove-unspecified.md
shaobo-he-aws: `Any` sounds better for me.

shaobo-he-aws:
> But retaining the semantics of unspecified entities is disappointing, because it's confusing: A default entity is not truly ignored, as we might expect/hope, it is just a special, but second-class, kind of entity. Exposing this truth will further confuse users who are already confused: Why am I not allowed to write the default entity in policies? Why can't I create my own entities with that type? This relates to the lack of symmetry Max was talking about.

I agree. I think whenever we decide in the end, we should record the rationales in a FAQ or best practice doc.

khieta:
Based on in person discussion, I have revised the RFC (again 🙂) to make dropping unspecified entities the main proposal. The general consensus in discussion was that the proposal to rebrand unspecified entities as “default” entities felt like a bandaid rather than a true fix, and we would eventually want to remove unspecified/default entities anyways.

However, to support backwards compatibility, for a period of time we will provide APIs (behind a flag) that maintain the prior behavior. See the new text for more details!

mwhicks1 (approved):
Looks good!

nakedible-p:
I'm slightly happier with this proposal now than the one with `__cedar::Default`, as it was originally. Like said earlier, we will anyway define our own entity types for unspecified entities, so this just makes it explicit and simplifies the API. So :shipit:

nakedible-p (approved):

shaobo-he-aws (approved):

khieta:
The final comment period (FCP) for this RFC is starting now, with intent to accept 🎉 

The FCP will end 2024-05-28 at noon PT / 3pm ET / 7pm UTC. Please add comments, and especially any objections, if you have any, before then. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

emina (approved):

emina (approved):

PR #57, opened 2024-03-07
General multiplication in Cedar

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/general-multiplication/text/0057-general-multiplication.md)

### DIFF
--- a/text/0057-general-multiplication.md
+++ b/text/0057-general-multiplication.md
@@ -1,0 +1,111 @@
+# General multiplication in Cedar
+
+## Related issues and PRs
+
+- Reference Issues: related to [RFC 50]
+- Implementation PR(s): [cedar#702]
+
+## Timeline
+
+- Started: 2024-03-07
+
+## Summary
+
+Allow multiplication of arbitrary expressions (that evaluate to Long), not just
+multiplication of an expression by a constant.
+
+## Basic example
+
+```
+permit(principal, action, resource) when {
+    context.foo * principal.bar >= 100
+};
+```
+
+## Motivation
+
+Today, Cedar only supports multiplication of two operands if (at least) one
+operand is a constant.
+In the more general n-ary (chained) multiplication case, Cedar requires that at
+most one operand is a non-constant, but allows the constants and non-constant to
+appear in any order.
+
+- This creates some confusion for Cedar users. No other popular programming
+  language has this restriction on multiplication.
+- This leads to bugs and confusing behavior. To enforce this restriction,
+  Cedar's internal AST node for multiplication only supports multiplying an
+  expression by a constant. But this requires the Cedar parser to re-associates
+  (and re-order) operands in a chained multiplication, which leads to confusion
+  and problems as described in [RFC 50].
+- This limits the power of Cedar by preventing users from writing policies
+  involving multiplication where neither operand is known at policy-authoring
+  time.
+
+This RFC proposes to relax all of these restrictions and allow multiplication
+of arbitrary expressions (that evaluate to Long).
+In other words, multiplication will be treated just like addition: `*` will be
+valid in all positions and situations where `+` is valid today.
+
+## Detailed design
+
+This will actually simplify the evaluator code, validator code, and internal AST.
+Multiplication will become an ordinary `BinaryOp`, and no longer need a special
+AST node.
+The evaluator and validator can handle multiplication in the same way as
+addition and in the same code paths.
+
+With this RFC, the policy parser becomes strictly more permissive:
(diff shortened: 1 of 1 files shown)


aaronjeline (approved):
Seems reasonable to me, maybe include a bit about it being easier to overflow via multiplication compared to additon

emina (approved):
I'm in favor of this generalization.

text/0057-general-multiplication.md
```diff
@@ -0,0 +1,91 @@
...
+## Motivation
+
+Today, Cedar only supports multiplication of two operands if (at least) one
+operand is a constant.
+In the more general n-ary (chained) multiplication case, Cedar requires that at
+most one operand is a non-constant, but allows the constants and non-constant to
+appear in any order.
+
+- This creates some confusion for Cedar users. No other programming language
+  that I know of has this restriction on multiplication.
```
emina, line 34: Style nit:  this is the only occurrence of first person singular in the RFC.  I'd replace it or drop the sentence.

cdisselkoen: Fixed

andrewmwells-amazon (approved):
Seems like a good idea.

text/0057-general-multiplication.md
```diff
@@ -0,0 +1,91 @@
+# General multiplication in Cedar
+
+## Related issues and PRs
+
+- Reference Issues: related to [RFC 50]
```
mwhicks1, line 5: Is there a correctness issue that this relates to, e.g, relating to overflow and ordering issues that you hint at below? If so, reference that here.

text/0057-general-multiplication.md
```diff
@@ -0,0 +1,91 @@
...
+more expensive than multiplication by an arbitrary constant (ignoring special
+cases such as 0 or constant powers of 2).
+Some multiplication expressions will be expensive to analyze (computationally),
+but that is already true for some other kinds of Cedar expressions that don't
+involve multiplication.
+In general, we'd prefer to use a linter to warn users about (these and other)
+expensive expressions, rather than use restrictions on the Cedar language.
+
+This RFC firmly closes the door on changing Cedar's default numeric type to
+bignum. However, that door is already basically closed.
```
mwhicks1, line 75: Might mention that bigints can always be added as an extension type later.

mwhicks1 (approved):
LGTM

text/0057-general-multiplication.md
cdisselkoen: Added a comment.

text/0057-general-multiplication.md
```diff
@@ -0,0 +1,92 @@
...
+## Detailed design
+
+This will actually simplify the evaluator code, validator code, and internal AST.
+Multiplication will become an ordinary `BinaryOp`, and no longer need a special
+AST node.
+The evaluator and validator can handle multiplication in the same way as
+addition and in the same code paths.
+
+This RFC does not introduce any new categories of errors, and does not result in
+changed behavior for any policies that are valid today.
```
khieta, line 58: Are there any expressions for which the previous semantics will not overflow, but the new one will? One half-baked idea based on #50: I would expect an expression like `CONST * CONST * content.value` to overflow for large `CONST` and `content.value=0` under the proposed semantics. But under the old semantics, during parsing this expression is converted to `content.value * CONST * CONST` which cannot overflow.

Should we consider this a behavior change?

khieta (approved):
I'm in favor of this proposal, and prefer it to #50.

text/0057-general-multiplication.md
cdisselkoen: There is not currently a referenceable issue, beyond conversation on RFC 50

text/0057-general-multiplication.md
cdisselkoen: Unfortunately, your example is correct.  I will update the language here.

cdisselkoen: Revised.

john-h-kastner-aws (approved):

cdisselkoen:
Given the large number of approvals, the final comment period (FCP) for this RFC is starting now, with intent to accept the RFC.

The FCP will end 2024-03-14 at noon PT / 3pm ET / 7pm UTC.  Please add comments, and especially any objections, if you have any. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

shaobo-he-aws:
Makes sense to me.

shaobo-he-aws (approved):

PR #58, opened 2024-03-12
Cedar standard library

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/cdisselkoen/standard-library/text/0058-standard-library.md)


aaronjeline (approved):
I really like this. I think having a stdlib is a good idea. 
This RFC doesn't actually say what the type deceleration for `OIDC::User` is, is that intended?
I like the listing of alternatives, but think they can all weight for a future RFC when needed.

text/0058-standard-library.md
```diff
@@ -0,0 +1,287 @@
...
+```
+entity MyEntityType extends A in [B];
+```
+We could instead require `extends` to appear _after_ `in` but before attributes:
+```
+entity MyEntityType in [B] extends A;
+```
+Or we could allow both.
+
+We could also consider some keyword other than `extends`, or some syntax other
```
aaronjeline, line 278: Downside of `extends` is that it implies object-oriented style inheritance. Which may be esp confusing since Cedar's type system doesn't have subtyping. I'll throw some other suggestions out:
`expands`
`copies`
`elaborates`

shaobo-he-aws:
A meta question: Is it for human-readable format only? If not, we should provide the JSON representation as well.

D-McAdams:
I support the "extends" feature and have several scenarios where it would help us manage schemas in large-scale systems with many services and application builders, in which central teams define a common structure which others can expand.

The Cedar standard library is a special case in which a common structure is so ubiquitous as to be nearly universal. I think this could be useful, but will naturally take longer to finalize these types and so it makes sense to decouple the standard type definitions (like OIDC User) from the rest of the proposal. We'd probably want to run the standard types through a process that looks like a standards community, including audiences from the OpenID Foundation, for example.

D-McAdams:
When it comes to bikeshedding the naming, if we sought to avoid "extends" because of it's object-oriented connotations, I would be in favor of the aforementioned suggestion of "expands".

text/0058-standard-library.md
aaronjeline: `augments`

cdisselkoen: These are good callouts.  I added a point about subtyping to Drawbacks, and I added these suggestions for alternate names for `extends` to Alternative H.

cdisselkoen:
@aaronjeline says
> This RFC doesn't actually say what the type deceleration for `OIDC::User` is, is that intended?

@D-McAdams says
> will naturally take longer to finalize these types and so it makes sense to decouple the standard type definitions (like OIDC User) from the rest of the proposal. We'd probably want to run the standard types through a process that looks like a standards community, including audiences from the OpenID Foundation, for example.

These are good thoughts.  I would say that this RFC proposes three things:

* the concept/mechanisms for a standard library
* the `extends` keyword (or whatever we end up bikeshedding it as)
* the idea of adding `OIDC::User` as a standard library type

but this RFC does not propose a specific declaration for `OIDC::User`, only proposes that we somehow arrive at a consensus one, and then add it to the standard library. I propose that we can delay hashing out the details of the `OIDC::User` declaration to an implementation PR, or to some other kind of process with more stakeholders as @D-McAdams suggests.

jszmajda:
I'm a big fan of `extends`, that will make my life easier in my own schema definitions. One detail on the stdlib - I like the idea of a reference library, but I'm not a fan of prepending it to the schema because from what I understand this would allow any policy author to reference those entities regardless of whether they're being used in a particular application. If there's a way to "not import unless referenced or used somehow," that would solve my concern.

text/0058-standard-library.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Basic example
+
+```
+type User = __cedar::OIDC::User;
+```
+or
+```
+entity Group { myAttribute: String };
+entity User extends __cedar::OIDC::User in [Group] { emailOptOut: Boolean };
```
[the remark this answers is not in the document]
mwhicks1: There is also a technical issue: Suppose I have
```
entity Group { myAttribute: String, sibling: [Group] };
entity HierarchicalGroup extends Group { tag: String };
```
A strict interpretation of `extends` would make `HierarchicalGroup` equivalent to 
```
entity HierarchicalGroup { myAttribute: String, sibling: [Group], tag: String };
```
But users might be expecting that `sibling`'s type is updated with the extended type, i.e.,
```
entity HierarchicalGroup { myAttribute: String, sibling: [HierarchicalGroup], tag: String };
```
Doing so would be important for strict validation, since it does not allow subtyping. With permissive validation, leaving `sibling` as `[Group]` is less problematic because `HierarchicalGroups` are effectively subtypes of `Group`s, and we could introduce subtyping edges explicitly to support that.

mwhicks1 (commented):
I would propose we table this RFC and instead consider a library mechanism for schemas, and then rely on that for different user communities to build a de-facto standard libraries.

text/0058-standard-library.md
```diff
@@ -0,0 +1,299 @@
...
+would develop organically without this RFC.
+3. Users may incorrectly assume that `extends` has some kind of impact at
+runtime -- for instance that concrete values of `MyEntityType` inherit attribute
+values from a concrete value of `BaseEntityType` somehow.
+Hopefully we can address this with clear documentation.
+There's also nothing preventing us from providing API helpers that allow
+extending concrete `Entity` objects of a base type corresponding with a schema
+`extends` declaration (although this is not proposed in this RFC).
+4. Users may incorrectly assume that `extends` provides subtyping, while this
+RFC does not propose that it provides subtyping.
```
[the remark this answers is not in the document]
mwhicks1: Actually, it won't unless we add a specific rule that says it does. But we can probably do that.

text/0058-standard-library.md
```diff
@@ -0,0 +1,299 @@
...
+For instance, users may assume that an action with principal type
+`BaseEntityType` could also be used with principal type `MyEntityType` which
+`extends` `BaseEntityType`; but this is not the case in the current proposal.
+It's also possible that the keyword being named `extends` in particular might
+make this problem worse due to existing associations with object-oriented
+inheritance; see Alternative H.
```
khieta, line 130: This is my main gripe with the word `extends`. If we don't allow for this kind of subtyping then I think we should switch to `expands` (or similar).

text/0058-standard-library.md
```diff
@@ -0,0 +1,299 @@
...
+### Alternative H: Bikeshedding the `extends` syntax
+
+This RFC proposes that `extends` must appear before `in` and attributes:
+```
+entity MyEntityType extends A in [B];
+```
+We could instead require `extends` to appear _after_ `in` but before attributes:
+```
+entity MyEntityType in [B] extends A;
+```
```
khieta, line 283: I like the RFC as-is; it's useful to have `MyEntityType` and `A` as close together as possible.

khieta (commented):
I like the idea of the `extends` operator in schema (although I'd prefer a different name). But I'm not so keen on the idea of the standard library -- I think the OIDC definition should live in its own repository controlled by a standards committee / invested third parties (so, Alternative A).

Would you consider splitting this RFC into two?

andrewmwells-amazon:
I think it would make more sense to break this into two RFCs: adding libraries to Cedar (combining this RFC with parts of https://github.com/cedar-policy/rfcs/pull/61) and an RFC proposing what types to include in the standard library.

nynymike:
Here's a [schema](https://github.com/nynymike/cedar-playground/blob/main/gluu_lock.cedarschema ) we're using for our [Janssen Project](https://jans.io) Cedar science project: 

A few things to think about that this schema has:

1. I left the user claim section pretty sparse, as domains can always extend.
2. I added the ubiquitous "Roles"
3. I added some standard "Context" claims... which will make all the AI/ML people excited
4. A lot of tokens share certain standard claims, like `iss` or `jti`... so I abstracted these
5. A lot of enterprise policies are about OAuth clients and OAuth access tokens, so these need to be modeled too
6. It's important to make the relationship between the trusted issuer and the assertion

cdisselkoen:
These last three comments (@khieta, @andrewmwells-amazon, and @nynymike) all make good points; I'll work on splitting this RFC into (at least) two parts, separating the ideas of library mechanisms, `extends`, and a Cedar-maintained "standard library".

cdisselkoen:
As the first step of that, I just posted [RFC 69: Schema libraries](https://github.com/cedar-policy/rfcs/pull/69). Comments welcome.

text/0058-standard-library.md
```diff
@@ -0,0 +1,299 @@
...
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2024-03-12
+
+## Summary
+
+Create a Cedar Standard Library of type declarations that are likely to be
+useful to many Cedar users, and are distributed along with Cedar.
```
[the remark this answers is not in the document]
nynymike: I agree.. I would call this **Add `extends` Mechanism for Cedar Schema**

PR #61, opened 2024-03-20
Cedar Functions

aaronjeline:
[Rendered](https://github.com/cedar-policy/rfcs/blob/aaronjeline/functions/text/0059-functions.md)


text/0059-functions.md
```diff
@@ -0,0 +1,282 @@
...
+vs
+```
+decimal(1,1)
+semver(2,3,5)
+```
+
+### Readability/Policies are no longer standalone
+A policy can no longer be read by itself, it has to be read in the context of all function definitions it uses.
+Policies that use an inordinate number of functions may be very opaque to read.
+
```
emina, line 212: Another downside of this proposal is that it's unclear (a) where the functions are defined and how they are managed, (b) how they are passed to Cedar at runtime, and (c) which functions are available when to what policies. 

In a simple world where all Cedar policies are stored in a single file, we can put the functions in the same file.

But what about a hosted service, where policies are stored in a DB, and the service manages policies for many different applications and clients? 

What functions are available to what clients?  Where are they stored?  How do clients discover functions?  Does the service have to effectively do runtime linking, or are functions inlined at typechecking / creation type?  Does the service have to gather the set of functions to pass in to Cedar for evaluation?

text/0059-functions.md
```diff
@@ -0,0 +1,282 @@
...
+2. Are type annotations enforced dynamically à la Contracts, or are they just ignored at runtime?
+  1. If they are dynamically enforced, that implies access to the schema to unfold type definitions. It also may introduce redundant type checking.
+3. Are type annotations required or optional?
+
+
+Leaving them out for this RFC only precludes only one future design decision, making type annotations required.
+This decision feels unlikely, as we want Cedar to be useful without using a schema/the validator.
+Adding optional type annotations in the future is backwards compatible. 
+
+### Let functions call other functions
```
emina, line 246: It would be impossible to forbid cycles, unless all functions are statically known upfront, which brings up the function linking and management question above.

aaronjeline: Even if it was, I think the readability constraint is also a reason to not allow this

text/0059-functions.md
mwhicks1: My perspective is that these are good questions but outside the scope of this RFC. At least for now, I propose we just support functions as being definable in the same file as policies. 

For context, I note we have similar questions around how to store and manage the "Cedar standard library" from [RFC 58](https://github.com/cedar-policy/rfcs/blob/cdisselkoen/standard-library/text/0058-standard-library.md). We also don't solve the problem of bringing together policies authored separately, e.g., by an app owner and its users. We can leave solutions of these questions for later, and perhaps even outside the language.

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+1. The function name (potentially namespaced)
+2. A comma separated list of arguments
+
+Here is an example:
+
+```
+foo(1,2, principal.name)
+```
+
+Function arguments are eagerly evaluated (i.e., call-by-value), as with extension functions today.
```
mwhicks1, line 110: We say the say below we will just inline the function body at each place it's called, when doing validation. If we are doing that, then we are essentially doing lazy evaluation, not eager evaluation. This could lead to validation soundness issues. In particular:
```
function drop(a,b) { a };
permit(principal,action,resource) 
when { drop(true,1+"hello") };
```
This policy will validate because once we've inlined the function we'd have
```
permit(principal,action,resource) 
when { true };
```
But if we actually do CBV evaluation then this policy will fail because `1+"hello"` will fail. We can probably solve this problem by validating the argument expressions of a function call individually, in addition to validating the entire policy after inlining.

But then there's also the problem with analysis: Our logical encoding would have to do CBV to be consistent with the actual evaluator, rather than just inlining, or else we need to prove that eager-eval(e) = lazy-eval(e) for all validated e.

For my money, I'd vote to do lazy evaluation and calling these things macros, replacing the `function` keyword with `def`. This allows users to define more expressive language features. For example:
```
def implies(e1,e1) {
  !e1 || (e1 && e2)
}
permit(principal,action,resource) when {
  implies(principal has attr,
          resource has attr && principal.attr == resource.attr)
};
```
The above won't work with CBV because the second expression to the call to `implies` will fail if `principal.attr` does not exist.

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+`principal`/`action`/`resource`/ `context` are considered to be more tightly bound then function names.
+This means that while you could name a function `principal`, you could never call it. (This should probably be a validator warning)
+Function name conflicts at the top level are a parse error.
+Function names may shadow extension functions (results in a warning).
+
+### Formal semantics
+This RFC adds one new evaluation rule to Cedar:
+If $f$ is the name of a declared function, _def_($f$) is the body of the definition, and $p_1, ..., p_n$ is the list of parameters in the definition:
+$f(v_1, ..., v_n) \rightarrow$ _def_$(f) [p_1 \mapsto v_1, ..., p_n \mapsto v_n]$
+Where $e[x \mapsto v]$ means to substitute the $v$ for $x$ in $e$, as usual.
```
mwhicks1, line 153: Per my comment above, I'd propose to do lazy evaluation instead. In other words:
```
def f(x1,x2) { body }
permit(...) when { ... f(e1,e2) ... };
```
is equivalent to
```
permit(...) when { ... body[x1 --> e1, x2 --> e2] ... }
```

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+The grammar of Cedar expressions is unchanged, as we re-use the existing call form.
+```
+function ::= 'function' Path '(' Params? ')' '{' Expr '}' ';'
+Params ::= Ident (',' Ident)? ','?
+```
+Note the `Params` non-terminal allows trailing commas in parameter lists.
+
+### Validation
+
+The validator typechecks functions at use-sites via inlining.
```
mwhicks1, line 165: This is unsound if you are doing eager evaluation. See my long comment above, which proposes lazy evaluation instead.

You should probably extend this section to be "Validation and Analysis" and propose the inlining as a first step before analyzing policies too.

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+This would allow functions to typechecked in absence of a use-site, and allow for easier user specification of intent.
+It would also probably result in clearer type checker error messages.
+
+This introduces the following questions:
+
+1. Do we allow `type` declaration in policies, or just in schemas?
+2. Are type annotations enforced dynamically à la Contracts, or are they just ignored at runtime?
+  1. If they are dynamically enforced, that implies access to the schema to unfold type definitions. It also may introduce redundant type checking.
+3. Are type annotations required or optional?
+4. Will we have types to support generics, i.e., polymorphism?
```
mwhicks1, line 252: My answer:
1. If we ever have type annotations, they should not be in policies. We have so far had a tenet of a clear separation between policies, which are dynamically typed and have no type annotations, and schemas, which support static validation and enumerate/annotate with types.
2. Ignored, per above.
3. Optional, for maximal flexibility and simplicity
4. Do not support generics, but that's Ok when types are optional -- you can leave them off to get this

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+### Let functions call other functions
+As long as cycles are forbidden and functions as arguments are disallowed, we could allow functions to call other functions without sacrificing termination.
+However, the potential complexity explosion is high, and it's backwards compatible to add this later.
+
+<!-- ### Make parse errors runtime errors
+Make (any subset of) the following errors runtime errors instead of parse errors: 
+
+1. Application of non-function
+2. Non-application of function
+3. Function application with incorrect arity -->
```
mwhicks1, line 268: I commented this out because it was hard for me to imagine why you would want this. You can always detect these problems perfectly (no false alarms) at compile-time, right?

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+  foo(foo) // Also fine
+}
+
+```
+We argue against this on principle of least surprise; very few modern languages work like this.
+
+## Unresolved Questions
+
+1. Do any functions ship with Cedar? If so, which ones? Leaving that out of this RFC and proposing it's handled equivalently to [RFC 58](https://github.com/cedar-policy/rfcs/pull/58)
+2. Can you import functions? Leaving that for a future RFC.
```
mwhicks1, line 295: Agree. Questions of management of functions, functions in namespaces, etc. are separable, just as management of type definitions from schemas for a standard library are separable.

text/0061-functions.md
aaronjeline: Unless we add first class functions, or some kind of late-binding mechanism

mwhicks1: What do you mean by "late binding" mechanism? In any case, we have neither of these so no need to discuss until we are actually proposing to have them (we can always relax the parser later, if we add them).

text/0059-functions.md
aaronjeline: > execution of even a single policy if the problem is not solved.
There are solutions to this. In the absolute absense of being able to load a function context they can be inlined. (Esp if we take the CBN semantics below, which I agree with)

text/0061-functions.md
aaronjeline: Thought about this some and I agree completely, going to update the RFC

cdisselkoen: Per Mike's comment, with macros, we could also relax the restriction that the body cannot reference Cedar variables.

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+};
+permit(principal,action,resource) when {
+    foo(1, "hello") // Run time type error
+};
+permit(principal,action,resource) when {
+    foo(1, "hello", principal) // Parse error
+};
+permit(principal,action,resource) when {
+    bar(1, "hello", principal) // Parse error
+};
```
cdisselkoen, line 139: This implies that the parser is stateful in a way it is not currently today.  In order to determine whether these examples are parse errors, the parser has to refer to information it has already parsed earlier in the file (the definition of `foo`).  I'm not sure this is the design we want.

An alternative would be that calling a nonexistent function, or calling with the wrong number of arguments, is not a parse error but an evaluation error, and also detected by validation of course.

Edit: With the below stipulation that function declarations do not have to lexically precede use, this is even more problematic.

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+function mydecimalLTE(d,e) {
+  d <= e    // d and e are just Long numbers
+}
+```
+These functions basically implement the equivalent of Cedar `decimal` numbers. But the approach has at least two drawbacks.
+
+First, if `i` and/or `f` are outside the allowed range, you will get a Cedar overflow exception at run-time. This exception is not as illuminating as the custom error emitted by the `decimal` extension function (whose message will be `"Too many digits"`).
+Moreover, custom errors from constructor parameter validity checks can be emitted during validation when using extension functions, but not when using Cedar functions.
+
+Second, there is no special parsing for Cedar functions. With Cedar's built-in `decimal`, you can write `decimal("123.12")` which more directly conveys the number being represented than does `mydecimal(123,12)`.
```
cdisselkoen, line 221: Worse, `mydecimal(123,12)` actually means `mydecimal(123,0012)`

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+
+<!-- ### Make parse errors runtime errors
+Make (any subset of) the following errors runtime errors instead of parse errors: 
+
+1. Application of non-function
+2. Non-application of function
+3. Function application with incorrect arity -->
+
+### Naming
+Should these really be called `function`s? They are actually `macro`s.
```
cdisselkoen, line 294: I'm not sure I agree that these are macros, in the current design.  They are pure functions.

If we want to avoid associations with other languages' functions or macros, we could also pick a third term, like `snippet`, `predicate`, `def`, ....

cdisselkoen (commented):
Generally approve of this RFC, but withholding my final approval pending resolution of some of the comments in other reviews

text/0061-functions.md
aaronjeline: Fine with `snippet` or `def`

text/0061-functions.md
aaronjeline: The argument against making them runtime errors is that that reifies functions. Under the current proposal, functions could be fully erased.

text/0059-functions.md
emina: Decimal is already solved by the `Decimal` extension type.  

As for Datetime, I'm having a hard time thinking of non-trivial Datetime functions that can be easily expressed in Cedar and for which this would be a good solution.  (Though that could be a lack of imagination on my part :)

For example, consider a function that takes as input a timestamp and calculates which day of the week this corresponds to.  This is a hairy complicated calculation and it would be very painful to write in Cedar (if it's even possible---I genuinely don't know if this algorithm can be expressed with loops that can be feasibly unrolled).

Also, inlining a function like this everywhere would not scale for analysis.  In fact, even if we don't inline them explicitly, that is effectively what the analysis will have to end up doing anyway.  (The encoding of the function's body will be inlined at each callsite---potentially with specific sequencing of argument evaluation if we want CBV.). So, while we can analyze Cedar functions in theory, our practical ability to do so will very much hinge on these functions between small and simple.

My other concern is that this proposal is a non-trivial breaking change to the language semantics and APIs.   We shouldn't underestimate just how pervasive the changes must be to implement this as proposed.

For example, to evaluate policies, the user now has to pass in an environment that binds function names to their implementations (at least as described in this RFC).  So, in addition to policy slicing and entity slicing, users may have to potentially do function slicing (i.e., figure out which functions to pass in if they have thousands of them scattered in various places).

Finally, this RFC is going to make our formalization and proofs significantly more complicated---especially deep semantic proofs that have to reason about termination of evaluation, typechecking, and symbolic compilation in the presence of functions.  We should not underestimate the proof engineering difficulty that this brings. 

In my view, we are yet to see a sufficiently compelling use case to justify adding this level of complexity to the language, formalization, analysis, and implementation.

aaronjeline: > So, while we can analyze Cedar functions in theory, our practical ability to do so will very much hinge on these functions between small and simple.
Isn't this also the case for cedar policies as they exist now? If people aren't using functions now and just writing everything out manually it's the same thing. The only argument is that writing functions makes it easier (and therefor more attractive) to write bigger policies, and given the fact that functions can't call other functions that seems weaker.

> Finally, this RFC is going to make our formalization and proofs significantly more complicated---especially deep semantic proofs that have to reason about termination of evaluation, typechecking, and symbolic compilation in the presence of functions. We should not underestimate the proof engineering difficulty that this brings.

Is this true if you inline things? I don't see why it would be.

aaronjeline: > Decimal is already solved by the Decimal extension type.
Yes but we've talked before about how we don't love that it's an extension since it's expressible in pure Cedar. my point was that _would_ have solved this need without needing an extension.

emina: My understanding of the proposal is that we want to support actual functions (lambdas) here, which involves passing in an environment that binds function names to bodies and evaluating all this on the fly---as well as adjusting the semantics, formalization, etc., accordingly.  

If we aren't proposing functions, and we are proposing macros, then formalization etc don't have to change.

But that's a different proposal and we have to worry about different things (hygiene etc).

aaronjeline: Yup, these are basically extremely restricted macros. No need for bindings on the fly. Hygiene is trivial since macros can't invoke other macros, and policies only bind `PARC`

emina: I think it's worth revising the proposal to make this clear :). 

I would also clarify that the formal semantics, AST, etc. don't actually change in an any way with this proposal.  If these are macros, then they desugar to core Cedar via a pre-processing pass, and we don't need a new evaluation rule for them:

https://github.com/cedar-policy/rfcs/blob/aaronjeline/functions/text/0061-functions.md#formal-semantics

Instead, we provide expansion / desugaring rules that are separate from the formal semantics of the language.

aaronjeline: Good callout

emina: (And we can separately prove facts about desugaring.)

aaronjeline: These macros can't create policies, whereas templates can. Coming up with a unified macro for both levels of the language sound annoying, but maybe im just not seeing it.

mwhicks1: Suppose we adjust this RFC's syntax to be
```
def semver(?major, ?minor, ?long) {
  { major : ?major, minor : ?minor, patch : ?patch }
};
```
That is
- We change `function` to `def`
- We require that macro parameter variables are prefixed with a `?`

I think these changes convey the nature of what we are providing as a macro: we `def` a macro, and `?`-prefixed parameters could be expressions, not just values. The change to variable names also resolves the thorny issue around name shadowing -- we don't have to worry about a parameter shadowing the `principal` variable for example, which is "global." With this change, we can allow macros to freely reference it.

I also think this syntax opens the door to more generalized templates. The `?var` parameters are the same in both. Future syntax for generalized templates could be like the following (which evokes one of the alternatives of [RFC 3](https://github.com/cedar-policy/rfcs/blob/placeholders-in-conditions/text/0003-placeholders-in-conditions.md), which was rejected until more was known):
```
def template cross_relation(?user, ?resourcebound)
  when {
       ?resourcebound in resource
    && principal in ?user.delegates
  };
```
I'm not saying we include templates in this RFC; I'm saying that taking this RFC could gel nicely with a future generalized templates RFC.

mwhicks1: And while I'm at it, let me suggest a different option for the syntax: Drop the `{` `}` around a macro definition. So we would have:
```
def semver(?major, ?minor, ?long)
  { major : ?major, minor : ?minor, patch : ?patch }
;
```
rather than
```
def semver(?major, ?minor, ?long) {
  { major : ?major, minor : ?minor, patch : ?patch }
};
```
Arguing in favor
1. Cedar expressions don't have curly braces around them, and dropping curly braces around the definition makes returning a record less ugly/error prone, e.g., `def semver(...) { ... };` instead of `define semver(...) { { ... } };`
2. Short macros will look better without the braces, e.g., `def ismanager(?e1,?e2) ?e1.manager == ?e2` vs. `def ismanager(?e1,?e2) { ?e1.manager == ?e2 }`
3. The use of `;` as a terminator looks more natural after an expression than it does after a closing curly brace of a top-level definition, i.e., `};` -- no other language does this, that I know of.

Arguing against:
1. The curly braces, while confusable with record literal syntax, may make the macro body easier to visually isolate. I can't think of better delimiters, but maybe you could use a separator instead. For example, if you use `:=` we'd have `def ismanager(?e1,?e2) := ?e1.manager == ?e2;` but I worry this is not easier to read and is no less confusing

Another orthogonal syntax idea is to change `def` to `define`, to evoke C-style macros, which developers will be familiar with. I don't like this as much though because (a) C-style macros need to be all on one line, which Cedar macros don't; (b) C-style macros don't have a special `?var` syntax for parameters; and (c) they don't enforce hygiene. Also, I like that `def` is sort of like `define`, but also is used to define functions in Ruby and Python. Our proposed macros are basically call-by-name functions, so this is not a terrible analogy to make.

Considering other language syntaxes for inspiration
- [Scala has macros](https://docs.scala-lang.org/overviews/macros/overview.html), and they use `def` as the keyword to introduce them, as with normal functions, but then add keyword `macro` prior to the body. So something like `def ismanager(?e1,?e2) macro ?e1.manager == ?e2;`. Looks a little odd but it could grow on me.
- [Scheme/LISP macros](https://en.wikibooks.org/wiki/Scheme_Programming/Macros) are the benchmark, but the S-expression syntax doesn't carry over. They use keyword `defmacro` or `define-syntax`.
- [Julia has macros](https://docs.julialang.org/en/v1/manual/metaprogramming/#man-macros) but it's not a good stylistic fit (and few people will know it)
- [Rust has macros](https://doc.rust-lang.org/book/ch19-06-macros.html) but again the syntax is not a great stylistic fit (e.g., use of `$` and `!` in the def, and `#` in annotations), and they are much more powerful

aaronjeline: Updated the RFC. Removed the section on var-vs-function namespaces since it's now moot

text/0061-functions.md
```diff
@@ -0,0 +1,295 @@
...
+
+This RFC adds a new top level form to Cedar policysets: the function declaration.
+
+A Cedar function declaration is composed of three elements:
+
+1. A name, which is a valid (possibly namespaced) identifier.
+2. A list of parameters, each of which is a non-namespaced identifier.
+3. A body, which is a Cedar expression.
+    1. The body may contain (non-namespaced) variables, drawn from the function's parameters.
+    2. The body may not contain calls to other functions.
```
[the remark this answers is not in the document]
mwhicks1: They _can_ contain these variables now that we have renamed function parameters to start with `?`.

And: I don't think we should call these functions, we should call them macros. More on this in response to a later comment.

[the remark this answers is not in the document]
aaronjeline: I still don't think they should, violates macro hygiene and could lead to confusion

text/0061-functions.md
mwhicks1: Can you not do the checking in two steps? Basically:
1. While processing an input file, store all function defs in a global table that logs name, args, body. For each function application, also consult the table for consistency. If the function is not yet in the table, then add it with the arity and "none" for the body. If a def comes after a call, check it for consistency and replace "none".
2. Make sure no functions in the table have "none" for the body.

Doesn't sound hard?

text/0061-functions.md
mwhicks1: Boom! Good catch.

text/0061-functions.md
aaronjeline: I don't think it's hard, I think Craig's point is that until now we have not had to deal with any kind of use-def relationships in the parser, and this RFC changes that. If we make it a runtime error, our parser stays the way it is.
I think it's worth the change, and I think the RFC doesn't meet our needs if it's a runtime error, but the callout is correct.

text/0061-functions.md
mwhicks1: I would claim these are best _communicated_ as macros, for two reasons:

1. Most readers are used to CBV languages. So they inherently assume (having not actually read the docs) that for _functions_ when you write `f(1+2,principal.id like "foo")` then you will evaluate `1+2` and then `principal.id like "foo"`, and then call the function with the results. They will not imagine that inlining will happen, and that short-circuiting and whatnot can change the results.
2. Readers are familiar with macros in C, and perhaps other languages. For _macros_ they know that when you write `f(1+2,principal.id like "foo")` you are substituting full expressions `1+2` and `principal.id like "foo"` into the body, you are not evaluating the them first. 

So I claim that by calling them macros we help users naturally go to the right semantics in their heads. Yes, it's true that we are really (or equivalently) providing is call-by-name functions, but calling them macros is more illuminating, IMO.

In sum: I propose changing the whole RFC to be "Cedar Function Macros" and then use "macro" for short, throughout.

aaronjeline: Writing up the function vs macro discussion in the RFC...

text/0061-functions.md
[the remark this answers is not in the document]
mwhicks1: It doesn't violate macro hygeine. `principal` etc. are not macro variables.

text/0061-functions.md
khieta: I also prefer `macro`. To me `function` suggests that I should be able to nest calls, whereas I can easily accept that a macro shouldn't call another macro. I'm not a fan of `predicate` since that suggests that the resulting expression must be boolean-typed. No preference on `snippet`, except that I haven't seen it used before so it feels non-standard.

Also, not a fan of "Cedar Function Macros". I found this terminology confusing on my initial read through -- just pick function or macro.

text/0061-functions.md
```diff
@@ -0,0 +1,337 @@
...
+when {
+    resource.apiVersion.major > 2 ||
+    (resource.apiVersion.major == 2 && resource.apiVersion.minor >= 1)
+};
+```
+
+Here it is instead represented with functions
+
+```
+def semver(?major, ?minor, ?long)
```
khieta, line 36:
```suggestion
def semver(?major, ?minor, ?patch) 
```

text/0061-functions.md
```diff
@@ -0,0 +1,337 @@
...
+};
+```
+The above won't work with CBV because the second expression to the call to `implies` will fail if `principal.attr` does not exist.
+
+### Let functions call other functions
+As long as cycles are forbidden and functions as arguments are disallowed, we could allow functions to call other functions without sacrificing termination.
+However, the potential complexity explosion is high, and it's backwards compatible to add this later.
+
+### Naming
+Should these really be called `function`s? They are essentially macros. Call them that?
```
khieta, line 332: Redundant with `Naming: Are these macros or are these functions?` section above.

text/0061-functions.md
```diff
@@ -0,0 +1,337 @@
...
+    2. Users may have data structures that are totally bespoke to their systems. It makes no sense to include these in the standard Cedar distribution at all, yet users may still want some way to build abstractions. 
+2. They are too powerful. Extensions are implemented via arbitrary Rust code, which is essential for encoding features that cannot be represented via Cedar expressions (such as IP Addresses), but opens the door for a wide range of bugs/design issues. It’s trivial to design an extension that is difficult to validate and/or logically encode for analysis. Problematically, extension functions can potentially exhibit non-determinism, non-termination, or non-linear performance; interact with the operating system; or violate memory safety. This raises the code review burden when considering an extension function's implementation.
```
khieta, line 67: This is a great summary of the current state of extension functions. This is why I'm in favor of this RFC: there is currently no way to build abstractions in Cedar outside of (built-in) extension functions.

So although I prefer "macro" to "function" for the feature in this RFC, I'd like to throw another name into the ring: _User-defined extension functions_

(Note that this would make the most sense if we used CBV for both types of calls.)

text/0061-functions.md
```diff
@@ -0,0 +1,337 @@
...
+
+In contrast, function macros written as simple abstractions over Cedar expressions, which themselves cannot call other Cedar functions, have none of these problems. They naturally inherit the properties of Cedar expressions. Cedar functions are guaranteed to terminate and be deterministic. Since they are compositions of Cedar expressions, it’s easy to validate and analyze them.
+
+## Detailed Design
+
+### Function Macro declarations
+
+This RFC adds a new top level form to Cedar policysets: the function macro declaration.
```
khieta, line 75:
```suggestion
This RFC adds a new top level form to Cedar policy sets: the function macro declaration.
```

text/0061-functions.md
```diff
@@ -0,0 +1,337 @@
...
+
+Structurally, a declaration is written like this:
+
+```
+def name(?param1, ?param2) body ;
+```
+Use of an unbound variable in the body is a syntax error.
+A parameter list may not declare the same variable twice, and may not list any standard Cedar variables.
+An unused variable is a syntax warning.
+Function and variable names share the same namesapce, with standard lexical scoping.
```
khieta, line 96:
```suggestion
Function and variable names share the same namespace, with standard lexical scoping.
```

text/0061-functions.md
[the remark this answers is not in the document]
aaronjeline: But your referencing a variable not bound in the macro body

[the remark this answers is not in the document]
aaronjeline: From Hygienic Macro Expansion
> Hygiene Condition for Macro Examples:
Generated identifiers that become binding instances in the completely expanded program must only bind variables that are generated at the same transcription step.

In this: bindings generated in the first phase (the syntax of policies) are bound in the second phase (macro-expansion) 

Why we care:
It's backwards compatible to add the ability to directly reference principal/action/resource. If we end up seeing examples that would be a lot clearer with this we can add it. On the other hand, if we expand the language at any point to include other binding forms, having strict hygiene will become much more important, and it will be backwards incompatible to remove direct PAR references.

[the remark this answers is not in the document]
mwhicks1: I don't think I agree that this is a violation of hygiene. These are completely different sorts of variables, and they are global. There is no risk of confusion/capture/etc. of variables bound/applied in the macro.

But maybe it's an issue of clarity generally. Would we be Ok with
```
def isOwner() principal == resource.owner;
permit(principal,action,resource) when { isOwner() };
```
or do we really want
```
def isOwner(?principal,?resource) ?principal == ?resource.owner;
permit(principal,action,resource) when { isOwner(principal,resource) };
```
I can imagine preferring the first example so that you cannot mistakenly swap the arguments, e.g.,
```
def isOwner(?principal,?resource) ?principal == ?resource.owner;
permit(principal,action,resource) when { isOwner(resource,principal) };
```

text/0061-functions.md
mwhicks1: > User-defined extension functions

I prefer not to call them functions because they are not CBV.

text/0061-functions.md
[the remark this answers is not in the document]
aaronjeline: Hmm, I prefer the first one, but I've been thinking about it and I can't come up with a good reason why, so maybe it's just taste.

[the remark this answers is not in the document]
mwhicks1: I'm lukewarm on it, to be honest. It's more future-proof to forbid the use of `principal`, `action`, etc. within macro bodies, so that might be the best way to proceed.

It occurs to me that you might want to add a section about interaction with templates. Suggested text:

Cedar macros use the same notation for their variables as do templates, since in both cases variables are "holes" that are filled in with a Cedar expression to make a full construct -- for templates, the construct is a policy, for macros it is an expression.

Templates can include calls to macros. For example:
```
def semver(?major, ?minor, ?patch) ... // same as initial example
def semverGT(?lhs, ?rhs) ... // same as initial example
permit (
  principal == ?principal,
  action == Action::"view",
  resource in ?resource)
when {
  semverGT(resource.apiVersion, semver(2,1,0))
};
```
When we link this template, we get a full policy, as usual.

A macro parameter can end up having the same name as a template variable, with the effect that it shadows it, which is expected with normal lexical scoping: 
```
def isOwner(?principal,?resource) ?principal == ?resource.owner;
permit(principal,action,resource in ?resource) 
when { isOwner(principal,resource) };
```
Here, notice that the `isOwner` macro's reference to `?resource` is to its parameter, not to the template slot. (We'd recommend to users to choose different parameter names to avoid confusion.)

As already mentioned, you cannot refer to a parameter in a macro that it does not bind, which enforces hygiene. So the following is not allowed.
```
def isOwner() ?principal == ?resource.owner; 
// ERROR: cannot refer to unbound variables ?principal, ?resource
```

text/0061-functions.md
mwhicks1: And FTR, `mydecimal(123,12)` represents the number 123.0012, which you didn't quite write above.

mwhicks1:
I've taken a pass over the whole RFC, bringing the terminology and presentation up to date with all the various discussion threads. I'm approving the current version.

mwhicks1 (approved):

emina:
Thank you @aaronjeline and @mwhicks1 for writing this RFC!

I appreciate that the current version of the proposal is simpler than the original.  In particular, by proposing macros with call-by-name semantics rather than functions with call-by-value semantics, we avoid complicating the language semantics, validation, and analysis.

Despite this, I'm not in favor of adding macros to the language at this time.  The pros don’t outweigh the cons (yet) in my opinion, and this is a major addition to the language that we cannot take back.  Macros will create a management problem for services, and they will introduce two significant sources of confusion and complexity into policies:  

1. **Bad error messages due to inlining**.  Consider this example:

```
def foo(?a, ?b, ?c) ?c + (?a + ?b);

permit(principal,action,resource) when {
    foo(0, 9223372036854775807,  2) // overflow error
};
```

When this policy errors at runtime, what source location should be shown as the source of the error?   Should it be the location in the macro body `(?a + ?b)`?  Should it be _just_ invocation of the macro in the policy?  Some combination of the two?  The same problem applies to validation errors too.  Without good error messages, errors like this are notoriously hard to understand and debug.  And providing good error messages involves tracking source locations through macro expansions and retaining this information in this AST, and potentially EST too.   All of this is non-trivial to implement (even for the restrict macro system proposed here), and it will require non-trivial (potentially breaking) changes to the EST.

2. **Bad performance due to hidden costs.**. Right now, Cedar enjoys the nice property that the runtime cost of a policy is evident from the policy text:  it is worst-case quadratic in the size of policy text and input.  This will no longer be the case with macros. Users will need to think about the runtime cost of their policies _after_ macro expansion.  And macros make it possible to write a policy whose actual expanded size is exponential in the size of the text written by the user. 

As a pathological example, consider this macro and its application:

This policy 

 (2) the ability to write small policies that expand into policies that are exponentially larger than the text the user wrote.   
These problems aren’t present in Cedar right now because each policy is self-contained, and its meaning and cost are evident from the text.  That’s the advantage of the lack of abstraction.  There are downsides, of course, as outlined in this RFC.  In my view, we are yet to see compelling evidence that abstraction provides enough benefits in a very restricted authorization DSL like Cedar to justify the added complexity.  For that reason, I’m voting against accepting the RFC at this time.

mwhicks1:
Thanks for providing this carefully thought-out critique! I am still optimistic that macros are a net win. Here are some responses.

> Bad error messages due to inlining.

This seems like a surmountable problem. I presume there is much we can learn from far more interesting macro systems' handling of errors. Off the top of my head:

For run-time evaluation, you don't have to in-line the call in advance, you can in-line by creating the expression on the fly, while evaluating, basically resulting in a call-by-name function call. The in-lined body can include source locations that reflect the macro body in the right places, and the macro args in the others. The error message can include a stack trace, which a user might expect. It should be possible to prove (or test) that that doing this is equivalent to in-advance in-lining. For validation, I think you can do a similar thing. 

For Lean models, obviously you don't need to track the source locations. You don't even need to do on-the-fly inlining, at least at first, by relying on DRT against the Rust, and PBT within the Rust, to show equivalence. We can prove equivalence in Lean later, if we wish.

> Bad performance due to hidden costs.

This is a fair point, and I worried about this, too. Here are some mitigating factors that come to mind:

- Services are welcome to impose bounds based on the in-lined policy size, rather than the concrete policy size. Users can be made aware that this is the case, in which case both users and services benefit: Policies are more compact when stored (and read), even if they are larger when evaluated. We could imagine warning users, with a policy linter, if they stray into pathological territory.
- In-lining on the fly, during evaluation, should help mitigate maximum in-memory size, in some cases at least.

Ultimately, I feel that the readability and reusability benefits of macros are significant; the increase in complexity to support them is small; and that we are failing to solve the problem posed by today's extension functions if we do not provide them. Yes, we cannot stop users from using them in unwise ways, but that's true in general. We can guide them with linting, documentation, etc. toward good solutions.

emina:
Can you say more about how on-the-fly macro expansion would work?  

The way I imagine it would work means we’d have to break the `is_authorized` API (among other breaking changes).  So, `is_authorized` would have to take an extra environment argument, which is a binding from macro names to their bodies. The evaluator would use this environment to inline macro arguments into macro bodies on the fly.  (Actually, I think the evaluator might have to expand the arguments before inlining them, for hygiene.)

In addition to breaking the API, we’d also have to extend the AST (and possibly break the EST), at least in Rust.

Doing macro expansion statically, prior to evaluation, avoids the above problems, but has the issue that constructing useful error messages becomes hard.

As for Lean, formalizing function calls (CBV or CBN) is possible but tricky.  For example, it will be non-trivial to prove termination (because the size of the AST won’t be usable as a natural decreasing measure during recursive calls). If we want to formalize this, we’ll definitely want to keep the alternative formalization separate from the core semantics / validation / analysis formalization, because we don’t want to handle on-the-fly inlining in these more complicated components too. 

So, we would need an additional concrete semantics for Cedar, with macros and CBN semantics, and a proof that this alternative semantics is equivalent to our existing semantics post-expansion. 

This all adds a hefty engineering and maintenance burden.  My concern continues to be that this burden as well as the potential confusion / usability / performance issues are not offset by the upsides of macros for Cedar, at least from I’ve seen so far.  

How to weigh these tradeoffs is, of course, subjective, and I realize that mine is a minority opinion :)

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+```
+
+For simplicity, safety, and readability, Cedar macros cannot call other macros, and cannot take macros as arguments.
+
+## Motivation
+
+Cedar currently lacks adequate mechanisms for users to build abstractions. The only existing mechanism is extension functions, which are limited for two reasons:
+
+1. They are extremely heavyweight, requiring modifications to the Cedar evaluator. This means that users who want to stay on official versions of Cedar have no choice but to attempt to submit a PR and get it accepted into the mainline. This process does not scale.
+    1. For data structures that are relatively standard (ex: SemVer, or OID Users as proposed in [RFC 58](https://github.com/cedar-policy/rfcs/blob/cdisselkoen/standard-library/text/0058-standard-library.md)), it’s hard for the Cedar maintainers to know what’s in-demand enough to be included, and how to balance that against avoiding bloat. There’s no way to naturally observe usage because the only way to “install” the extension pre-acceptance is to vend a modified version of Cedar.
```
mtrottie, line 64: nit: OIDC

aaronjeline:
> The way I imagine it would work means we’d have to break the is_authorized API (among other breaking changes). So, is_authorized would have to take an extra environment argument, which is a binding from macro names to their bodies. The evaluator would use this environment to inline macro arguments into macro bodies on the fly. (Actually, I think the evaluator might have to expand the arguments before inlining them, for hygiene.)

I'm not sure there would be much API breakage. The environment argument could be wholly contained within the `PolicySet` object, which would make sense as that's already the result of parsing. Beyond that it seems like everything else is just an expansion of the API. 

The error messages problem seems like the most important to solve. As far as the problem around where the source is located, services already have to solve this is the hold cedar policies as separate entities instead of one blob of a policyset.

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+```
+// Permit access when the api is greater than 2.1
+permit (principal, action == Action::"someAction", resource)
+when {
+    resource.apiVersion.major > 2 ||
+    (resource.apiVersion.major == 2 && resource.apiVersion.minor >= 1)
+};
+```
+This policy does accomplish the user's goal of encoding the SemVer relationship.
+The problem is readability. A reader of this policy has to work out that it is implementing the standard semantic version comparison operator, and not some bespoke versioning scheme. Another problem is maintainability. If you have multiple policies that reason about SemVers, you have to repeat the logic, inline, in each policy. This violates basic software engineering tenets (Don’t Repeat Yourself), allowing bugs to sneak in via typos or botched copy/paste.
```
mtrottie, line 238: While this example is a bit more verbose it is serializable / deserializable across different cedar applications. How do we store these policies with macros? I see the management becoming a bit more complicated.

emina: The management burden shouldn't be underestimated.

As far as I see it, there are two options:

(1)  The service / application stores a representation of the policy with the macro calls.  Then, the caller is responsible for gathering all the necessary macro definitions and passing them to the authorizer, just as they are currently responsible for gathering all the relevant policies and entities and passing them to the authorizer.  This same burden extends to services that may need to ship policies and macros to a local agent.  

(2) The services / application stores a representation of the expanded policy---that is, all the macros are inlined before storing the policy.  This is how macros are handled in mature languages with macro systems. And it means that macros no longer need to be shipped.  However, it also becomes impossible to restore the original policy text from the expanded representation, and if there is a bug in the macro that is propagated to all policies that use the macro, it cannot be fixed by just fixing the macro:  the macro must be fixed and all policies that use it must be re-expanded.

aaronjeline:
> > 5\. Are type annotations required or optional?
> 
> I'd make them required at the beginning. We can relax the rule in the future.

This raises two issues:
1. This feels weird if you aren't using a schema at all
2. This gives you no way to write polymorphic functions unless we extend the type annotations with parametric polymoprhism (which seems like a log)

max2me:
> > > 5. Are type annotations required or optional?
> > 
> > 
> > I'd make them required at the beginning. We can relax the rule in the future.
> 
> This raises two issues:
> 
> 1. This feels weird if you aren't using a schema at all
> 2. This gives you no way to write polymorphic functions unless we extend the type annotations with parametric polymoprhism (which seems like a log)

Except almost everyone should be using schemas. So.. :)

mwhicks1:
Comments from Dan Wang (who doesn't have a GitHub account), plus my responses:

> I would consider not calling this concept a macro or using the `def` syntax. Theorem provers have the concept of [abbreviation](https://twelf.org/wiki/abbrev-declaration/#_top). For which the proposal smells more like. Lean has a similar concept. I find the call by name conversation a bit confusing to. To me this just "expand the parse tree".

Abbreviations in Twelf, per the link, look like type abbreviations, not term-level functions. They don't seem to have parameters. In Lean, an ["abbreviation"](https://leanprover-community.github.io/glossary.html#unicode-abbreviation) seems to be a non-unicode way of specifying a unicode character.

It is call-by-name in the sense that you can really do the calls on the fly, rather than all in advance, as an evaluation strategy.

> The fact these cannot refer to other macros is a bit confusing. There very second class nature, make me want to make their invocation standout with a similar to the rusts "!" convention.

> The `double(double(..)` example shows that even without non-recursive-self-reference you have blowup, so why not allow macros to refer to other macros?
> 1. consider some name other than macro.
> 2. make the call of these things standout (aka don't treat them like normal function calls.. because they are not)
> 3. remove self-reference restriction to make them more powerful.

It's a fair point that even without self-calls there is an opportunity for bad behavior, and thus we might relax that. Our thinking was that we can add internal calls later, if customers demand it. But maybe we should consider it now.

> I assume a debugging aid there will be a mode to see the fully expanded view of the code

That seems reasonable.

emina:
> Abbreviations in Twelf, per the link, look like type abbreviations, not term-level functions. They don't seem to have parameters. In Lean, an ["abbreviation"](https://leanprover-community.github.io/glossary.html#unicode-abbreviation) seems to be a non-unicode way of specifying a unicode character.

Lean `abbrev` is more powerful than that.  It can be used to define type abbreviations but it can also be used to define function-like macros.  Something like:

```
abbrev foo (v : Nat) : Nat := 1 + v
```

philhassey:
Hi, I did some thinking here with a few folks, just some observations:

1. This would begin a journey into having a fully featured Cedar "programming" language.  I could imagine it would be very useful to be able to define constants and even types with macro-methods on them.  At that point, I'd also want to be able to define packages, so I wouldn't always be populating a global namespace anytime I used these features.

2. Related to 1, a Cedar policy is currently just a list of policy statements.  This change would be a notable departure from that, as now a Cedar policy would be not only policy statements, but whatever macros, etc, are also in the document.  So a tradeoff is made in terms of understandability - you have to understand more language features, but your policy statements could become more terse.

3. I could also see accomplishing all the same stuff through a non-Cedar templating language of choice.  This would keep the core Cedar be just policy, and leave the meta-templating outside the scope of Cedar.  The drawback here, being you'd have to learn Cedar and whatever your templating language is.

In any case, I really liked some of the comments / notes in the RFC such as:
- Wanting to keep performance provable
- Wanting to avoid user-made extensions because they could have non-deterministic results which breaks the ability to re-play a decision
- The use of the word macro seems better than function for what these are

In the Go implementation, I think adding user-made extensions wouldn't be terribly difficult, but the concerns about breaking the analyzability / determinism give me pause to think.

200sc:
Similar concerns are already present in comments, but:
1. Is this problem not adequately solved by external templating today? 
2. Is this the right part of the cedar ecosystem to insert macros? 
2a. Given this introduces the ability to define the same macro twice, presumably causing an error, then without a packaging system this makes macros unusable once you reach a large enough mass of cedar policy. Schemas already exist to contain these kinds of unique identifiers. 
3. Philosophically is cedar the place where users spend their time and define their logic (constants, packages and scopes, functions) or is cedar a data format for the result of higher-level language? Currently Cedar is trivial to understand because everything can be interpreted in the same way: forbid or permit according to these rules. From an auditing perspective cedar is thus an ideal, if verbose, format for expressing exactly what is and is not allowed. Adding macros within cedar itself adds a layer of indirection to cedar's auditability; if Cedar is eventually going to become something like a Turing complete programming language, that seems appropriate, but I'd appreciate an intentional decision stating that intended direction.

emina:
Thank you @philhassey and @200sc for the comments!  

My take is along the same lines:  I want _all_ the abstraction in a general purpose language :) but not necessarily in a DSL like Cedar, for the reasons mentioned in the discussion and in your comments.

To answer the last point raised by @200sc:  Cedar won't become Turing complete.  Analyzability is a key tenet of the design.  

It is worth noting that macros won't compromise logical analyzability---that is, the ability to mechanically reason about policy behavior by reducing its semantics to a logical formula.  But they can hurt understandability as much as they help, as all abstractions do.

In my view, the upsides of abstraction (terseness and expression reuse) are not worth the added complexity for policy writers, readers, services and applications, and implementors of Cedar.

philhassey:
Hi @mwhicks1 - just to clarify my statement:

> I could imagine it would be very useful to be able to define constants and even types with macro-methods on them. At that point, I'd also want to be able to define packages, so I wouldn't always be populating a global namespace anytime I used these features.

I'm not objecting in general, but I am saying "if we do this, we'll want a few more things".   Which relates well to your comment:

>But I don’t feel that we need a package system before introducing macros. It can be a fast-follow.

I feel the same way about adding in constants, and types with macro-methods.  At which point Cedar would be a much richer language.  I do think it's reasonable at this point to consider if that's where we want to end up though.  I'm not sure.

text/0061-functions.md
andrewmwells-amazon: I'm happy with `macro`. Maybe `inline function` would be another option (though it still may suggest CBV to some readers)?

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+```
+  if principal has attr then
+     resource has attr && principal.attr == resource.attr
+  else true
+```
+Notice how the argument expression `principal has attr` has been substituted for parameter `?e1` whole-cloth, without evaluating it first (and likewise for the other parameter/argument). This means that if `principal` indeed has optional attribute `attr`, then evaluating sub-expression `principal.attr` in the `then` clause is safe. If we eagerly evaluated a macro call's argument expressions then `principal.attr` would be evaluted eagerly, producing an error if `principal has attr` turns out to be `false`.
+
+### Errors
+
+Macros do not exist at run-time -- they cannot be stored in entity attributes, requests, etc. As such, referencing a function by its name, without calling it, is a syntax error.
```
andrewmwells-amazon, line 124:
```suggestion
Macros do not exist at run-time -- they cannot be stored in entity attributes, requests, etc. As such, referencing a macro by its name, without calling it, is a syntax error.
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+when {
+    resource.apiVersion.major > 2 ||
+    (resource.apiVersion.major == 2 && resource.apiVersion.minor >= 1)
+};
+```
+This policy does accomplish the user's goal of encoding the SemVer relationship.
+The problem is readability. A reader of this policy has to work out that it is implementing the standard semantic version comparison operator, and not some bespoke versioning scheme. Another problem is maintainability. If you have multiple policies that reason about SemVers, you have to repeat the logic, inline, in each policy. This violates basic software engineering tenets (Don’t Repeat Yourself), allowing bugs to sneak in via typos or botched copy/paste.
+
+### No custom parsing or error-handling
+Extension functions provide more full-featured constructors through custom parsing and error handling, but Cedar functions provide no such facilities. This may make them harder to read, write, and understand.
```
andrewmwells-amazon, line 241:
```suggestion
Extension functions provide more full-featured constructors through custom parsing and error handling, but Cedar macros provide no such facilities. This may make them harder to read, write, and understand.
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+
+For example, you could encode a decimal number as a `Long`, and then make Cedar functions to construct and compare decimals:
```
andrewmwells-amazon, line 243:
```suggestion
For example, you could encode a decimal number as a `Long`, and then make Cedar macros to construct and compare decimals:
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+    ?i * 10000 + ?f    // will overflow if i too big
+  else
+    18446744073709551615 + 1    // fraction f too big: induce overflow
+;
+
+def mydecimalLTE(?d,?e) {
+  ?d <= ?e    // d and e are just Long numbers
+}
+```
+These functions basically implement the equivalent of Cedar `decimal` numbers. But the approach has at least two drawbacks.
```
andrewmwells-amazon, line 256:
```suggestion
These macros basically implement the equivalent of Cedar `decimal` numbers. But the approach has at least two drawbacks.
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+
+First, if `i` and/or `f` are outside the allowed range, you will get a Cedar overflow exception at run-time. This exception is not as illuminating as the custom error emitted by the `decimal` extension function (whose message will be `"Too many digits"`).
+Moreover, custom errors from constructor parameter validity checks can be emitted during validation when using extension functions, but not when using Cedar functions.
```
andrewmwells-amazon, line 259:
```suggestion
Moreover, custom errors from constructor parameter validity checks can be emitted during validation when using extension functions, but not when using Cedar macros.
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+
+Second, there is no special parsing for Cedar functions. With Cedar's built-in `decimal`, you can write `decimal("123.12")` which more directly conveys the number being represented than does `mydecimal(123,1200)`. (Note that `mydecimal(123,12)` represents the number 123.0012, which may surprised some readers!).
```
andrewmwells-amazon, line 261:
```suggestion
Second, there is no special parsing for Cedar macros. With Cedar's built-in `decimal`, you can write `decimal("123.12")` which more directly conveys the number being represented than does `mydecimal(123,1200)`. (Note that `mydecimal(123,12)` represents the number 123.0012, which may surprised some readers!).
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+
+Of course, these drawbacks do not necessarily speak against Cedar functions generally, but suggest that for suitably general use-cases (like decimal numbers!), an extension function might be warranted instead.
```
andrewmwells-amazon, line 263:
```suggestion
Of course, these drawbacks do not necessarily speak against Cedar macros generally, but suggest that for suitably general use-cases (like decimal numbers!), an extension function might be warranted instead.
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+
+One mitigation is to expose the hidden costs, when needed. For example:
+- A policy authoring tool could reveal the post-expansion policy size to users, compared to the pre-expansion size, and could warn users in pathological cases like the above.
+- Services storing user-provided Cedar policies that wish to protect themselves can impose bounds based on the post-expansion policy size (shared wth users), rather than the concrete policy size.
+
+Ultimately, this RFC takes the position that it is better to give the tool of macros to users to improve the readability and maintainability policies, than it is to withhold that tool for fear that they could misuse it. Without the tool of macros, users must "implement" their effects by cut-and-paste, which still blows up policy size while also making policies harder to read and maintain.
+
+### Readability: Policies are no longer standalone
+A policy can no longer be read by itself, it has to be read in the context of all function definitions it uses.
```
andrewmwells-amazon, line 287:
```suggestion
A policy can no longer be read by itself, it has to be read in the context of all macro definitions it uses.
```

aaronjeline:
Responding to various comments:
From @philhassey 
> In the Go implementation, I think adding user-made extensions wouldn't be terribly difficult, but the concerns about breaking the analyzability / determinism give me pause to think.

Even worse, it's another step where we need to make sure we have the exact same behavior between the Rust and the Go, since they won't be able to share this code easily. While this is also true of the macro system, it's not true of each individual macro. 

From @200sc 
> Is this problem not adequately solved by external templating today?
No, as A) template holes can only be `?principal` and `?resource`, and even if we relax that restriction, template holes can't take arguments.

> Given this introduces the ability to define the same macro twice, presumably causing an error, then without a packaging system this makes macros unusable once you reach a large enough mass of cedar policy. 
Not disagreeing with the general point, however the proposal as stated now does allow for namespacing your macros.

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+Policies that use a large number of functions may be hard to read.
```
andrewmwells-amazon, line 288:
```suggestion
Policies that use a large number of macros may be hard to read.
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+;
+```
+
+Using type annotations would allow macros to be typechecked independently of policies that use them, and would add checked documentation of intent.
+Doing so may also make it easier to provide clear validation error messages.
+
+But introducing type annotations for macros introduces several questions.
+
+1. Do we allow `type` declarations allowed in policy sets, or just in schemas?
+2. Are type annotations on functions enforced dynamically à la "contracts," or are they just ignored at runtime?
```
andrewmwells-amazon, line 376:
```suggestion
2. Are type annotations on macros enforced dynamically à la "contracts," or are they just ignored at runtime?
```

text/0061-functions.md
```diff
@@ -0,0 +1,412 @@
...
+These functions basically implement the equivalent of Cedar `decimal` numbers. But the approach has at least two drawbacks.
+
+First, if `i` and/or `f` are outside the allowed range, you will get a Cedar overflow exception at run-time. This exception is not as illuminating as the custom error emitted by the `decimal` extension function (whose message will be `"Too many digits"`).
+Moreover, custom errors from constructor parameter validity checks can be emitted during validation when using extension functions, but not when using Cedar functions.
+
+Second, there is no special parsing for Cedar functions. With Cedar's built-in `decimal`, you can write `decimal("123.12")` which more directly conveys the number being represented than does `mydecimal(123,1200)`. (Note that `mydecimal(123,12)` represents the number 123.0012, which may surprised some readers!).
+
+Of course, these drawbacks do not necessarily speak against Cedar functions generally, but suggest that for suitably general use-cases (like decimal numbers!), an extension function might be warranted instead.
+
+### Hidden performance costs
```
andrewmwells-amazon, line 265: To me, this (tied with the general principal to avoid redundant language features) are the most concerning parts of this RFC. Since there are reasonable mitigations (e.g., the vscode plugin can display the expanded policy) I think this is ok if there is a sufficient level of interest in macros  (let's say 5 independent users asking for the feature).

So far, I don't think we've reached this level. So I'm not opposed to this RFC, but I'd vote against doing it now.

D-McAdams:
Adding my two cents: I'm inclined to "no".

It puts a huge burden on services built around Cedar. To keep up with Cedar releases, these parties need to schedule a large engineering project for storing these fragments and utilizing them efficiently. If they don't, they fall behind the language treadmill. Or, we end up with a bunch of opt-in features for Cedar such that when someone says they are "Cedar compliant", nobody knows what that means until they read the matrix of opt-in features within the service documentation.

We also have audiences who are already concerned that Cedar may be growing too complex when their customers are trying to achieve simple things. Evolving Cedar toward something that starts looking like a full-fledged programming language is a slippery slope that runs a real risk of turning away these audiences.

I'm sympathetic to the goals, and agree with other comments that audiences could use their favorite standard templating language and preprocess the Cedar themselves. And, I wouldn't be adverse to vendors supporting this feature on their own timelines, as an extension. Or, even for Cedar to suggest a common approach to this.

I also noted this discussion began with a request for a "Semver" extension, which I'm also sympathetic too. Somewhat ironically, I'm wondering if it would have been faster to just implement "Semver"!

aaronjeline:
> to keep up with Cedar releases, these parties need to schedule a large engineering project for storing these fragments and utilizing them efficiently

This is a good point, and suggests what others are saying about needing a higher bar for requests from users

> I'm sympathetic to the goals, and agree with other comments that audiences could use their favorite standard templating language and preprocess the Cedar themselves. And, I wouldn't be adverse to vendors supporting this feature on their own timelines, as an extension

I disagree with this for two reasons:
1. Having each service implement their own templating feature means they won't be compatible, you can't easily move between them. If people come to expect a templating feature, then you still have to solve the all the service-oriented problems around storing/retrieiving functions. In the worst case, we evolve towards an ad-hoc, poorly specified standard instead of an intentionally thought out one. (Regardless of it's this RFC or another shot at it). It won't be able to integrate with features like Cedar namespaces or a potential future module system, so that will also be left for service owners to figure our ad-hoc.

2. Text based templates are fundamentally the wrong tool for working with Cedar policies. Cedar policies are not linear sequences of characters, they are trees. A generic templating tool such as the C preprocessor is not able to reason about the actual structures users are manipulating, just the text based shadow of it. In the current proposal for example: every macro invocation is guaranteed to produce a valid cedar expression. It's fairly trivial to extend the current proposal to allow for type checking over macro functions in isolation. These two things are impossible for the text-based case. If we worry about the cognitive complexity of reading policies with macro invocations (a fair concern!) then the overhead of reading policies with arbitrary text substitutions is going to be worse and easier to mess up.

>  for Cedar to suggest a common approach to this.
I think if we want to suggest a common approach to this, it should be something like this RFC.

text/0061-functions.md
cdisselkoen: I think this slightly overstates the management burden.  The service / application is already storing policies, so also storing macros is not that huge of a lift.  @aaronjeline also talked to engineers from at least one managed Cedar service in writing this RFC, and they did not have concerns about the management burden.

mwhicks1:
> I also noted this discussion began with a request for a "Semver" extension, which I'm also sympathetic too. Somewhat ironically, I'm wondering if it would have been faster to just implement "Semver"!

To address this last point: We didn't just implement SemVer because doing so has maintenance costs which don't scale well. Each extension function we add increases the testing, validation, analysis, etc. burden of core Cedar. Instead, we wanted to explore a single mechanism that addresses them in one fell swoop. This is more scalable on the core Cedar team, and gives more agency to Cedar users.

I am still bullish that what's proposed here is the right thing to do. But I also appreciate the words of caution. The right course is to gather more information about uses/needs, and to think about open questions like packaging/management of macro definitions (which still aren't solved by the current extension function mechanism, even if we thought that was the right approach instead). 

One thought: Could we add Cedar macros as an experimental feature, and let customers use it and report back?

cdisselkoen:
Chiming in here to substantially agree with @aaronjeline and @mwhicks1, particularly their most recent two comments each.  I think it would be a big mistake to encourage folks to use "standard" templating languages on top of Cedar, instead of building something like this RFC, which can enforce Cedar macro hygiene, syntax checking, etc etc.  To the extent that one of the concerns raised in comments here is that this feature increases the cognitive complexity of Cedar : using a general-purpose templating language on top of Cedar would be even worse in terms of cognitive complexity, and wouldn't integrate with Cedar validation and analysis, like the macros in this proposal would.

text/0061-functions.md
mtrottie: Thank you for the responses. In both cases it seems challenging as it assumes a policy is already written.

A common use case is the policy author != the policy evaluator. For example, a system administrator writes a policy in the browser via some typescript plugin. This policy is sent to some backend system in java that wants to validate it against a schema and store it in a persistent store. The actual pulling of the policy and evaluation is a different system (a rust system).

It isn't clear how I would support this cleanly with macros. Not this RFC needs to address those concerns but this is how I see Cedar being used in practice.

emina:
> wouldn't integrate with Cedar validation and analysis, like the macros in this proposal would.

Can you explain what you mean by integration? 

If an external templating system simply expands whatever the user writes into a Cedar policy, the analysis will handle the expanded policy fine and so will validation.

It’s worth noting that custom macros don’t provide any advantage for analysis, which would have to inline every invocation of the macro in order to capture its semantics at the invocation point.  (That’s how we handle extension functions too—the definition is inlined at the call site, after appropriate sequencing of argument evaluation to account for errors.)

In this proposal specifically, macros aren’t helping with validation either, because they aren’t typed and cannot be type checked independently of call sites.  We could consider writing a type inference algorithm that infers types for the polymorphic macros, but that’s another big endeavor, source of complexity, and engineering burden.

An advantage that this proposal has over an external templating system is the potential for better error messages, because a custom macro system (like this one) could propagate source location information through expansion.  

Still, I strongly urge caution here, and specifically, not underestimating the burden that this feature imposes on users, services, and language implementers.

If I had to take a guess at the implementation and maintenance burden for Cedar alone, I would say we’re looking at an additional ~10K LOC, including the Rust implementation (with proper expansion tracking, error messages, etc), Lean model and proof (by having a parallel semantics with CBN and proving it equivalent to the core semantics after expansion), and diff testing.

We need to be really sure that this is the right direction for Cedar before proceeding, and not just because of the engineering burden.

In my view, this is not the right direction from a language design point of view either.  As several other commenters noted, this would push Cedar toward looking more like a general-purpose language.  And that’s a slippery slope, which inevitably leads to more complexity and feature creep.

aaronjeline:
> Can you say what you mean by integration?

syntax highlighting, autocomplete, hover tooltips, potential for type annotations, using the cedar namespace system, using whatever stdlib mechanism we may decide on, checking if a macro is a valid expression, rendering the macro as an EST, runtime errors that track proper source locations, type checking errors that track proper source locations, anything that requires parsing or thinking about the macro in isolation becomes impossible in the "generic template" system answer.

>  As several other commenters noted, this would push Cedar toward looking more like a general-purpose language. And that’s a slippery slope, which inevitably leads to more complexity and feature creep.

This feels like a very ill-defined argument. We still have the guardrail of analysis preventing us from expanding. This adds no expressive power to the language. Templates did not meet this criticism, were we wrong to miss it there? Action groups likewise. Surely opening of the extension mechanism so users can easily share extensions would hit this bar. I'm not saying it's an incorrect argument, it just feels vague. 

I think there's three arguments happening here, and I'll try to distill them.

1. This feature is both a large implementation lift and are large conceptual addition to the language. We have not met the bar of evidence that users want this feature to even begin to consider it. 

This is fine. I may personally disagree, but I'm ok with declining for now and seeing if more evidence accumulates. If that evidence ends up accumulating, we could potentially revive this RFC or consider a different mechanism for the same goal. If it doesn't, we won't.

2. We think users will want this feature, but we don't want to own it (for the reasons above), so we should recommend they use an external template system.

This feels very wrong to me. For all the reasons I've described above. I'll add one more here. If a macro system within cedar brings it to close to a general purpose languages, why does a recommended macro system outside of cedar not do that? Again, operationally,  you have all the same concerns. Semantically/readability wise, you have all the same concerns except with less guadrails. Is there an existing macro system that we can show as an example that more readable than this proposal is? Standards wise it's a mess. If we know people want this feature, and the feature is feasible, then it feels weird to recommend what is essentially a workaround.

3. Regardless of whether people want this feature, we shouldn't do it because we believe abstraction mechanisms like this one don't belong in Cedar. In this world, we also don't recommend using external templating systems, because we think reading/writing cedar policies in terms of a context of non-locally-defined substitutions is not a good idea. 

Obviously I disagree (I wrote the RFC :D ), but there's probably more to talk about here.

max2me:
The more I think about this feature, the more torn on it I become. I see value it provides in streamlining policies and reducing duplication. On another hand, introducing brand new standalone concepts like these macros feels a bit heavy handed, with implications for integrated services.

May I share an idea to get some feedback on it? 

Extension methods in C# inspired this approach. Basically, we'd allow users to define their own type extensions in schema directly. Semver example could then look like this:

```
// Schema declaration
type SemVer = {
	major: Long,
	minor: Long,
	patch: Long,
	greaterThan(?other: SemVer): {
		if this.major == ?other.major then 
			if this.minor == ?other.minor then
				this.patch > ?other.patch
			else
				this.minor > ?other.minor
		else
			this.major > ?other.major
	}
}

// Use in policy
// Permit access when the api is greater than 2.1
permit (principal, action == Action::"someAction", resource)
when {
    resource.apiVersion.greaterThan({ major: 2, minor: 1, patch: 0 })
};
```

I understand that this goes against one of our tenets that schema is not supposed to affect authorization outcome. But I also feel like we've taken incremental steps to relax that stance with schema-based parsing of the request data and schema-based validation of the request.

emina:
Thanks @aaronjeline for summarizing the three arguments!  That's really helpful.  I'm in camps (1) and (3), and below, I'll try to explain why this RFC is triggering my spidey sense, with the full acknowledgement that this opinion is biased (informed? :D) by my own programming experience (Racket, Lean, and a long time ago, Java) and design preferences (minimalist).  

 > 1. This feature is both a large implementation lift and are large conceptual addition to the language. We have not met the bar of evidence that users want this feature to even begin to consider it.

I believe this is true.  We haven't had any users (that I know of) ask for this feature.  We have had many users ask for all other features that have required this level of engineering effort (e.g., partial evaluation).  

We have also not considered a feature that departs this much from the original language design, which focuses on simplicity and "what you see is what you get" behavior.  This focus is lost with no gain in expressiveness and with minor (in my eyes) gain in convenience. It is worth noting that this RFC doesn't solve the problem that originally inspired it, SemVer comparison, because the original ask was for the comparison function to provide a similar API to decimals and IP addresses, which requires writing an extension function to parse SemVer strings.

The effects of supporting this feature are wide-ranging and global.  This feature is not only an engineering lift for us, but also for all applications and services that use Cedar.  We would be forcing them to deal with macro storage, management, distribution, and updates---or risk falling behind the language treadmill as @D-McAdams's comment explains.  Services and applications are our main customers, even more so than individual policy writers, and I would want to see at least 5 services / applications actively advocate for this feature before adding it to the language (borrowing 5 from @andrewmwells-amazon's comment).

> 3. Regardless of whether people want this feature, we shouldn't do it because we believe abstraction mechanisms like this one don't belong in Cedar. In this world, we also don't recommend using external templating systems, because we think reading/writing cedar policies in terms of a context of non-locally-defined substitutions is not a good idea.

I believe this is true as well, and the following is going to stray a bit into subjective preferences and design philosophy, so please bear with me :) 

Cedar is not a general-purpose language, and it was designed explicitly to be analyzable _and_ simple.  The current proposal is to add a feature to the language that is extremely limited (for good reasons), and not really capable of letting you build real abstractions---which are built through composition.  Cedar macros cannot call other macros (for good reasons).  They are not real CBV functions (for good reasons), which is what most programmers expect from functions.  They are also a very weak syntactic abstraction (for good reasons), because they are lacking all the features you find in powerful macro systems (such as Racket or Lean).  In other words, they land us in the uncanny valley of language design (in my eyes), where Cedar looks sort-of general purpose without actually being so.  This then creates the expectation of even more features you find in a general purpose language (as noted by @philhassey):  global constants, module system, packaging system, etc.  And that lands us in a state where Cedar is not beautiful (again, in my eyes) either through simplicity or power.

I cannot think of any popular, simple, standalone DSLs that have things like macros, modules, packages, etc.  (I can think of some that do have these features but are considered complex, and are well out of bounds of what is analyzable.)  This is likely my own blind spot, and I'd be super curious if someone can point me at some DSLs where this has been done well, where it's considered simple and ergonomic, and that we should emulate :)

Cedar is not used like most programming languages.  Most programming languages are file-oriented; people write a single conceptual unit (application) that is thousands of lines of interconnected code; and in that world, things like modules and abstractions are absolutely essential.  While some people may use Cedar by putting all policies in a single file, that is not the case for a large number of users who interact with Cedar through a service or an application.  Here, we see two modes of use:  a (relatively!) small number of static policies, or a small number of templates with thousands of instantiation.  In both cases, the actual logical unit (the authz model) is _not_ thousands of lines of hand-written Cedar code, and in my view, macros or other heavyweight abstraction mechanism are not needed to build and maintain this logical unit.   

Finally, thank you all for a really great discussion.  I've found it immensely helpful in crystallizing my own understanding of my gut reaction :))

PR #63, opened 2024-04-12
Doc comments

john-h-kastner-aws:
[Rendered](https://github.com/cedar-policy/rfcs/blob/doc-comments/text/0063-doc-comments.md)

### DIFF
--- a/text/0063-doc-comments.md
+++ b/text/0063-doc-comments.md
@@ -1,0 +1,226 @@
+# Policy and Schema Doc Comments
+
+## Related issues and PRs
+
+- Reference Issues:
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2024-04-12
+
+## Summary
+
+Cedar should support documentation comments (doc comments) on both policy and schema elements.
+
+Doc comments are a refinement to our current comments, distinguished by a third `/` and restricted in where they can appear.
+These comments are designed specifically to support rendering documentation as HTML pages (or similar), but this RFC does not propose what the final rendered form should be.
+
+## Basic example
+
+A policy can be documented by placing a doc comment, indicated by `///` following the style of Rust, immediately before the policy and on scope conditions.
+
+```cedar
+/// Allow users to create list and see the lists they own.
+@id("policy0")
+permit (
+    principal,
+    /// The `GetLists` action only returns the lists owned by that users, so this
+    /// policy does not allow users to see lists created by others.
+    action in [Action::"CreateList", Action::"GetLists"],
+    resource == Application::"TinyTodo"
+);
+```
+
+Doc comments may also be placed immediately prior schema elements to document the purpose of entity types and actions and before each attribute for an entity type or action context.
+
+```cedarschema
+/// This entity defines our central user type
+entity User { 
+    /// The location where this user works. E.g., ABC17 or XYZ77.
+    location : String,
+    /// A numeric level describing this users job.
+    jobLevel : Long,
+}
+```
+
+## Motivation
+
+Even though the syntax of Cedar policies is designed to be easy to read and understand, policies will still require some amount of documentation. 
+Cedar naturally supports comments so policy authors can leave remarks in their code, but these comments do not exist in our policy JSON format or in our internal AST.
+Automated generation of documentation from comments is a common capability found in almost all modern programming languages, so we should update Cedar to include automatically processable comments that support this use case.
+
+This RFC is largely motived by [this comment on RFC 48](https://github.com/cedar-policy/rfcs/pull/48#issuecomment-1937735924).
+It also addresses the request in [Cedar issue 660](https://github.com/cedar-policy/cedar/issues/660) for comments in the JSON policy format. 
+
+## Detailed design
+
+Doc comments are designed to be very similar to annotations, and could reasonably be implemented as syntactic sugar for a special annotation annotation key.
+The primary difference is that, in order to fully satisfy the documentation use case, doc comments need to appear in more positions than are supported for annotations.
+We may choose to extend annotation to be supported in all positions proposed for doc comments, but this is not a necessary extensions to accept this RFC.
+
+Doc comments will be defined by this grammar.
+
+```
+DocCommentLine ::= '///' [^\r\n]* [\r\n]
+DocComment ::= {DocCommentLine}
+```
+
+### Policies
+
+Documentation for the policy as whole appears before the policy effect.
+It may be anywhere in relation to the annotations on that policy.
+Documentation for individual policy scope clauses must appear immediately before that clause.
+
+The policy grammar is now:
+
+```
+Policy ::= (Annotation|DocComment)} Effect '(' Scope ')' {Conditions} ';'
+Scope ::= [DocComment] Principal ',' [DocComment] Action ',' [DocComment] Resource
+```
+
+### Schema
+
+Documentation for a declaration in a schema appears immediately before the `entity`, `action`, `type`, or `namespace` keyword at the start of the declaration.
+The fields of the attributes in a record type can also be documented with the comment appearing directly before the attribute name.
+A record type might be written for an entities attributes, actions context, or common type definition, or it might be nested inside of some other type.
+Doc comments are supported in all of these locations.
+
+The schema grammar is now:
+
+```
+Namespace := ([DocComment] 'namespace' Path '{' {Decl} '}') | Decl
+Decl := [DocComment] (Entity | Action | TypeDecl)
+AttrDecls := [DocComment] Name ['?'] ':' Type [',' | ',' AttrDecls]
+```
+
+### Improperly placed comments
+
+Comments starting with the `///` that are not in a legal positions for doc comments are an error.
+The definition of comments is accordingly updated to
+
+```
+COMMENT ::= '//' ([^/\r\n] [^\r\n]*)? [\r\n]
+```
+
+This makes the RFC a breaking change for any users who wrote comments in this style, but I believe that is preferable to silently ignoring what a user intended to be a doc comment.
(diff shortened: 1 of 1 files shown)


max2me:
LGTM. I'm especially excited about support for doc comments in JSON policy format.

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+permit (
+    principal,
+    /// The `GetLists` action only returns the lists owned by that users, so this
+    /// policy does not allow users to see lists created by others.
+    action in [Action::"CreateList", Action::"GetLists"],
+    resource == Application::"TinyTodo"
+);
+```
+
+Doc comments may also be placed immediately prior schema elements to document the purpose of entity types and actions and before each attribute for an entity type or action context.
```
cdisselkoen, line 35: what about before `principal` and `resource` in an `appliesTo` block?

(I do acknowledge the comment in Drawbacks that we need not block this RFC on discussing additional positions for doc comments, and additional ones can be added later as a non-breaking change.  We can punt on my comment here if we want.)

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+```
+
+This makes the RFC a breaking change for any users who wrote comments in this style, but I believe that is preferable to silently ignoring what a user intended to be a doc comment.
+For an initial and backwards compatible change, improperly place doc comments could be allowed with a warning.
+We can then make this an error in a future major version of the Cedar, or never make it an error if the break is not acceptable.
+
+### Alternative doc comment syntax
+
+Choosing to use Rust style doc comments feels natural after developing Cedar in Rust, but we should consider alternative syntax options.
+Some options include using block style doc comments (`/** .. */`), using another common comment character for doc comments (e.g., `#`), and picking a different third character following hte usual two slashes (e.g., `//@` to emphasize the connection with annotations).
```
cdisselkoen, line 113: I agree that we should consider all of these options; I'd like to hear others' thoughts on this.

cdisselkoen (approved):
LGTM, with the caveat below about considering all of our syntax options

text/0063-doc-comments.md
john-h-kastner-aws: These would also be good additions. I wasn't sure where to draw the line for what to include in this RFC. Hopefully extra comments positions can be added later without another RFC.

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+    resource == Application::"TinyTodo"
+);
+```
+
+Doc comments may also be placed immediately prior schema elements to document the purpose of entity types and actions and before each attribute for an entity type or action context.
+
+```cedarschema
+/// This entity defines our central user type
+entity User { 
+    /// The location where this user works. E.g., ABC17 or XYZ77.
```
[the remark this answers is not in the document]
john-h-kastner-aws: I'm not a fan of using `//` for doc comments and regular comments. Feels like it could make parsing a pain, and it would be easy to misplace  a doc comment without an way for the parser to notice and warn you.

text/0063-doc-comments.md
john-h-kastner-aws: @philhassey suggests just using `//` as in Godoc which I hadn't considered as an option. Not a huge fan, but it's worth considering.

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+```
+Policy ::= (Annotation|DocComment)} Effect '(' Scope ')' {Conditions} ';'
+Scope ::= [DocComment] Principal ',' [DocComment] Action ',' [DocComment] Resource
+```
+
+### Schema
+
+Documentation for a declaration in a schema appears immediately before the `entity`, `action`, `type`, or `namespace` keyword at the start of the declaration.
+The fields of the attributes in a record type can also be documented with the comment appearing directly before the attribute name.
+A record type might be written for an entities attributes, actions context, or common type definition, or it might be nested inside of some other type.
```
khieta, line 86:
```suggestion
A record type might be written for an entity's attributes, action's context, or common type definition, or it might be nested inside of some other type.
```

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+Namespace := ([DocComment] 'namespace' Path '{' {Decl} '}') | Decl
+Decl := [DocComment] (Entity | Action | TypeDecl)
+AttrDecls := [DocComment] Name ['?'] ':' Type [',' | ',' AttrDecls]
+```
+
+### Improperly placed comments
+
+Comments starting with the `///` that are not in a legal positions for doc comments are an error.
```
khieta, line 99:
```suggestion
Comments starting with the `///` that are not in a legal position for doc comments are an error.
```

text/0063-doc-comments.md
khieta: No strong feelings on my end. I like the look of `///`, but likely I'm biased from having mostly used Rust lately. My second preference would be `//@` (or some other third character like `!` or `#`).

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+The definition of comments is accordingly updated to
+
+```
+COMMENT ::= '//' ([^/\r\n] [^\r\n]*)? [\r\n]
+```
+
+This makes the RFC a breaking change for any users who wrote comments in this style, but I believe that is preferable to silently ignoring what a user intended to be a doc comment.
+For an initial and backwards compatible change, improperly place doc comments could be allowed with a warning.
```
khieta, line 107:
```suggestion
For an initial and backwards compatible change, improperly placed doc comments could be allowed with a warning.
```

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+We can then make this an error in a future major version of the Cedar, or never make it an error if the break is not acceptable.
```
khieta, line 108:
```suggestion
We can then make this an error in a future major version of Cedar, or never make it an error if the break is not acceptable.
```

text/0063-doc-comments.md
khieta, line 113:
```suggestion
Some options include using block style doc comments (`/** .. */`), using another common comment character for doc comments (e.g., `#`), and picking a different third character following the usual two slashes (e.g., `//@` to emphasize the connection with annotations).
```

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+We would need to accept at least [RFC 48](https://github.com/cedar-policy/rfcs/pull/48) to document schema elements.
+
+### Existing tools for literate programming
+
+Rather of building a custom solution for Cedar, we could point users to existing literate programming tools.
+One powerful option is [Entangled](https://entangled.github.io/) which would allow users to structure their documentation however they like.
+
+The drawback to this alternative is that it would require that users install a second tool distributed separately from Cedar.
+Different users may also decide to use different literate programming tools, meaning their policy and schema source may not be directly compatible without a pre-processing pass to remove the literate comments.
+This concern could be addressed by explicitly recommending particular tools, or making directly Cedar parsers directly depend on Rust crates implementing literate programming.
```
khieta, line 223:
```suggestion
This concern could be addressed by explicitly recommending particular tools, or making Cedar parsers directly depend on Rust crates implementing literate programming.
```

text/0063-doc-comments.md
```diff
@@ -0,0 +1,226 @@
...
+Our position would be that documentation should be written in standard Cedar comments. 
+If someone wants to attach a particular meaning to and write documentation tools around comments starting with `///`, they are free to do so, but we do not explicitly support this.
+This requires no changes to Cedar, and is maximally flexible (standard comments can appear anywhere).
+
+Asking users to develop their own documentation tools could resulted in multiple ad-hoc tools that are likely to fail to handle certain edge cases (e.g., `///` in a string) unless they substantially re-implement the Cedar parser.
+Alternatively, we could provide the tools for generating documentation in this manner, but this implies standardizing the format and position of these comments, which is essentially what this RFC proposes, just without considering JSON formats.
+
+Comments do not exist in the JSON format, so this alternative provides nothing for documenting a JSON formatted policy or schema.
+
+### Annotations as documentation
```
khieta, line 190: This is an interesting option to me, but I still prefer the look of `///` (or whatever syntax is chosen). I think it's fine to allow some overlap in functionality between annotations and doc comments. It might also be useful to unify their implementation in the future so that `///` is syntactic sugar for `@doc(...)`.

khieta (approved):
I'm happy with the proposal as-is (although I left some small comments).

text/0063-doc-comments.md
john-h-kastner-aws: Yes. The doc comment syntax being syntactic sugar under the hood is something I had in mind while writing this. It requires a substantial extension to extensions beyond ever RFC #48, but perhaps people will find other application of generalized annotations.

text/0063-doc-comments.md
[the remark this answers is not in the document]
hakanson: I saw it referenced but can you add the link to Rust [Doc comments](https://doc.rust-lang.org/reference/comments.html#doc-comments) in the RFC. C# also uses `///` for [Documentation comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments)

john-h-kastner-aws:
We've tagged this as `archived` because we think that doc-comments should sugar for `@doc` annotations regardless of what the doc-comment syntax is. We plan to first extend where annotations can be placed (do be determined based on the outcome of rfc #48). After that documentation tooling can be developed against the annotation API. We can then re-examine this RFC to introduce syntactic sugar for doc comments.

PR #64, opened 2024-05-07
Referring to types in the empty namespace

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/cdisselkoen/refer-to-types-empty-namespace/text/0064-refer-to-types-in-empty-namespace.md)

### DIFF
--- a/text/0064-refer-to-types-in-empty-namespace.md
+++ b/text/0064-refer-to-types-in-empty-namespace.md
@@ -74,6 +74,9 @@ namespace where there is no collision with a typename in the active namespace.
 The explicit-`::` change deals with referencing a typename in the empty
 namespace where there _is_ a collision with a typename in the active namespace.
 
+This issue has been encountered "in the wild" by multiple people, according to
+a [comment on cedar#579](https://github.com/cedar-policy/cedar/issues/579#issuecomment-2077482637).
+
 ## Detailed design
 
 ### Change 1: Fully qualified syntax for referring to typenames in the empty namespace
--- a/text/0064-refer-to-types-in-empty-namespace.md
+++ b/text/0064-refer-to-types-in-empty-namespace.md
@@ -127,6 +127,18 @@ primtiive and extension types, where we first check for an entity or common type
 `String` in the active namespace before falling back on resolving to
 `__cedar::String`.
 
+### Change 3: Warn when defining a typename that shadows an existing definition
+
+This is a small change (adding a validation warning) that wouldn't require an
+RFC by itself, but since it is related, and came up in comments on this RFC
+thread, it is listed here.
+
+If the user defines an entity/common type that shadows an existing definition
+(i.e., a definition in the empty namespace, or the name of a primitive/extension
+type), we will produce a validation warning for that definition.
+(However, this is not a validation error, and this RFC's Change 1 will still
+allow you to refer to the shadowed name.)
+
 ## Drawbacks
 
 * The schema implementation becomes more complex, as we need more complex
@@ -150,10 +162,15 @@ and JSON schema grammars.
 ## Unresolved questions
 
 * Should we also support the `::id` syntax in policies?
-It's totally unnecessary from an expressiveness perspective, as policies do not
-have a notion of active namespace, so the unqualified name `id` always refers to
-the `id` in the empty namespace and is never implicitly prefixed with any other
-namespace.
-But, users who are used to writing `::User` in schemas (for some `User` type
-they defined in the empty namespace), might expect to also refer to that type as
-`::User` in policies.
+    [2024-05-17: Based on discussion in the RFC thread, the proposed resolution is "no"]
+    * Arguments for:
+        * Consistency: Users will write `::User` in schemas and might expect to
+        be able to write `::User` in policies as well
+    * Arguments against:
+        * It's unnecessary from an expressiveness perspective, as policies do not
+        have a notion of active namespace, so the unqualified name `id` always refers to
+        the `id` in the empty namespace and is never implicitly prefixed with any other
+        namespace.
+        * We can catch `::User` in the policy parser and provide a specific, informative
+        error message, which makes this less of a sharp edge
+        * We can always go back and add support for `::id` in policies later, if needed
--- a/text/0064-refer-to-types-in-empty-namespace.md
+++ b/text/0064-refer-to-types-in-empty-namespace.md
@@ -2,7 +2,7 @@
 
 ## Related issues and PRs
 
-- Reference Issues: [cedar#579](https://github.com/cedar-policy/cedar/issues/579)
+- Reference Issues: [cedar#579]
 - Implementation PR(s):
 
 ## Timeline
@@ -38,9 +38,10 @@ namespace Demo {
 
 Currently, the `Demo::User` declaration is an error, because in its `id`
 attribute, the `id` type implicitly refers to `Demo::id`, which doesn't exist.
-With this RFC, this example would implicitly work correctly, because when
-`Demo::id` is not found, the resolver would fall back to looking for `id` in the
-empty namespace, and find the appropriate declaration.
+With [cedar#579], this example
+would implicitly work correctly, because when `Demo::id` is not found, the
+resolver would fall back to looking for `id` in the empty namespace, and find
+the appropriate declaration.
 
 As a separate issue, in the `owner` attribute of `Demo::Account`, the `User`
 type implicitly refers to `Demo::User` and not the `User` in the empty
@@ -60,22 +61,27 @@ The fact that RFC 24 did not provide a way to refer to typenames in the empty
 namespace by using some kind of fully qualified name seems like an oversight,
 and should be corrected for consistency.
 
-The other change proposed in this RFC, implicitly falling back to the empty
+The other change discussed in this RFC, implicitly falling back to the empty
 namespace (so that the `Demo::User` declaration is legal in the example above),
-is an ergonomic change to avoid having to use the new syntax in the common case
-where there is actually no typename collision (i.e., where the typename is not
-actually defined in both the active and the empty namespace).
+can be viewed as an ergonomic change to avoid having to use the new syntax in
+the common case where there is actually no typename collision (i.e., where the
+typename is not actually defined in both the active and the empty namespace).
+However, it is actually better characterized as a bugfix and has been pulled out
+of this RFC into [cedar#579]; see discussion there.
 
-The two changes are theoretically separable and/or severable, but make sense to
-address together, as they both relate to how we reference typenames in the empty
+Both changes relate to how we reference typenames in the empty namespace.
+The implicit-fallback change ([cedar#579]) deals with referencing a typename in
+the empty namespace where there is no collision with a typename in the active
 namespace.
-The implicit-fallback change deals with referencing a typename in the empty
-namespace where there is no collision with a typename in the active namespace.
-The explicit-`::` change deals with referencing a typename in the empty
-namespace where there _is_ a collision with a typename in the active namespace.
+The explicit-`::` change (Change 1 below, in this RFC) deals with referencing a
+typename in the empty namespace where there _is_ a collision with a typename in
+the active namespace.
 
 This issue has been encountered "in the wild" by multiple people, according to
 a [comment on cedar#579](https://github.com/cedar-policy/cedar/issues/579#issuecomment-2077482637).
+It's unclear whether Change 1 is needed in any of these cases though, or if
+Change 2 / the newly-rescoped [cedar#579] is sufficient for these real-world
+cases.
 
 ## Detailed design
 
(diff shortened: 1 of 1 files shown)


chrnorm:
+1 to the `::EntityType` syntax, especially the implicit fallback to the default namespace. I recently hit an issue similar to https://github.com/cedar-policy/cedar/issues/579 myself. My use case was similar to this:

```
entity User {};

namespace MyService {
    entity Customer {};

    action GetCustomer appliesTo {
        principal: User,
        resource: Customer
    };
}
```

As a developer using Cedar my expectation was that `User` would be inferred to be in the default namespace automatically. I worked around this by shifting `User` into a namespace:

```
namespace MyService {
    entity User {};
    entity Customer {};

    action GetCustomer appliesTo {
        principal: User,
        resource: Customer
    };
}
```

This is a pain though because it makes policies less concise, now I have to use `MyService::User` instead of `User`.

No comment on the parsing trade-offs as I haven't worked with the Cedar parsing internals -- more so a comment as a Cedar user that I would definitely benefit from this feature :)

mwhicks1:
Can you comment about how these issues are handled in other languages that have namespaces?

For example, other languages have an explicit `import` construct that controls name resolution. Suppose I defined `User` in `Foo::Bar` and defined `User` in the top-level namespace. A reference to `User` will resolve to the first only if it is fully qualified. If you try to `import` it, that creates an ambiguity: `User` could refer to `Foo::Bar::User` or `User`. In Java, [such an `import` would be an error](https://docs.oracle.com/javase/specs/jls/se9/html/jls-7.html). What about other languages?

andrewmwells-amazon (approved):

andrewmwells-amazon:
Happy to take this as-is. I prefer the `::Foo` syntax to `_cedar::Foo`. I don't think we need to allow `::Foo` in policies since it's easy to suggest `Foo` in a linter.

andrewmwells-amazon:
> If you try to `import` it, that creates an ambiguity: `User` could refer to `Foo::Bar::User` or `User`. In Java, [such an `import` would be an error](https://docs.oracle.com/javase/specs/jls/se9/html/jls-7.html). What about other languages?

I'm not aware of a language where that isn't an error.

Some language allow you to alias. E.g., in Rust you can do:
```
mod Foo {
    pub fn my_fun() -> usize {
        0
    }
}

mod Bar {
    pub fn my_fun() -> usize {
        1
    }
}

fn main() {
    use Foo::my_fun;
    use Bar::my_fun as other_my_fun;

    println!("Hello, world! {}", my_fun());
    println!("Good bye! {}", other_my_fun());
}
```

hakanson:
`_root_` in Scala (see https://docs.scala-lang.org/style/naming-conventions.html#root) was a "fully-qualify" escape hatch.

I might prefer something like `__::User` over `::User` - I think it needs to be different that `__cedar`

I also wondered why `::User` only works in schemas vs also in Cedar Policies (e.g. `principal is ::User`)

cdisselkoen:
> how these issues are handled in other languages that have namespaces?

Python's [LEGB rule](https://realpython.com/python-scope-legb-rule/) and [JavaScript's identifier resolution rules](https://medium.com/@danparkk/javascript-basics-identifier-resolution-and-scope-with-es6-block-scope-897abacb1458) are both similar to this RFC's Change 2.

The comparison to `import` in Java or `use` in Rust is not quite apples-to-apples in my mind.  `import` and `use` are about bringing items from other non-empty namespaces into the empty namespace so that they can be referenced without qualification.  Cedar doesn't have a way to do this, and this RFC doesn't propose one.  In both Java and Rust, to my knowledge, you can always refer to items in the empty namespace without qualification (which is what this RFC's Change 2 proposes).  Java and Rust prohibit shadowing these definitions with definitions in the active namespace; Cedar, Python, and JS do not prohibit this, and changing our rules to prohibit this would be a breaking change.

text/0064-refer-to-types-in-empty-namespace.md
```diff
@@ -0,0 +1,159 @@
...
+## Unresolved questions
+
+* Should we also support the `::id` syntax in policies?
+It's totally unnecessary from an expressiveness perspective, as policies do not
+have a notion of active namespace, so the unqualified name `id` always refers to
+the `id` in the empty namespace and is never implicitly prefixed with any other
+namespace.
+But, users who are used to writing `::User` in schemas (for some `User` type
+they defined in the empty namespace), might expect to also refer to that type as
+`::User` in policies.
```
cdisselkoen, line 159: Starting a thread here because threaded conversation (attached to a particular point in the text) is easier to follow than the free-flowing conversation on the main RFC thread.

@andrewmwells-amazon said:

> I don't think we need to allow `::Foo` in policies since it's easy to suggest `Foo` in a linter.

@hakanson said:

> I also wondered why `::User` only works in schemas vs also in Cedar Policies (e.g. `principal is ::User`)

cdisselkoen: The arguments against allowing `::User` in policies appear to be
1. it's unnecessary since `principal is User` is already completely unambiguous -- this is argued in the RFC text quoted above
2. we can catch this and provide an informative error message as suggested by @andrewmwells-amazon 

I'll add a third one, which is that we can always go back and support this later, even if we don't right now.

The arguments _for_ appear to be
1. consistency, since users will write `::User` in schemas and might expect to be able to write `::User` in policies as well (as argued in the RFC text and also by @hakanson)

hakanson: I agree with @cdisselkoen this can be added to policies later (if needed), since "best practice" to have a namespace, and this fixes the gap from mixing no namespace with a namespace in the same schema.  Would there be plans to add a "shadows warning" when validating the schema like with builtin Cedar names?

> The name `String` shadows a builtin Cedar name. You'll have to refer to the builtin as `__cedar::String`.

> The name `User` shadows a un-namespaced name. You'll have to refer to the  un-namespaced as `::User`.

andrewmwells-amazon: > The name User shadows a un-namespaced name. You'll have to refer to the un-namespaced as ::User
Because policies aren't namespaced, this doesn't come up. `User` can only refer to `User` in the empty namespace.

andrewmwells-amazon: Or maybe I misunderstand you and you're asking if we'll get a warning in the schema when in namespace `Bar` I define and mention `User` and there is also a type `::User`. In that case, yes I think we could give a warning, but we won't want it to error to preserve backwards compatibility.

hakanson: @andrewmwells-amazon - in the schema (not policy) as a schema validation warning similar to the other shadows errors

text/0064-refer-to-types-in-empty-namespace.md
```diff
@@ -0,0 +1,159 @@
...
+This does not apply:
+* to primitive and extension types. From the user's perspective, those live in
+the `__cedar` namespace and not in the empty namespace. `::String` refers to an
+entity or common type `String` defined in the empty namespace, while
+`__cedar::String` refers to the builtin primitive type `String`.
+* to "names" like `::A::B` -- that syntax remains an error. `A::B` is already an
+unambiguous, fully-qualified name, and will never have any namespace implicitly
+prepended.
+
+### Change 2: Implicitly falling back on the empty namespace
```
shaobo-he-aws, line 103: I think this solution doesn't solve the fundamental issue where there are type name conflicts between a namespace and the empty space.

shaobo-he-aws:
> In both Java and Rust, to my knowledge, you can always refer to items in the empty namespace without qualification (which is what this RFC's Change 2 proposes).

IIRC, in Rust, we need to use `super` to refer to names of the "empty namespace", should there exist name conflicts.

shaobo-he-aws:
Now my understanding is that we don't want to design a hierarchy of namespaces in a schema. And hence type names like `::id` make sense. It's just the anomalies in names that worry me.

text/0064-refer-to-types-in-empty-namespace.md
john-h-kastner-aws: I think it solves the main problem people encounter. Conflicting type names feels like an edge case, what's more important is just being able to refer to types in the empty namespace.

IMO this change could go in without an RFC

text/0064-refer-to-types-in-empty-namespace.md
```diff
@@ -0,0 +1,159 @@
...
+One proposal in the original
+[cedar#579](https://github.com/cedar-policy/cedar/issues/579) thread was to use
+something in the `__cedar` namespace, e.g., `__cedar::schema::top_level::id`.
+Other possibilities include other special characters, like `^id` or `#id`, or even
+a special keyword, like `root id`, which should be unambiguous in both the human
+and JSON schema grammars.
+
+## Unresolved questions
+
+* Should we also support the `::id` syntax in policies?
```
john-h-kastner-aws, line 152: At a minimum we'll need to update our parsers so they don't barf on `::id`. People will make this mistake.

john-h-kastner-aws (approved):
Supporting this without adding `::id` to the policy grammar, but I wouldn't object too strongly to that option.

text/0064-refer-to-types-in-empty-namespace.md
cdisselkoen: We should definitely put in warnings like that, during validation

text/0064-refer-to-types-in-empty-namespace.md
cdisselkoen: @shaobo-he-aws I think Change 1 (not Change 2) is intended to solve that issue?  Or am I misunderstanding?

cdisselkoen:
> IIRC, in Rust, we need to use `super` to refer to names of the "empty namespace", should there exist name conflicts.

I don't think this is quite accurate -- `super` allows you to refer to names in your parent namespace.  E.g., if the active namespace is `A::B::C`, then`super::foo` refers to `A::B::foo`

aaronjeline (approved):

text/0064-refer-to-types-in-empty-namespace.md
```diff
@@ -0,0 +1,159 @@
...
+
+This applies to both the human (RFC 24) and JSON schema syntaxes.
+
+This is reminiscent of, but not dependent on, the existing similar behavior for
+primtiive and extension types, where we first check for an entity or common type
+`String` in the active namespace before falling back on resolving to
+`__cedar::String`.
+
+## Drawbacks
+
```
mwhicks1, line 131: Another possible drawback is that syntax `::Foo` is unprecedented for existing languages, as far as I understand it. So customers would not reach for it intuitively.

It also looks a little strange to me to see `owner: ::User`; that extra space is now really important. You might want to specially deal in the parser with the mistake of writing `owner:::User`

text/0064-refer-to-types-in-empty-namespace.md
mwhicks1: I am worried about conflicts, too.

I am also wondering if we should be looking ahead to schemas that will reference other schemas and leverage mechanisms like `import`. If we had `import`, would we feel the need for `::User`? Is `::User` basically an anti-pattern -- you shouldn't be defining entity types at the top level, and should be using namespaces and `import`?

text/0064-refer-to-types-in-empty-namespace.md
hakanson: if `owner:User` and `owner:Foo::User` are valid without a space after the first `:`, then you could argue `owner:::User` should be parsable.  I found that this seems to parse just fine without extra spaces:

```
entity P;
action b;
action"a"in[b]appliesTo{principal:P};
```

I still like `__::User` since `__` is already a special prefix and can't collide with a user defined namespace.  `owner:__::User` is perfectly readable.

cdisselkoen: > `__` is already a special prefix and can't collide with a user defined namespace

The current draft of [RFC 52](https://github.com/cedar-policy/rfcs/pull/52) proposes reserving just `__cedar` and not all namespaces starting with `__`.

hakanson: I misremembered the ident rules and thought `__` wouldn't be valid today and would be safe.  Thanks for the clarification.

```
IDENT ::= ['_''a'-'z''A'-'Z']['_''a'-'z''A'-'Z''0'-'9']* - RESERVED
```

text/0064-refer-to-types-in-empty-namespace.md
khieta: Based on this thread, I lean towards not allowing `::EntityType` in policies, but making sure we add a special error message / help text in case a user (reasonably) tries to write this.

khieta (approved):
Looks great. I'm happy with the proposal as-is.

text/0064-refer-to-types-in-empty-namespace.md
cdisselkoen: > I am worried about conflicts, too.

Can you say more about what you're worried will happen?

text/0064-refer-to-types-in-empty-namespace.md
cdisselkoen: In my opinion, the concerns in this thread are valid and sufficient to make us prefer a syntax other than `::User`.  In particular, I agree that `owner:::User` would have to be legal and that it's confusing to mentally parse.

I'll come up with an alternative and revise this RFC.

text/0064-refer-to-types-in-empty-namespace.md
cdisselkoen: Updated the "Unresolved questions" section of the RFC with the conclusion of this thread.  Also added a "Change 3" to this RFC specifying the validation warning (not the policy-parsing warning) discussed in this thread.

cdisselkoen:
After some offline discussion, pulled Change 2 back into cedar-policy/cedar#579, as it's uncontroversial, better characterized as a bugfix, and doesn't need an RFC.  This RFC will remain open, for discussion of Changes 1 and 3 only.  Revised the RFC text accordingly.

cdisselkoen:
After the above comments and offline discussion, we failed to arrive at a satisfactory syntax for Change 1, and instead the group's conclusion was in favor of disallowing this kind of collision altogether, mooting Change 1, and making Change 3 an error (at least for the case of shadowing an empty-namespace definition).  Note again that Change 2 was pulled back into cedar-policy/cedar#579 where it will be addressed.  As a result, I am closing this RFC and opening #70 instead as the successor.

PR #66, opened 2024-05-24
open records for simple maps

mwhicks1:
Proposes to extend syntax of schemas, and validation, to support _open_ records, which are useful for implementing simple maps from string to values.

[Rendered](https://github.com/cedar-policy/rfcs/blob/open-record/text/0066-open-record.md)

EDIT: Please shift your attention to a [new RFC that proposes entity tags](https://github.com/cedar-policy/rfcs/blob/tags/text/0068-entity-tags.md). It maximizes the benefits of this RFC (and adds some additional ones) while losing the drawbacks.

### DIFF
--- a/text/0066-open-record.md
+++ b/text/0066-open-record.md
@@ -1,0 +1,164 @@
+# The default attribute type for open records
+
+## Related issues and PRs
+
+- Reference Issues: [#305](https://github.com/cedar-policy/cedar/issues/305)
+- Supercedes [RFC 27](https://github.com/cedar-policy/rfcs/pull/27)
+- Implementation PR(s): (leave this empty)
+
+## Timeline
+
+- Start Date: 2024-05-22
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
+
+The Cedar validator requires a user to enumerate all possible record and entity attributes, both names and types, in the schema. This is limiting when you'd like to treat a record like a map, i.e., there are attributes whose names are not known in advance, but whose value types are.
+
+This RFC proposes to extend the validator to support _open records_. An open record type extends a normal record type with a _default attribute type_. Policies can check for the presence of any attribute in an open record using `has`, and if present they can access it assuming its value has the default attribute type.
+
+## Motivating example: Tags as Attribute-based Maps
+
+As a key motivating example, you can use open records to implement _tags_, which are basically a map from `String` keys to values.
+
+For example, suppose that `principal.tags` has type `{ } default String`, meaning it is a record with default attribute type `String`. Then you can write `principal.tags has "priority" && principal.tags["priority"] == "green"`. Just writing `principal.tags["priority"] == "green"` without a preceding `has` check would result in a validation error.
+
+Compare this approach to another way of implementing tags, as sets of key-value pairs. For example, `principal.tags` could have type `Set<{ key:String, value:String }>` and a tag check would involve an expression like `principal.tags.contains({ key: "priority", value: "green" })`. Attribute-based tags are useful when you want to compare tags on different entities, as in 
+`resource.tags["project"] == principal.tags["project"]` 
+or 
+`resource.tags["projects"].contains(principal.tags["project"])` (when tags have type `{ } default Set<String>`, so tags' values are sets of strings). Doing this is not possible when encoding tags as sets.
+
+Compared to the status quo, open records allow you to implement attribute-based tags without changing the schema each time you add a tag. See the [Workarounds](#workarounds) section below for discussion.
+
+## Detailed design
+
+Supporting open records will requre extensions to the Cedar schema format and the validator. Neither the JSON format for entities nor the JSON or natural syntax for policies needs to change to support open record objects.
+
+### Schema
+
+We extend schemas to support a new _default attribute type_ on both records and entities. Here's the natural syntax extension:
+```
+Entity     := 'entity' Idents ['in' EntOrTyps] [['='] RecType] ';'
+RecType    := '{' [AttrDecls] '}' ['default' Type]
+AttrDecls  := Name ['?'] ':' Type [',' | ',' AttrDecls]
+Type      := PRIMTYPE | Path | SetType | RecType
+```
+In essence: We simply extend the definition of record types (used for records and entities) with an optional, default attribute type. Here's a simple example:
+```
+entity User = {
+    name: String,
+} default Long;
+```
+This example shows that `User` has known attribute `name` of type `String`, and a default attribute type `Long`. The JSON syntax requires a new `default` attribute which contains a type. Here's the `User` example in JSON format:
+```
+"User" : {
+    "shape" : {
+        "type" : "Record",
+        "attributes" : {
+            "name" : {
+                "type" : "String"
+            }
+        },
+        "default": {
+            "type" : "Long"
+        }
+    }
+}
+```
+The grammar specifies the defaul attribute type _outside_ the syntax defining the record's named attributes to avoid confusion. In particular, since all legal strings are legal record attributes, having syntax inside the `{` `}` could lead to ambiguity. For example, we might like to write `{ *:String }` to indicate that the "default attribute" `*` has type `String`. But `*` is itself a legal attribute name, so we'd be able to write `{ "*": Long, *:String }` to define a record type with a concrete attribute `*` of type `Long` and a default attribute type of `String`. As another example, we might like to write the default type separately from the other attributes' definitions inside the braces. For example, we might like `{ level: Long | String }` to define a record with attribute `level` of type `Long`, and a default attribute type of `String`. But users might see such a type and think that `level` is a union type (which Cedar does not currently support, but could in the future) of type either `Long` or `String`. We could think of syntax within the braces that solves both problems, but adding the default type outside the attribute declarations avoids these and other sources of confusion, and also adds structural parity with the JSON version of the syntax (where `default` is separate from `attributes`). See [Alternatives](#alternatives) for other possible syntax.
+
+### Validating policies
+
+We extend the way the policy validator handles records to support the default attribute.
+
+#### Capabilities
+
+While typechecking a policy, the validator tracks _capabilities_, which represent the attribute-accessing expressions that are sure to succeed. If `principal` has type `User` and `User` has an optional `Boolean` attribute `sudo`, then the expression `principal.sudo` only validates if `principal.sudo` is present in the _current capability set_. Capabilities are added to that set by a preceding expression of the form `principal has sudo`, e.g., as in `principal has sudo && principal.sudo`.
+
+Capability tracking must be generalized to support open records. Today, if you were to write expression `principal has barney` and `principal`'s type `User` has no `barney` attribute, the expression will be given type `False` since the validator knows `principal` can never have this attribute. But if `User` has default attribute type `String` and we see `principal has barney && principal.barney like "friend"`, then `principal has barney` should be given type `Bool` and `barney` should be added to the capability set. Then, when checking `principal.barney like "friend"` the validator will see `barney` in the capability set, and expression `principal.barney` should be given the default attribute type `String`.
+
+#### Record expressions
+
+A record literal is given _closed_ type, i.e., one without a default attribute type. For example, the literal `{ name: "jose", jobLevel : 7 }` has type `{ name: String, jobLevel: Long }`. The validator does this because it knows that the record has no additional attributes beyond those in the literal.
+
+#### Subtyping
+
+For permissive validation, we adjust the subtyping rules as follows. First, we add the following rule (1):
+```
+{ f1-: T1, ..., fn-: Tn } <: { f1-: T1, ..., fn-: Tn } default T
+```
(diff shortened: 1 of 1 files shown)


jszmajda:
Working with tags remains challenging for my project—we currently create entities that have the property `tags: Set<String>`, and artificially remap tags into strings of form `key: value`. We can then write policies like `resource.tags.contains("label: foo")`. This works out ok, but feels a bit clunky and forces us to rewrite `:`s in key names (to `-`). 

My project would be improved with this RFC if it's possible to implement.

text/0066-open-record.md
```diff
@@ -0,0 +1,164 @@
...
+Compare this approach to another way of implementing tags, as sets of key-value pairs. For example, `principal.tags` could have type `Set<{ key:String, value:String }>` and a tag check would involve an expression like `principal.tags.contains({ key: "priority", value: "green" })`. Attribute-based tags are useful when you want to compare tags on different entities, as in 
+`resource.tags["project"] == principal.tags["project"]` 
+or 
+`resource.tags["projects"].contains(principal.tags["project"])` (when tags have type `{ } default Set<String>`, so tags' values are sets of strings). Doing this is not possible when encoding tags as sets.
+
+Compared to the status quo, open records allow you to implement attribute-based tags without changing the schema each time you add a tag. See the [Workarounds](#workarounds) section below for discussion.
+
+## Detailed design
+
+Supporting open records will requre extensions to the Cedar schema format and the validator. Neither the JSON format for entities nor the JSON or natural syntax for policies needs to change to support open record objects.
```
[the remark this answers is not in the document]
mwhicks1: > Every component of Cedar will be affected, including partial evaluation and analysis.

I agree that analysis is affected inasmuch as it relies on validation.

I don't understand why partial evaluation is affected, because in my understanding partial evaluation is about evaluation, defined independently of validation. And there is nothing about evaluation that is affected by this RFC. What am I missing?

> (core datastructures need to change).

Can you elaborate on why this is necessary? Again, on the face of it we are only changing the validator, not the evaluator.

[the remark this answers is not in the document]
emina: Partial evaluation will have to deal with open records too.  

Basically, implementing this requires changing the definition of `Cedar.Spec.Value`, which is a core definition in the model that everything depends on.

Plus, PE models will also have to change their representation of partial values to support open records.

Analysis does not depend on validation, but it will have to fundamentally change to support open records.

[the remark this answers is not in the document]
emina: Ah, I see that this is _not_ suggesting adding _default_ at the `Value` level, just the type level.  And we can get away with this because of dynamic semantics for concrete and partial evaluation.

Unfortunately, the difficulties with analysis still remain because it's type-directed.  (It will have to change significantly.)

text/0066-open-record.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Schema
+
+We extend schemas to support a new _default attribute type_ on both records and entities. Here's the natural syntax extension:
```
emina, line 41: It occurred to me that if we were to support the default attribute type on _just_ entities, then we wouldn’t need to use the theory of arrays in the encoding.  In fact, we could get away with encoding open entities (but _not_ open records) using just uninterpreted functions.  The reason is that we define equality on entities as equality on their names.  Therefore, we never compare the special “attribute record” of two entities for equality, and it’s not possible to nest such record inside of each other, or inside of sets, directly.  As a result, we don’t need to represent them with arrays—they don’t need to be first-class values on which equality is defined.  They can be represented simply as uninterpreted functions.

How do we feel about making entities (more) special in this regard, and allowing open entities but not open records?

emina: And making only entities open gets rid of some of the sharp edges in strict typing, outlined in the Drawbacks section below.

emina: (I’m not sure how I feel about making entities special in that way but it’s an interesting option in terms of analyzability at least.)

mwhicks1: The motivating use-case is not supported by this restriction, unfortunately; you'd have to do a workaround.

In particular, `tags` is an attribute of an entity, like `User` or `Document`, and it is implemented as an open record, e.g., 
```
entity User = {
  jobLevel: Long,
  manager: User,
  tags: {} default String,
};
```
with a policy like
```
permit (principal is User, action == Action::"writeDoc", resource is Document) when {
  principal.tags has write && resource.tags has write && principal.tags.write == resource.tags.write
};
```
You could work around this problem by lifting the tags into the entity itself. To avoid clashes with the names of other attributes, you could use a naming convention that prefixes the tags with `tag_`, and then policies and requests would have to remember to add that prefix. E.g.,
```
entity User = {
  jobLevel: Long,
  manager: User,
} default String;
```
and
```
permit (principal is User, action == Action::"writeDoc", resource is Document) when {
  principal has tags_write && resource has tags_write && principal.tags_write == resource.tags_write
};
```
This workaround probably doesn't lose too much flexibility under strict validation because a record literal is never given an open type, e.g., you could not write something like the following:
```
permit( principal, action, resource ) when { 
  principal.tags == { write: "owner" } &&
  !resource.tags has "nowrite"
```
This seems contrived -- I don't see that tags would ever be used in a first-class way. But nevertheless this is more restrictive, and a bit kludgey.

> And making only entities open gets rid of some of the sharp edges in strict typing, outlined in the Drawbacks section below.

What are these?

emina: > You could work around this problem by lifting the tags into the entity itself.

A nicer workaround is to define a separate Tags entity type and use that as you would an open record.  That way, the motivating use case remains the same in terms of how the policy is written:

```
entity Tags = { _ : default String };
entity User = {
  jobLevel: Long,
  manager: User,
  tags: Tags,
};
```

We can also change the proposed schema to allow _named_ (rather than default) "open attributes", and you could have as many of those as you wanted, e.g.:

```
entity User = {
  jobLevel: Long,
  manager: User,
  tags: {_ : default String}, // or some other syntax
  otherStuff: {_ : default Long}
};
```

> you could not write something like the following: 
```
permit( principal, action, resource ) when { 
  principal.tags == { write: "owner" } &&
  !resource.tags has "nowrite"
```

We can't write this anyway in the current proposal because it won't pass strict type checking:  `principal.tags` is an open record while `{write: "owner"}` is a closed record with a single attribute.

The more I think about it, the more strongly I want to urge us to consider supporting just open-entities (in whatever form---either arbitrary named one or a single default one), without supporting first-class open records.  It is going to be very tricky to define a function for extracting finite counterexamples out of array-based ones produced by the solver if we allow types like `Set<Set<{_ : default Set<{_ : default Set<Set<String>>>}}>>`.  

For example, if we have a policy that references N sets that have open records embedded somewhere in the type definition, then we have to finitize their contents in such a way that the finitized CEX preserves all the intersection / containment / equality / membership relations that hold for the array-based CEX.  For soundness, we would also need to show that using array-based semantics won't _miss_ any CEXs (i.e., something is valid under the infinite semantics but not under the finite one).   

> What are these?

Everything described in "Unintuitive strict validation behavior".  Note that all these examples rely on open records, and this confusion is not possible with open entities, since entities are already treated as nominal types for the purpose of checking conditionals, equalities, etc.

mwhicks1: > A nicer workaround is to define a separate Tags entity type

Both workarounds are problematic, to me. In your proposal, you have to generate a fresh Id for the entity (e.g., `Tags::"tag1234"`) and then create an entity in the input entities with that name, containing the tags. This tags entity is included separately from the one that refers to it (the "host"). In my proposed workaround, the tags are still part of the host entity, but all attribute names are prefixed with `tags_` (say). Neither one is as clean as allowing tags to be open records.

>  allow named (rather than or in addition to default) "open attributes",

You suggested syntax like this:
```
entity User = {
  jobLevel: Long,
  manager: User,
  tags: {_ : default String}, // or some other syntax
  otherStuff: {_ : default Long}
};
```
How are such entities represented? Are `tags` and `otherStuff` now references to other entities somehow? Are they "namespaces" around attributes within the `User` record itself, basically sugar around the name prefixing I suggested above? Either way, what does the JSON representation of entities look like for these?

> The more I think about it, the more strongly I want to urge us to consider supporting just open-entities (in whatever form---either arbitrary named one or a single default one), without supporting first-class open records.

Will this limit nested use of open records? We have a use-case that nests them, written:
```
entity User {
  allowedTagsForRole: { } default ({ } default Set<String>)
};
```
Assuming I make these tags entities, I'd write something like
```
entity User {
  allowedTagsForRole: RoleTags
};
entity RoleTags {} default Tags;
entity Tags {} default Set<String>;
```
Is that right? This seems like it might be more error prone than writing the tags as nested records inside the host, but maybe validation will catch any mistakes. I am wondering: Why is this approach more likely to net better counterexamples? It seems that it might be hard to link together related entities to their tags in a helpful way.

emina: > How are such entities represented?

If they are embedded in the entity, then it's just a map, rather than an indirection through another entity.  So, for the user Alice, the `attributes` part would contain something like `{ tags: {"foo": "bar", "bar": "baz"}}`---regular record syntax.  This is okay because these embedded records are not first-class:  that's the key from the analyzability perspective.

> Will this limit nested use of open records?

Yes, there will be no nested records, so achieving this effect will require one of the above workaround.  

Alternatively, we can let you have open attributes that accept an _arbitrary_ number of keys, not just one.  For example, something like:

```
entity User {
  allowedTagsForRole: String -> String -> Set <String>
};
```
The above syntax is not great, so ignore it for now.  It's a placeholder to convey the idea that `allowedTagsForRole` is basically a map where the key consists of a pair of strings, and you could dereference such a map as `principal.allowedTagsForRole["foo", "bar"]`.

I'm not necessarily advocating for this, just mentioning that it's an option in terms of what's analyzable.

>  Why is this approach more likely to net better counterexamples? 

Ah, the alternative proposal is not about getting better counterexamples: it's about getting finite counterexamples from infinite models.  When the "open" part is _not_ a first-class value (so, just open entities), this is relatively straighforward to do---in fact, we already have to do it because we use uninterpreted functions to represent the entity store (which is also not a first class value).  But if we have a first class value with an "open" part that is represented by an infinite array, then doing this finitization correctly (and proving that it's correct) becomes very difficult.  The issue is that the SMT solver would give us a model that contains infinite arrays (defined over _all_ strings), and we have to turn each such infinite array into a finite map (defined over a subset of strings), and make sure this finitization works through arbitrary nestings of infinite arrays in each other and in sets etc.

Here is a concrete example. 

Suppose that the SMT solver returns the following representations for fields `tags` and `stuff`, each of which is an open record over strings:

```
tags  := lambda (key : String) : if key = "foo" then "bar" else "baz" --- infinite array
stuff := lambda (key : String) : if key = "foo" then "bar" else if key = "bar" then "baz" else "baz" --- infinite array
```

It is clear than these two infinite arrays are the same.  But these are not valid Cedar values:  we have have to turn these infinite arrays into maps over a finite set of keys.  Which keys?  It depends on how the policies and property that we just analyzed use these maps. 

For example, suppose this is the only relevant constraint: `principal.tags["foo"] == principal.stuff["foo"]`. Then, I can finitize them as just `{ "foo": "bar"}`.  But this finitization would not be valid if we had other constraints (e.g., we'd want `princinpal.tags != principal.stuff` to evaluate to false).   

This gets really complicated when the infinite arrays are arbitrarily embedded into sets (of records of sets of ...), because the finitization then has to figure out how to preserve all equality, containment, intersection, and membership relations between such nested values, so that all policy / property subexpressions behave the same under the finitized model as they do under the array-based model.

emina: To my mind, the main thing that's ugly about embedding arbitrary "open" attributes into an entity is that you couldn't compare them for equality (because they aren't first-class).  So, you couldn't write `prinicpal.tags == principal.stuff`, for example.  You could only use these named attributes with the `has` and `.` operators, which I agree is not ideal and is a sharp edge.  That's the place where the second-class status of open attributes becomes apparent.

mwhicks1:
I've dropped a [new RFC that proposes entity tags](https://github.com/cedar-policy/rfcs/blob/tags/text/0068-entity-tags.md). I believe it properly responds to @emina 's concerns while still satisfying the motivating use-cases. Please shift your attention to that one.

text/0066-open-record.md
mwhicks1: I've dropped a [new RFC that proposes entity tags](https://github.com/cedar-policy/rfcs/blob/tags/text/0068-entity-tags.md). I believe it properly responds to @emina 's concerns while still satisfying the motivating use-cases. Please shift your attention to that one.

mwhicks1:
> Working with tags remains challenging for my project—we currently create entities that have the property `tags: Set<String>`, and artificially remap tags into strings of form `key: value`. We can then write policies like `resource.tags.contains("label: foo")`. This works out ok, but feels a bit clunky and forces us to rewrite `:`s in key names (to `-`).
> 
> My project would be improved with this RFC if it's possible to implement.

@jszmajda I've dropped a [new RFC that proposes entity tags](https://github.com/cedar-policy/rfcs/blob/tags/text/0068-entity-tags.md). I believe it properly responds to @emina 's concerns while still satisfying the motivating use-cases. Does it still work for your project?

PR #68, opened 2024-05-29
Embedded Attribute Maps

mwhicks1:
Proposes to add entity attribute maps, a feature useful for encoding tags that are attached to entities.

[Rendered](https://github.com/cedar-policy/rfcs/blob/tags/text/0068-entity-tags.md)

### DIFF
--- a/text/0068-entity-tags.md
+++ b/text/0068-entity-tags.md
@@ -1,4 +1,4 @@
-# Entity tags
+# Embedded attribute maps
 
 ## Related issues and PRs
 
@@ -15,13 +15,13 @@
 
 ## Summary
 
-This RFC proposes to extend the Cedar type system with the ability to attach _tags_ to entities. Entity tags are a set of key/value pairs with an interface similar to records: keys are like record attributes and values are like attribute values. But unlike records, keys need not be enumerated in advance, as is required with record attributes, and all tag values must have the same type.
+This RFC proposes to extend the Cedar type system with the ability to include _embedded attribute maps_ (EA-maps for short) in entity types. For evaluation purposes, EA-maps have the same programming interface as records: keys are like record attributes and values are like attribute values. The difference is in how they are validated: the keys of EA-maps need not be enumerated in advance, as is required with record attributes, and all values must have the same type. Moreover, EA-maps are treated by the validator as second-class, meaning valid usage scenarios are somewhat restricted.
 
 The main body of this RFC proposes that tag keys must be _literals_ when used to look up a tag's value, just like record attributes, but Alternative A generalizes this proposal to allow keys to be dynamically computed. Alternative A thus adds expressive power but also implementation cost.
 
 ### Basic example
 
-Here is a schema defining two entities, each of which contains entity tags.
+Here is a schema defining two entities, each of which contains embedded attribute maps (EA-maps, for short), used to implement tags.
 ```
 entity User = {
   jobLevel: Long,
@@ -32,7 +32,7 @@ entity Document = {
   policyTags: { ?: Set<String> },
 };
 ```
-The `User` and `Document` entities each have an attribute of type `{ ?: Set<String> }`, identifying tags whose values are sets of strings. The syntax for this type evokes a record, per the curly braces, but without named attributes; by providing just a `?` followed by a colon and a type, we indicate an unspecified number of optional attributes that have values of that type (here, `Set<String>`).
+The `User` and `Document` entities each have an attribute of type `{ ?: Set<String> }`, an EA-map, implementing tags whose values are sets of strings. The syntax for EA-map types evokes a record, per the curly braces, but without named attributes; by providing just a `?` followed by a colon and a type, we indicate an unspecified number of optional attributes that have values of that type (here, `Set<String>`).
 
 Here's a policy that conforms to this schema:
 ```
@@ -48,27 +48,27 @@ when {
     resource.policyTags["write"].containsAny(principal.authTags["write"]))
 };
 ```
-This policy states that for a `User` to carry out the _writeDoc_ action on a `Document`, the user must own the document, or else the user's job level must be at least 6 and the document and the user must each have a `write` tag in their attached tags, where at least one of the user's write-tag's values must be present in the document's write-tag's values.
+This policy states that for a `User` to carry out the _writeDoc_ action on a `Document`, the user must own the document, or else the user's job level must be at least 6 and the document and the user must each have a `write` tag in their EA-maps, where at least one of the user's write-tag's values must be present in the document's write-tag's values.
 
 ## Detailed design
 
-Neither the Cedar evaluator/authorizer, the JSON format for entities, nor the JSON or natural syntax for policies needs to change to support entity tags. This is because tags are represented as records and support the same operations. Only the Cedar schema format, the validator, and the schema-based entity parser need to change to account for a new `{ ?: T }` type.
+Neither the Cedar evaluator/authorizer, the JSON format for entities, nor the JSON or natural syntax for policies needs to change to support embedded attribute maps. This is because the notion of EA-map only arises at validation time, not evaluation time. At evaluation time, EA-maps are represented as records and support the same operations. In other words, for policy writers who do not use validation, there is no need for EA-maps: You can just use records. To support changes to the validator, the Cedar schema format, the validator, and the schema-based entity parser need to change to account for a new `{ ?: T }` type. 
 
 ### Schema
 
-Tags have type `{ ?: T }`, where `T` is the type of tag values. Only entity attributes can be given type `{ ?: T }`, and `T` cannot directly mention another tags type.
+EA-maps have type `{ ?: T }`, where `T` is the type of tag values. Only entity attributes can be given type `{ ?: T }`, and `T` cannot directly mention another EA-maps type.
 
-We extend schemas as follows to support the new tags type. Here's the natural syntax:
+We extend schemas as follows to support the new EA-maps type. Here's the natural syntax:
 ```
 Entity           := 'entity' Idents ['in' EntOrTyps] [['='] EntityRecType] ';'
 EntityRecType    := '{' [EntityAttrDecls] '}'
-EntityAttrDecls  := Name ['?'] ':' [Type | TagsType] [',' | ',' EntityAttrDecls]
-TagsType         := '{' '?' ':' Type '}'
+EntityAttrDecls  := Name ['?'] ':' [Type | EAMapsType] [',' | ',' EntityAttrDecls]
+EAMapsType       := '{' '?' ':' Type '}'
 Type             := PRIMTYPE | Path | SetType | RecType
 ```
-In essence: We alter the definition of entity types to include attributes with type `{ ?: T }`, where `T` is any type not mentioning another tags type. Not shown here are the productions for `RecType` and `AttrDecls`, which apply to records rather than entities. These productions are unchanged---normal records may not include attributes with `{ ?: T }` types.
+In essence: We alter the definition of entity types to include attributes with type `{ ?: T }`, where `T` is any type not mentioning another EA-maps type. Not shown here are the productions for `RecType` and `AttrDecls`, which apply to records rather than entities. These productions are unchanged---normal records may not include attributes with `{ ?: T }` types.
 
-The JSON syntax for schemas specifies tags as records with a `default` element, rather than an `attributes` element. Doing so leverages the analogy that tags are like records in which all attributes are optional, with the same type. Here's our introductory example schema in this format:
+The JSON syntax for schemas specifies EA-maps as records with a `default` element, rather than an `attributes` element. Doing so leverages the analogy that EA-maps are like records in which all attributes are optional, with the same type. Here's our introductory example schema in this format:
 ```
 "entityTypes": {
     "User" : {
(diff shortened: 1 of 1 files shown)


emina:
I like the main proposal, and I think it’s a good place to start.  If it turns out that we need dynamic attributes, we can always extend the implementation to support Alternative A, without breaking any existing policies.

emina (approved):

text/0068-entity-tags.md
```diff
@@ -0,0 +1,296 @@
...
+- Implementation PR(s): (leave this empty)
+
+## Timeline
+
+- Start Date: 2024-05-29
+- Date Entered FCP:
+- Date Accepted:
+- Date Landed:
+
+## Summary
```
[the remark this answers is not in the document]
john-h-kastner-aws: I agree that this RFC presupposes that "tags" are something users want on their data types. IMO it's kind of a surprising feature to support directly in a language.

text/0068-entity-tags.md
```diff
@@ -0,0 +1,296 @@
+# Entity tags
```
john-h-kastner-aws, line 1: I find "tags" to be a confusing name for this feature. When I read "tag", my first thought is that entities just have some set of strings attached, but tags was proposed are really extra attributes. I think it would be much less confusing if we could find a way to fold this into the existing notion of records.

mwhicks1: RFCs 66 and 27, linked from this RFC are proposals to reuse the existing notion of records either as "open records" (records have a default type that is associated with unspecified attributes), or map types (a different type given to records where all attributes are unspecified), respectively. Both of these RFCs have been rejected due to concerns about undecidable or inefficient logical encodings, precluding effective analyzability. By contrast, this RFC _is_ analyzable efficiently. Moreover, it turns out that those prior proposals were all motivated by wanting tags. The more general language feature (first-class records with default attribute types) might be pleasing at some level, but was not at all necessary for the customer use-case.

text/0068-entity-tags.md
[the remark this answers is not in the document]
mwhicks1: The motivation is to support tags, the style of AWS resource tags (key/value pairs, where values can have arbitrary type). The basic example in the summary is meant to serve as an exemplar of the motivation. Several customers have asked us for tagging support. We have suggested to them alternative workarounds, such as optional attributes or sets of key/value pairs, but these are inadequate for the reasons given in the RFC, and as discussed in prior RFCs 66 and 27, which this RFC supercedes.

text/0068-entity-tags.md
```diff
@@ -0,0 +1,296 @@
...
+
+Permissive validation supports subtyping, which we extend so as to allow `Tags` types to be treated co-variantly:
+```
+Tags<T> <: Tags<U>    iff T <: U
+```
+We might consider adding rules that permit subtyping between records and tags, but doing so introduces some complications with width subtyping.
+
+We could choose to grant `Tags` first-class status under permissive validation. For example, we could allow `Tags` values to appear anywhere, e.g., in `Set`s or records, and we could allow equality between `Tags` values. We choose not to do this for now, but may revisit in the future.
+
+## Drawbacks
```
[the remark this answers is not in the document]
mwhicks1: > the same expressive power proposed in this RFC (or something of similar power, and sufficient for the motivating use cases), but exposed to users at the syntax level 

We need to extend _entity_ types for feasible analysis. And that is precisely what we are doing at a syntactic level. If you are suggesting that we extend record types but encode them as entities under the covers, perhaps just for analysis, then _maybe_ that could work, but it would add a fair bit of complexity to the analysis code (especially for generating counterexamples). It would also be confusing/disappointing to users that default-typed records are more restricted than normal records. E.g., adding a default type to an existing record may cause validation to break, and could require a data-model rethink.

> maybe records are allowed to have some kind of default type, but if they do they are not allowed to have other attributes

This will not help the analyzability problem, since records with unknown/arbitrary attributes could appear in equality operations, which can't be feasibly encoded. (Extending an _entity_ with unknown attributes works because we never test entity _contents_ for equality, only entity references.)

> I think there's some available design space here and I think we should explore it as a community. 

Have you read RFCs 27 and 66 closely? These explore what I see as a big chunk of the available design space, but they have proved to be inadequate from an analysis perspective. You may be right that there are other options, but we _have_ thought seriously and hard about this problem for many months now, and we haven't come up with anything. Please present your ideas here!

text/0068-entity-tags.md
```diff
@@ -0,0 +1,296 @@
...
+```
+This policy states that for a `User` to carry out the _writeDoc_ action on a `Document`, the user must own the document, or else the user's job level must be at least 6 and the document and the user must each have a `write` tag in their attached tags, where at least one of the user's write-tag's values must be present in the document's write-tag's values.
+
+## Detailed design
+
+Neither the Cedar evaluator/authorizer, the JSON format for entities, nor the JSON or natural syntax for policies needs to change to support entity tags. This is because tags are represented as records and support the same operations. Only the Cedar schema format, the validator, and the schema-based entity parser need to change to account for a new `Tags<T>` type.
+
+### Schema
+
+Tags have type `Tags<T>`, where `T` is the type of tag values. Only entity attributes can be given type `Tags<T>`, and `T` cannot directly mention another `Tags` type.
```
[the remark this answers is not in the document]
john-h-kastner-aws: Regarding indirectly referencing tags: based on a later example I believe indirect references through an entity type are allows, so `Tags<UserTags>` where `UserTags` has to be a entity and not a common type.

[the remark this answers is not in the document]
mwhicks1: Yes, normal records cannot include a `Tags` type as an attribute, per the schema grammar. Also per the grammar, you cannot write `Tags<Tags<Long>>` but you could write `Tags<Foo>` where `Foo` is an entity type that has a `Tags<Long>` attribute.

> Can common types be `Tags` types, 

This is not in the proposal, but we could add it. I left it out because of `Tags`' second-class nature -- you can only use them for entity attributes, so I worry it's misleading to define a general type for them.

> Is `Tags<T>` allowed, where T is (or contains) a common type;

Yes. Per the grammar the `T` is a normal `Type` nonterminal.

> we didn't clarify if normal sets can, i.e., if `Set<Tags<T>>`

The grammar precludes this: The `T` in `Set<T>` must be a `Type` non-terminal, and `Type` does not include `Tags<>` types.

text/0068-entity-tags.md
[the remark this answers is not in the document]
mwhicks1: At one point @apandays had a motivating example he shared. Perhaps repeat in the main discussion thread, Abhi?

text/0068-entity-tags.md
john-h-kastner-aws: Starting from the idea that AWS style tags are a good feature, then I like this RFCs proposed implementation of tags in Cedar.

I suppose it must be a good feature to the extent that a customer wants to use it, but  I'm not fully convinced that beyond that they're a good general purpose feature customers outside the AWS ecosystem will want.

mwhicks1: What do you mean by "outside of the AWS ecosystem?" We have non-AWS customers who want tags. See https://github.com/cedar-policy/cedar-examples/tree/main/cedar-example-use-cases/tags_n_roles as an example of policies that use tags following one of the listed workarounds, which would be improved by this RFC (no need to change schemas on the fly, as you define new tags or roles). This example is based on a customer application.

max2me:
Let me see if I understand it correctly. Tags are like records except:

* Tag keys in policies must be string literals
* All values are of the same type
* They can only be at the "top" level of entities

Use cases this unlocks make a lot of sense. None of limitations concern me too much but I do wonder if we can somehow allow tags to be part of context to address use cases similar to RequestTags in IAM (I don't know how else someone would model that feature in Cedar).

mwhicks1:
> allow tags to be part of context to address use cases similar to RequestTags in IAM

You can define request tags as a special entity that is referenced in context. E.g., 

```
entity RequestTags = {
  Tags<Set<String>> tags,
};
...
action MyAction appliesTo {
  ...
  context: { tags: RequestTags }
};
```

Does that seem acceptable?

max2me:
> Does that seem acceptable?
Yes, it does

> > allow tags to be part of context to address use cases similar to RequestTags in IAM
> This is the case now, for simplicity, but Alternative A is doable later. Just more work.
If we are keeping door open here, then this seems like a great way to start.

aaronjeline:
I like the proposal, but echo Craig's concern that the syntax makes it look like a first class type.

aaronjeline (approved):
Happy now that we have a different syntax.

text/0068-entity-tags.md
```diff
@@ -0,0 +1,309 @@
...
+
+The main body of this RFC proposes that tag keys must be _literals_ when used to look up a tag's value, just like record attributes, but Alternative A generalizes this proposal to allow keys to be dynamically computed. Alternative A thus adds expressive power but also implementation cost.
+
+### Basic example
+
+Here is a schema defining two entities, each of which contains entity tags.
+```
+entity User = {
+  jobLevel: Long,
+  authTags: { Set<String>> },
```
cdisselkoen, line 28:
```suggestion
  authTags: { Set<String> },
```

text/0068-entity-tags.md
```diff
@@ -0,0 +1,309 @@
...
+```
+This policy states that for a `User` to carry out the _writeDoc_ action on a `Document`, the user must own the document, or else the user's job level must be at least 6 and the document and the user must each have a `write` tag in their attached tags, where at least one of the user's write-tag's values must be present in the document's write-tag's values.
+
+## Detailed design
+
+Neither the Cedar evaluator/authorizer, the JSON format for entities, nor the JSON or natural syntax for policies needs to change to support entity tags. This is because tags are represented as records and support the same operations. Only the Cedar schema format, the validator, and the schema-based entity parser need to change to account for a new `{ T }` type.
+
+### Schema
+
+Tags have type ` { T }`, where `T` is the type of tag values. Only entity attributes can be given type `{ T }`, and `T` cannot directly mention another tags type.
```
cdisselkoen, line 59:
```suggestion
Tags have type `{ T }`, where `T` is the type of tag values. Only entity attributes can be given type `{ T }`, and `T` cannot directly mention another tags type.
```

text/0068-entity-tags.md
```diff
@@ -0,0 +1,309 @@
...
+                "jobLevel" : {
+                    "type" : "Long"
+                },
+                "authTags" : {
+                    "type" : "Record",
+                    "attributes" : {
+                        "default": {
+                            "type" : "Set",
+                            "element": "String"
+                        }
```
cdisselkoen, line 87: What does the JSON for 
```
authTags: { default: Set<String> },
```
look like, and how does it compare to the JSON for
```
authTags: { Set<String> },
```
?

(Unless I'm missing something -- which is very possible -- both of these are valid in human-syntax schemas, assuming that we're inside an entity and not a record, because `default` is not a reserved attribute name; I assume this RFC isn't proposing that we reserve the attribute name `default`, which would potentially be disruptive.)

text/0068-entity-tags.md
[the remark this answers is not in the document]
cdisselkoen: My concern in this thread is sufficiently addressed by the new syntax proposed in the latest revision.  I'm happy to see a syntax that looks a lot more like an extension of the existing `entity` type, rather than an entirely new third kind of key-value mapping.

cdisselkoen (approved):
I'm much happier with the new syntax after the latest revision.  One question about the proposed schema-JSON format changes.

text/0068-entity-tags.md
mwhicks1: Good catch! I've updated with the most recent commit.

cdisselkoen: Fix looks great

shaobo-he-aws (approved):

cdisselkoen:
The final comment period (FCP) for this RFC is starting now, with intent to accept. The FCP will end 2024-07-17 at 10am PT / 1pm ET / 5pm UTC. Please add comments, and especially any objections, if you have any. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

jeffsec-aws:
I like the final revision of this RFC which will unlock new use cases.

text/0068-entity-tags.md
emina: + 1 for changing the name of this concept.  How about "Open Entities"?

emina (approved):

text/0068-entity-tags.md
mwhicks1: I tried to come up with an alternative name but didn't manage to. To me, an "open entity" is an entity with the feature proposed here, i.e., a thing called an _X_ which has type `{ ?: T }`. The presence of an attribute of that type makes the entity open, but the attribute _itself_ is not an open entity. This RFC calls an entity attribute of type `{ ?: T }` "entity tags".

If you can come up with a replacement name _X_ whereby you mostly search-and-replace "entity tags" with _X_ and it works then I'm happy to entertain it!

jeffsec-aws: Tags have a lot of similar meanings to what we want to achieve here without being affiliated to AWS. It always means a way to organize objects and we aimed here at passing those organizing metadata for policies evaluation.
Tags exists in devops world (Github, Gitlab, JIRA), IAC world (Terraform, Chef, Ansible), SaaS world (Salesforce), other security products... I am pretty sure I can go on :)

emina: > It always means a way to organize objects and we aimed here at passing those organizing metadata for policies evaluation.

I agree that tags are a common concept but, in my experience at least, "tag" usually means a single piece of meta-data that is attached to an object.  I haven't seen the word "tags" used to denote a set of key-value pairs that are part of the object's data, which is what we have here.   

> If you can come up with a replacement name X whereby you mostly search-and-replace "entity tags" with X and it works then I'm happy to entertain it!

How about "embedded maps" or "entity maps"?  The word "map" conveys what the underlying datastructure is a set of key-value pairs whose keys aren't known in advance, and the word "embedded" or "entity" signals that these aren't regular maps / first-class values.

mwhicks1: I like "embedded maps" but: The keys to the maps are always strings, so they are not as general as maps. Will that be confusing?

Another thought is "embedded attribute maps" to emphasize that such a map is from attributes to values (or, that it must always be stored in an attribute).

emina: I like "embedded attribute maps" too.

max2me:
FWIW All top 3 cloud providers (AWS, Azure, GCP) have a concept of "tags" that represents key-value pairs on a resource/entity. I

PR #69, opened 2024-06-07
Schema libraries

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/cdisselkoen/schema-libraries/text/0069-schema-libraries.md)

### DIFF
--- a/text/0069-schema-libraries.md
+++ b/text/0069-schema-libraries.md
@@ -1,0 +1,169 @@
+# Schema libraries
+
+## Related issues and PRs
+
+- Reference Issues: Inspired by [RFC 58], but includes new material not present in RFC 58
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2024-06-07
+
+## Summary
+
+Allow Cedar schemas to include/import "libraries" of definitions from remote
+URLs.
+
+This RFC does not propose that the Cedar team would build or maintain any such
+libraries; it only proposes the mechanism for importing libraries from URLs.
+
+## Basic example
+
+Human schema format:
+```
+import "https://raw.githubusercontent.com/cedar-policy/cedar-examples/release/3.2.x/cedar-example-use-cases/document_cloud/document_cloud.cedarschema"
+import "https://example.com/cedar_schemas/oidc.cedarschema"
+import "https://example.com/cedar_schemas_json/foobar.cedarschema.json"
+
+namespace "MyApp" {
+  ...
+}
+```
+
+JSON schema format:
+```
+{
+  "imports" : [
+    "https://raw.githubusercontent.com/cedar-policy/cedar-examples/release/3.2.x/cedar-example-use-cases/document_cloud/document_cloud.cedarschema",
+    "https://example.com/cedar_schemas/oidc.cedarschema",
+    "https://example.com/cedar_schemas_json/foobar.cedarschema.json"
+  ],
+  "MyApp": {
+    ...
+  }
+}
+```
+
+## Motivation
+
+Some data sources are common across many applications and useful to many Cedar
+users, either within the same organization or even across organizations.
+
+### 1. Within the same organization
+
+This RFC would allow an organization to define its own libraries of schema
+definitions, which could be reused across many different schemas (say, for
+different webapps owned by the organization).
+For instance, the organization might have common definitions of `User` or
+`Account` that apply in many different applications, and although those
+applications may not want to share entire Cedar schemas, with this RFC they
+could share just the definitions of `User` or `Account`, which could be defined
+once in a central location (in the same or separate libraries).
+
+### 2. Across organizations
+
+For another motivating example, consider identity providers (IdPs) which comply
+with the OpenID Connect standard (OIDC).
+The OIDC standard includes a list of attributes that exist on a user type; this
+is naturally declared as a Cedar entity type.
+With this RFC, anyone could provide a "library" representing Cedar definitions for
+OIDC types, and provide that library as a Cedar schema file at some URL; and then
+other Cedar users could use those definitions simply by importing the file from that URL.
+This would allow the Cedar community to gradually coalesce on the "best" way to
+represent an OIDC user in Cedar.
+
+### Motivations common to both scenarios
+
+In both of the above scenarios (within-organization and cross-organization),
+we obtain three key benefits:
+1. Saving each Cedar user the effort of writing common declarations themselves.
+  This facilitates code reuse in schemas, and makes it easier to get started
+  with Cedar.
+2. Providing a way to define common types and actions in a centralized way,
+  which ensures many schemas agree on the "correct" or "best practices"
+  definitions, and provides a single place to make updates if updates are
+  required.
+3. Facilitating code reuse for Cedar authorization calls, not just schemas.
+  When everyone shares a common definition of `OIDC::User`, the community could
+  conceivably converge on a reusable library function for, e.g., converting an
+  OIDC token into Cedar entity data.
+  This would further make it easier to get started with Cedar.
+  (Note that this RFC does not propose the Cedar team writing or maintaining
+  either library definitions for use in schemas or library functions for use in
+  Cedar authorization calls. It only points out that the community could
+  converge on these things.)
+
+## Detailed design
+
+Import statements are only allowed at the top level, outside of all namespace
+declarations.
+(In the future, another RFC could propose allowing imports in other positions.)
+
+The target of the import must be a raw file containing a valid Cedar schema.
+This schema may contain any definitions that are valid today in Cedar schemas,
+including namespaces, entity type definitions, common type definitions, and
+action declarations.
+The validator will essentially concatenate all of these definitions into the
+schema at the location of the `import` statement.
+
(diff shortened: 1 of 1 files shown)


text/0069-schema-libraries.md
```diff
@@ -0,0 +1,169 @@
...
+  Cedar authorization calls. It only points out that the community could
+  converge on these things.)
+
+## Detailed design
+
+Import statements are only allowed at the top level, outside of all namespace
+declarations.
+(In the future, another RFC could propose allowing imports in other positions.)
+
+The target of the import must be a raw file containing a valid Cedar schema.
```
blubben, line 102: Should they instead be a namespace? And then we leave providing the schemas from a location up to an upstream process?

text/0069-schema-libraries.md
```diff
@@ -0,0 +1,169 @@
...
+with the OpenID Connect standard (OIDC).
+The OIDC standard includes a list of attributes that exist on a user type; this
+is naturally declared as a Cedar entity type.
+With this RFC, anyone could provide a "library" representing Cedar definitions for
+OIDC types, and provide that library as a Cedar schema file at some URL; and then
+other Cedar users could use those definitions simply by importing the file from that URL.
+This would allow the Cedar community to gradually coalesce on the "best" way to
+represent an OIDC user in Cedar.
+
+### Motivations common to both scenarios
```
blubben, line 75: Could we also have mechanism to `use` a particular namespace? For instance, say I'm building an app where everything is under the AWS namespace. It would be nice if I could just `use AWS::IdentityCenter` and then policy authors only need to specify the sub-entity. e.g. `User` vs `AWS::IdentityCenter::User`.

text/0069-schema-libraries.md
blubben: Sort of like how Smithy does it: https://smithy.io/1.0/spec/core/idl.html

text/0069-schema-libraries.md
cdisselkoen: I can totally see this being useful, but I'm torn whether it should be part of this RFC or a separate RFC.

It would be harder to envision what a `use` would look like for policies (as opposed to schemas); I'm not sure if users will be happy with a `use` mechanism for schemas but not having one for policies?

blubben: Yeah, its most useful for policies IMO. Perhaps a separate RFC is better.

text/0069-schema-libraries.md
```diff
@@ -0,0 +1,169 @@
...
+perpetuity.)
+In particular, all valid JSON-format schemas must have `{` as their first
+non-whitespace character, and no valid human-format schemas have `{` as their
+first non-whitespace character.
+
+This RFC does not propose any mechanism for versioning libraries.
+Instead, it proposes that versioning would be done _above_ the Cedar layer,
+i.e., should be the responsibility of library authors.
+For instance, library authors could provide a different URL for different
+versions of their library, avoiding changing the contents of the URL for the
```
john-h-kastner-aws, line 122: We should provide something akin to [sub resource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#subresource_integrity_with_the_script_element) so users can guard against a changing import.

text/0069-schema-libraries.md
```diff
@@ -0,0 +1,169 @@
...
+## Detailed design
+
+Import statements are only allowed at the top level, outside of all namespace
+declarations.
+(In the future, another RFC could propose allowing imports in other positions.)
+
+The target of the import must be a raw file containing a valid Cedar schema.
+This schema may contain any definitions that are valid today in Cedar schemas,
+including namespaces, entity type definitions, common type definitions, and
+action declarations.
```
john-h-kastner-aws, line 105: Since the actions in a schema can impact authorization, a schema library adding or removing an action from a group could result in unexpected authorization decisions.

text/0069-schema-libraries.md
cdisselkoen: > Should they instead be a namespace?

I might be misunderstanding something -- this RFC as-written allows importing items in a namespace; in fact, all declared items in the library (import target) _must_ be in a (possibly empty) namespace, as otherwise it wouldn't be a valid Cedar schema.

You might be looking for a mechanism for "opening" a namespace, i.e., to use items from a namespace unqualified in the rest of the schema/policies?  (Akin to Rust's `use`.)  I think it makes sense to separate that into a different RFC; this one is about providing definitions from reusable libraries, while another one could be about making namespaced definitions usable without qualification.

(I realize that Java's `import`, and maybe `import`s in other languages, provide both functionalities -- introducing the library definitions and also making them available for use unqualified.  The `import` being proposed here provides only the first functionality.  For that reason maybe we should consider a different keyword to avoid confusion?)

text/0069-schema-libraries.md
cdisselkoen: Would you propose that libraries cannot include action declarations, and can only include entity type definitions and common type definitions?

john-h-kastner-aws: I don't that's worth it. Presumably, if you're using a schema library, you're consuming entity data from the IdP as well. The issue is just that you might not expect a schema to impact authorization. Sub-resource integrity plus documentation should mitigate this concern.

text/0069-schema-libraries.md
```diff
@@ -0,0 +1,178 @@
...
+implications which may not be acceptable for some users. Of course, those users
+could simply not use this feature.
+2. Cedar schemas would no longer be self-contained, in that a single (hopefully
+readable) file contains all of the relevant definitions. To mitigate this, we
+could provide a utility that displays the schema with all imports expanded.
+3. The Cedar Rust code would have to bring in substantial new dependencies, so
+that it could download libraries from remote URLs. To mitigate this for users
+who are concerned about this and don't need/want this feature (e.g., in
+resource-constrained environments, offline environments, Wasm, etc), we could
+put this RFC's functionality behind a Cargo feature, so that it and its
```
john-h-kastner-aws, line 146: Possibly local file system dependency enabled by default, with network dependencies behind a flag?

cdisselkoen: Your comment makes me realize I should definitely explicitly call out allowing local file system URIs in addition to URLs, which is not currently made explicit in the RFC. I'll amend

text/0069-schema-libraries.md
blubben: I was thinking about how I would use this for a specific use-case I have now. But after thinking a bit more, I think the answer is: I wouldn't. In my case, I want to define my schema dependencies in "packages" and then use a package/dependency manager tool to declare my dependencies and deal with retrieving the artifacts. All I really need from the Cedar library is that ability to load a schema from a set of files (you can already mostly do this but there are some sharp edges.) I do think the "opening" a namespace feature would be useful, but I agree thats a separate RFC.

But that raises the question: do we want "import" in the schema spec at all? Are we bringing unnecessary complexity into Cedar that is best left to some upstream system more suited to deal with it (e.g. a package manager)? Thoughts?

text/0069-schema-libraries.md
blubben: Take a look at what IonSchema does with their "SchemaAuthority": https://amazon-ion.github.io/ion-schema/docs/isl-2-0/spec. Essentially decoupling the loading of dependencies from their declaration. 

But I'm wondering if we really need/want to bring this into Cedar. (See my other comment.)

text/0069-schema-libraries.md
cdisselkoen: > you can already mostly do this but there are some sharp edges

Can you talk more about the sharp edges?  We'd like to track these as issues :)

> Are we bringing unnecessary complexity into Cedar that is best left to some upstream system more suited to deal with it (e.g. a package manager)?

Valid point; this kind of question was one reason for posting the RFC.  Interested to see if a majority of folks feel this way, or if there are folks who feel that `import` is solving a problem for them.  Probably this will become clear as this RFC generates more feedback.

blubben: The biggest sharp edge was I wanted to load each schema file as a SchemaFragment and then produce the final schema and validate. However, there were cases where it would fail to validate the fragment since it referenced things in other fragments. So I end up handing to convert all the schemas to the human readable format and append them together. Then load and validate that.

cdisselkoen: Interesting.  In my understanding, referencing things in other fragments should "just work", and if it doesn't, that is a bug.  If you have any specific examples, it would be great if we could get reproducers so we can fix the bugs.

text/0069-schema-libraries.md
cdisselkoen: Bringing here the summary of an offline conversation -- it makes a lot of sense to have some feature like this, but since including hashes directly in the schema file is a little ugly, maybe we'd want a separate lockfile, and at that point we might want to consider some separate Cedar.toml file (analogous to Cargo.toml) where you could declare all your dependencies, and possibly other global configuration.  Does the long-term path see Cedar having something more and more equivalent to `cargo`, complete with dependency/lockfile manager, package repository, etc?

PR #70, opened 2024-06-26
Disallow shadowing definitions in the empty namespace

cdisselkoen:
[Rendered](https://github.com/cedar-policy/rfcs/blob/cdisselkoen/disallow-empty-namespace-shadowing/text/0070-disallow-empty-namespace-shadowing.md)

### DIFF
--- a/text/0070-disallow-empty-namespace-shadowing.md
+++ b/text/0070-disallow-empty-namespace-shadowing.md
@@ -1,0 +1,155 @@
+# Disallow shadowing definitions in the empty namespace
+
+## Related issues and PRs
+
+- Reference Issues: Successor to [RFC 64]
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2024-06-26
+
+## Summary
+
+In schemas, disallow definitions of entity and common types that would shadow
+definitions of other entity/common types in the empty namespace.
+
+## Basic example
+
+Borrowing and slightly tweaking the example from [RFC 64]:
+
+```
+entity User {
+    email: String,
+};
+
+namespace Demo {
+    entity User {
+        id: String,
+    };
+
+    entity Account {
+        owner: User,
+    };
+}
+```
+
+Today, this schema is accepted, and `Account.owner` is assumed to be
+`Demo::User` and not the empty-namespace `User`.
+As noted in [RFC 64], there is no way for the user to indicate that they intend
+`Account.owner` to be the empty-namespace `User`.
+Rather than add syntax to enable this (as proposed in [RFC 64]), this RFC
+proposes to make this situation an error -- disallowing declaring a `User` type
+in the `Demo` namespace when one already exists in the empty namespace.
+
+## Motivation
+
+Name shadowing is complicated and confusing; schemas using name shadowing may
+look like they are doing one thing when they are actually doing another.
+For Cedar policy and schema authors, this can cause misunderstandings and make
+it more difficult to write correct schemas and policies.
+By disallowing shadowing in the most egregiously ambiguous cases -- when the
+name being shadowed was declared in the empty namespace -- we nudge users to
+make the schema clearer, for instance by renaming a type, or moving a
+conflicting empty-namespace declaration into a new namespace.
+
+## Detailed design
+
+We disallow shadowing a type in the empty namespace with a new declaration in a
+nonempty namespace.
+This applies regardless of whether the shadowed declaration is an entity or
+common type, and whether the shadowing declaration is an entity or common type.
+
+This RFC does not disallow shadowing altogether; in particular, it does not
+disallow shadowing a entity typename with a common typename in the same namespace,
+as originally proposed in [RFC 24] (see "Disambiguating Types in Custom Syntax",
+and in particular Rule 5 in that section):
+```
+// Allowed today, and still allowed after this RFC
+namespace NS {
+    entity User {
+        id: String,
+    }
+    type User = User;
+    // or, more confusingly, type User = String;
+}
+```
+
+This RFC also does not disallow shadowing the names of primitive or extension
+types, where you can still refer to the shadowed type using `__cedar`, again as
+explained in [RFC 24].
+Conceptually, primitive and extension types are not declared in the empty
+namespace; they are declared in the `__cedar` namespace, and implicitly imported
+as part of a prelude.
+That is how we could explain (e.g. in the Cedar docs) why the rule is different
+for shadowing entity/common types and shadowing primitive/extension types.
+(Internally, another reason for continuing to allow shadowing
+primitive/extension types, is that we want to be able to introduce new extension
+types without any possibility of breaking existing schemas.)
+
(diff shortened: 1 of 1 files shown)


mwhicks1 (approved):
Let's do it. For posterity in this RFC, you could even be more forceful in expressing the opinion that we want to be nudging users away from unqualified, top-level entity names.

aaronjeline:
> We could even codify that rule, requiring that all libraries have a name and all declarations in the library live in the corresponding namespace, or alternately, implicitly prefixing all library declarations with the library name as a toplevel namespace. Many existing package systems have similar rules or mechanisms.
This is addressing my primary worry.

aaronjeline (approved):

shaobo-he-aws:
LGTM.

shaobo-he-aws (approved):

text/0070-disallow-empty-namespace-shadowing.md
```diff
@@ -0,0 +1,155 @@
...
+name being shadowed was declared in the empty namespace -- we nudge users to
+make the schema clearer, for instance by renaming a type, or moving a
+conflicting empty-namespace declaration into a new namespace.
+
+## Detailed design
+
+We disallow shadowing a type in the empty namespace with a new declaration in a
+nonempty namespace.
+This applies regardless of whether the shadowed declaration is an entity or
+common type, and whether the shadowing declaration is an entity or common type.
```
john-h-kastner-aws, line 61: Would this apply in JSON schema as well? There's never any ambiguity between entity type and and common types in the JSON format, so they're not technically shadowing. Probably worth it to restrict the JSON for the sake of consistency, but the main motivation for this RFC doesn't apply.

cdisselkoen: This is a good point.  I can see arguments both ways.

For the current proposal:
* We probably want to avoid introducing more edge cases where converting JSON to human-syntax schema is difficult or fallible.  (Existing edge cases of this sort are discussed in [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md).)  
* This RFC is already technically a breaking change (although the break only applies in the particular edge case where users were relying on shadowing an empty-namespace declaration).  Making it "a little more" breaking by disallowing a few more edge-case JSON schemas probably doesn't make much of a difference to end-users.
* We end up in a more consistent state, where more similar rules apply to the human and JSON formats.

For amending the current proposal, to allow an entity type with the same name as an empty-namespace common type, and vice versa, in the JSON format only:
* You are right that there's no technical reason we can't allow this; these declarations do not conflict and are not ambiguous in the JSON format, which distinguishes entity typenames from common typenames.
* It makes this RFC a little less breaking, by avoiding breaking users who were already doing this in a JSON schema.

mwhicks1: We have a rule (guideline?) that you can convert back-and-forth between JSON and human readable formats. If you allow this in JSON but not human-readable we break the tenet. IMO we should change the JSON.

khieta (approved):
Lgtm. But I would like to see a note in the RFC about the JSON vs. human-readable schema syntax.

My opinion is that we should enforce the same restrictions on both, but I would still approve of this RFC even if we only enforced this restriction on the human-readable syntax.

john-h-kastner-aws (approved):

text/0070-disallow-empty-namespace-shadowing.md
cdisselkoen: Based on this thread and @khieta's review below, I will amend the RFC to make clear we will disallow this in the JSON as well.

In particular I'm referencing these comments:

@john-h-kastner-aws: "Probably worth it to restrict the JSON for the sake of consistency"
@cdisselkoen: "We end up in a more consistent state, where more similar rules apply to the human and JSON formats"
@mwhicks1: "If you allow this in JSON but not human-readable we break the tenet. IMO we should change the JSON"
@khieta: "My opinion is that we should enforce the same restrictions on both"

cdisselkoen: Added text at the end of the `Detailed design` section

andrewmwells-amazon (approved):

cdisselkoen:
The final comment period (FCP) for this RFC is starting now, with intent to accept. The FCP will end 2024-07-17 at 10am PT / 1pm ET / 5pm UTC. Please add comments, and especially any objections, if you have any. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

text/0070-disallow-empty-namespace-shadowing.md
```diff
@@ -0,0 +1,166 @@
...
+To address this drawback, we should guide library authors to not declare library
+types in the empty namespace (and instead, e.g., declare them in a namespace
+named after the library).
+We could even codify that rule, requiring that all libraries have a name and all
+declarations in the library live in the corresponding namespace, or alternately,
+implicitly prefixing all library declarations with the library name as a
+toplevel namespace.
+Many existing package systems have similar rules or mechanisms.
+
+## Alternatives
```
cdisselkoen, line 136: A note: This RFC does not say anything about shadowing for _action_ declarations.  In fact, for actions (unlike for entity/common types) we already have syntax that allows you to explicitly reference empty-namespace declarations:

```
// refers to `action Group` in the current namespace,
// or, after #579 is fixed, `action Group` in the empty namespace
// if there is no `action Group` in the current namespace
action A in [Group];

// Same as the first example; this syntax allows for action IDs that 
// contain spaces or special characters etc.
action A in ["Group"];

// refers to `action Group` in another specified namespace
action A in [OtherNS::Action::"Group"];

// refers to `action Group` in the empty namespace specifically, 
// even if there is also an `action Group` in the current namespace
action A in [Action::"Group"];
```

At least, this is my current understanding. Happy to be corrected if I'm wrong.

john-h-kastner-aws: Based on cedar-policy/cedar#150 I'd expect that the last option would also be qualified with the current namespace

cdisselkoen: Actually, from looking at the code, I suspect the fourth example is also equivalent to the first two examples.  (Not sure if that's what we want.  Also not sure if changing it is worth a breaking change though.)

cdisselkoen: oh didn't see your comment @john-h-kastner-aws when I wrote that.  So we're in agreement.

In that case, the situation for actions is more similar to the situation for entity/common types.  Should we disallow action shadowing as well?  (I.e., disallow declaring `action Group` in both the empty namespace and a non-empty namespace?)

cdisselkoen: Updated the RFC to disallow this shadowing for action names as well.

cdisselkoen:
Merging this, having addressed the concern that was raised during FCP.

PR #71, opened 2024-06-27
Trailing Commas

aaronjeline:
[Rendered](https://github.com/cedar-policy/rfcs/blob/trailing-commas/text/0071-trailing-commas.md)


text/0071-trailing-commas.md
```diff
@@ -0,0 +1,170 @@
...
+- Reference Issues: https://github.com/cedar-policy/cedar/issues/1024
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2024-06-26 
+
+## Summary
+
+The Cedar grammar (both policies and schemas) should accept trailing
```
shaobo-he-aws, line 14: Human-readable schemas do allow trailing commas.

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,170 @@
...
+```
+Idents    := IDENT {',' IDENT}
+```
+becomes:
+```
+Idents    := IDENT {',' | ',' IDENT}
+```
+
+### Cedar Policy Formatter
+This RFC proposes a change to the style guide, that having a trailing
```
shaobo-he-aws, line 132: I think Rust doesn't do this. I actually prefer this way.

aaronjeline: image

text/0071-trailing-commas.md
aaronjeline: They do in some places, not in others. This proposes to make it so it's allowed in every possible spot.

text/0071-trailing-commas.md
shaobo-he-aws: TIL. I still think it's unnecessary.

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,170 @@
...
+* Go
+* JavaScript
+* OCaml
+
+(Frustratingly: not JSON!)
+
+In addition, it will make our schema grammar and policy grammar more
+consistent, as the schema grammar allows trailing commas in many places.
+
+## Detailed design
```
john-h-kastner-aws, line 65: It'd be good to list out an example for every expression where we'd support it before digging in to the grammar.

What comes to mind is:

```
permit(principal, action, resource, )'
[1, 2, ]
{a: 1, b: 2, }
ip("0.0.0.0", )
principal.tags.contains("x", )
```

Anything else?

Annotations feel like they could also take a trailing comma if we wanted to stretch the limits of this RFC. Something like `@id("permit-admin", )`

john-h-kastner-aws (approved):
Support. We definitely need to bring the schema and policy into alignment, and trailing commas are just nicer.

text/0071-trailing-commas.md
aaronjeline: My only argument against doing it for annotations is it implies you _could_ supply more than one arg: which you can't.

text/0071-trailing-commas.md
aaronjeline: Are you saying that you think the formatter should prefer the other way (ex: `[1,2,3,]` would auto-style to `[1,2,3]`) or that the formatter should not modify what the user writes. (ex: `[1,2,3,]` would auto-style to `[1,2,3,]` and `[1,2,3]` would auto-style to `[1,2,3]`)

john-h-kastner-aws: It looks like `rustfmt` adds a trailing comma when it wants to split something across multiple lines. That's a good behavior IMO.

Adding a trailing comma in `[1,2]` is kind of ugly, but desirable in the record split across multiple lines. Also, for method calls we really don't want to add the trailing comma (in fact, no Cedar methods or functions take more than one argument)

aaronjeline: Good callout on method calls

shaobo-he-aws: I meant option C.

cdisselkoen: Chiming in to agree with @john-h-kastner-aws: my preference is to add the trailing comma when split across multiple lines, remove the trailing comma when not split across multiple lines.  But I don't feel strongly, and @aaronjeline's Option C is also fine with me.

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,187 @@
...
+## Basic example
+
+Currently, this is a parse error:
+```
+permit(principal, action, resource) when {
+    { foo : 3, bar : 2, } == context  
+};
+```
+This RFC would make it not a parse error. 
+It would also change this is several places beyond records.
```
cdisselkoen, line 26:
```suggestion
It would also change this in several places beyond records.
```

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,187 @@
...
+RecInits ::= (IDENT | STR) ':' Expr {',' | ',' (IDENT | STR) ':' Expr}
+```
+
+### Cedar Schema Grammar
+The schema grammar already allows trailing commas in several places
+(such as record types and entity lists), but not everywhere:
+
+We want to allow trailing commas in the following expressions:
+1. membership lists: `entity Photo in [Account, Album,];`
+2. Multi-entity declarations: `entity Photo,Video;`
```
cdisselkoen, line 116:
```suggestion
2. Multi-entity declarations: `entity Photo, Video, ;`
```

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,187 @@
...
+3. Record types: `{ foo : Long, bar : String, }` (currently allowed)
+4. Multi-action declarations: `action foo,bar,;`
+5. Apply specs: `appliesTo { principal : ["A"], resource: ["B"],"`
+   (currently allowed)
```
cdisselkoen, line 120: What about `principal : ["A", "C", ], resource : ["B", "C", ],`"?

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,187 @@
...
+* Trailing commas are potentially confusing.
+* Policies produced by the new formatter won't parse with old parsers
+
+## Alternatives
+
+* Keep status quo: this is a non critical change
+
+## Unresolved questions
+
+* We could forbid the trailing the comma after `resource`.
```
cdisselkoen, line 187: I'm in favor of allowing it, particularly since scopes are so often split across multiple lines

cdisselkoen (approved):
Strongly approve of this change.  For the formatter, I do prefer a different behavior than currently proposed in the RFC, as I noted in the comment thread.

text/0071-trailing-commas.md
khieta: I'm also in favor of allowing it. No need to add an extra special case to worry about.

text/0071-trailing-commas.md
```diff
@@ -0,0 +1,187 @@
...
+```
+Scope ::= Principal ',' Action ',' Resource ','?
+```
+
+```
+EntList ::= Entity {',' Entity}
+```
+becomes:
+```
+EntList ::= Entity {',' | ',' Entity} ','?
```
khieta, line 91:
```suggestion
EntList ::= Entity {',' | ',' Entity}
```

I think you don't want this extra `,?` because it would allow extra commas (e.g., `User::"alice", User::"bob",,`). Same comment for a few other places in the rfc.

khieta (approved):
Approve of this idea in general. Left some minor comments on the new grammar rules & formatting discussion.

text/0071-trailing-commas.md
john-h-kastner-aws: lol, i think grammar as written would allow as many commas as you like between entities `User::"alice",,,,,,,,,,,,,User::"bob"`

chrnorm (approved):

andrewmwells-amazon (approved):

jmccarthy (approved):
Consensus from StrongDM is 👍 for trailing commas

shaobo-he-aws (approved):

PR #72, opened 2024-06-29
Object-Entity Mapping (OEM)

shaobo-he-aws:
[Rendered](https://github.com/USERNAME/REPONAME/blob/BRANCHNAME/text/FILENAME.md)

### DIFF
--- a/text/0072-oem.md
+++ b/text/0072-oem.md
@@ -1,0 +1,97 @@
+# Object-Entity Mapping (OEM)
+
+## Related issues and PRs
+
+- Reference Issues: https://github.com/cedar-policy/cedar/issues/986
+- Implementation PR(s): (leave this empty)
+
+## Timeline
+
+- Started: 2024-06-28
+
+## Summary
+
+Similar to [Object-Relational Mapping (ORM)](https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping), this RFC proposes a macro-based conversion between Rust objects and entities.
+
+## Basic example
+
+The following Rust code defines how type `Team` and type `Member` map to entity type `Example::Team`.
+
+```Rust
+#[derive(Debug, Clone, Cedar)]
+#[cedar(rename="Example::Team")]
+struct Team {
+  #[cedar(eid)]
+  id: String,
+  members: HashSet<Member>
+}
+
+#[derive(Debug, Clone, Cedar)]
+#[cedar(transparent)]
+struct Member {
+  name: String,
+  #[cedar(rename="region")]
+  location: Option<String>,
+}
+```
+An object `Team { id: "A", members: { Member { name: "alice", location: Some("West")}, Member { name: "bob", location: None}}}` is converted to the following entities in JSON,
+
+```JSON
+[
+ {
+    {
+        "uid": { "type": "Team", "id": "A"},
+        "attrs": {
+            "members": [
+                {
+                    "name": "alice",
+                    "location" "West"
+                },
+                {
+                    "name": "bob"
+                }
+            ]
+        },
+        "parents": []
+    }
+ }
+]
+```
+## Motivation
+
+There currently do not exist any convenient methods to programmatically convert Rust objects to entities or vice versa. The status quo of converting a Rust object to an entity is to create an entity without any attributes and then insert attributes. The whole process is akin to that of *manually* serializing Rust objects to JSON, the latter of which is automated by serde. Users should be able to do the same for Rust object to entity conversion. This RFC addresses this need using Rust derive macros.
+
+Moreover, OEM can potentially facilitate the adoption of Cedar on applications using databases. For instance, developers can leverage ORM to seamlessly convert results of database queries into Rust objects and then into entities. That is, developers are able to conveniently construct authorization requests by fetching data from databases.
+
+## Detailed design
+
+Given limited knowledge about Rust macros the proposer has, the design details are subject to significant changes.
+
+The macros over a Rust type eventually expand to implementations of `Into<Entities>`. In the meantime, they also give rise to implementations of `Into<Value>` for this type. By default, the implementation of `Into<Value` for a type produces an entity `uid` unless the type is annotated with the `#[cedar(transparent)]` clause, where it produces a record.
+
+The implementation of `Into<Entities>` for any type annotated with the `#[cedar(transparent)]` clause produces an empty entity slice. Those for types without such annotations construct an entity by iterating its fields. If a field type corresponds to an entity type, the entity slice derived from the field is merged into the final resulting entity slice. That is the reason why a Rust object converts to an entity slice instead of a single entity.
+
+### Value mapping
+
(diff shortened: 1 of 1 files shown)


text/0072-oem.md
```diff
@@ -0,0 +1,97 @@
...
+
+Moreover, OEM can potentially facilitate the adoption of Cedar on applications using databases. For instance, developers can leverage ORM to seamlessly convert results of database queries into Rust objects and then into entities. That is, developers are able to conveniently construct authorization requests by fetching data from databases.
+
+## Detailed design
+
+Given limited knowledge about Rust macros the proposer has, the design details are subject to significant changes.
+
+The macros over a Rust type eventually expand to implementations of `Into<Entities>`. In the meantime, they also give rise to implementations of `Into<Value>` for this type. By default, the implementation of `Into<Value` for a type produces an entity `uid` unless the type is annotated with the `#[cedar(transparent)]` clause, where it produces a record.
```
john-h-kastner-aws, line 70: > produces an entity `uid` unless the type is annotated with the `#[cedar(transparent)]` clause

I would rather see this made explicit. E.g., `#[cedar(record)]` or `#[cedar(entity)]`, perhaps with the entity case being assumed by default.

This could also support a nice syntax like `#[cedar(long)] struct MyLong(i32)` for wrapper types that might exist.

khieta:
Thanks for the writeup @shaobo-he-aws! I think this looks like a useful addition to Cedar, but I disagree that it should be an RFC. It's a change to the Rust API rather than the Cedar language, so I think it should remain as an issue in the `cedar` repository. Maybe you could move the details here into [cedar#986](https://github.com/cedar-policy/cedar/issues/986)?

How do others feel about an RFC vs. issue here?

shaobo-he-aws:
> Thanks for the writeup @shaobo-he-aws! I think this looks like a useful addition to Cedar, but I disagree that it should be an RFC. It's a change to the Rust API rather than the Cedar language, so I think it should remain as an issue in the `cedar` repository. Maybe you could move the details here into [cedar#986](https://github.com/cedar-policy/cedar/issues/986)?
> 
> How do others feel about an RFC vs. issue here?

I'm afraid that moving it to an issue reduces discussion and eventually makes it buried in the backlog list.

cdisselkoen:
> It's a change to the Rust API rather than the Cedar language, so I think it should remain as an issue

We've previously done RFCs for other major changes that aren't technically changes to the Cedar language itself, for example, [RFC 32](https://github.com/cedar-policy/rfcs/blob/main/text/0032-port-formalization-to-lean.md) and arguably [RFC 8](https://github.com/cedar-policy/rfcs/pull/8).  Similarly, I think there is consensus that partial evaluation itself would have been an RFC except that its development predates the RFC process.  I think that this is a major enough change that an RFC is warranted, and particularly since @shaobo-he-aws has already gone to the trouble of writing one up, I think it makes sense to treat it as an RFC.  If we didn't already have an RFC writeup and the primary author was pushing hard to keep this an issue instead of an RFC, I could see us maybe making a different decision on RFC vs issue, but since in this case the primary author is welcoming the RFC process, I see no reason to deny that.

> I'm afraid that moving it to an issue reduces discussion and eventually makes it buried in the backlog list.

Even getting an RFC accepted (into `Active` state) does not promise any particular implementation priority, and the resulting implementation issue could easily get no discussion and become buried in the backlog, like, for example, [RFC 53](https://github.com/cedar-policy/cedar/issues/981), where it took almost 3 months for the implementation issue to even be made.  The RFC process isn't a magic way to get implementation prioritized, vs any other issues in the backlog.  See the [RFC life-cycle](https://github.com/cedar-policy/rfcs/blob/main/README.md).

B-Lorentz:
What about a trait that would allow an user-provided struct to be borrowed as an entity/context?

Something like
```rust
trait AsEntity {
    pub fn uid(&self) -> EntityUID;

    pub fn get(&self, attr: &str) -> Option<&AsPartialValue>;
   
    
    pub fn get_tag(&self, tag: &str) -> Option<&AsPartialValue>;

   pub fn ancestors -> Impl<Iterator....
   // whatever other things the evaluator could want to know.
}

```
or the same with Result return types.

Basically you could implement this trait if you have a type that you know its representing an entity, whatever its internal structure is (it could be a serde_json::Value, your own struct, whatever), and the evaluator could use it to try to view it as one, without necessarily  needing to allocate an actual owned 'Entity' structure.

At first, this api would be provided by the evaluator in fact aggressively converting to Entity, but later the conversion could be optimized away.

This wouldn't of course matter for use over the network, but could make it more seamless to integrate cedar directly to a rust codebase.

shaobo-he-aws:
> What about a trait that would allow an user-provided struct to be borrowed as an entity/context?
> 
> Something like
> 
> ```rust
> trait AsEntity {
>     pub fn uid(&self) -> EntityUID;
> 
>     pub fn get(&self, attr: &str) -> Option<&AsPartialValue>;
>    
>     
>     pub fn get_tag(&self, tag: &str) -> Option<&AsPartialValue>;
> 
>    pub fn ancestors -> Impl<Iterator....
>    // whatever other things the evaluator could want to know.
> }
> ```
> 
> or the same with Result return types.
> 
> Basically you could implement this trait if you have a type that you know its representing an entity, whatever its internal structure is (it could be a serde_json::Value, your own struct, whatever), and the evaluator could use it to try to view it as one, without necessarily needing to allocate an actual owned 'Entity' structure.
> 
> At first, this api would be provided by the evaluator in fact aggressively converting to Entity, but later the conversion could be optimized away.
> 
> This wouldn't of course matter for use over the network, but could make it more seamless to integrate cedar directly to a rust codebase.
> 
> One might chose to then implement this manually, and the macros originally mentioned in the rfc would implement this too, instead of Into

This is a good idea. Do you mind writing up a RFC for it?

B-Lorentz:
> > What about a trait that would allow an user-provided struct to be borrowed as an entity/context?
> > Something like
> > ```rust
> > trait AsEntity {
> >     pub fn uid(&self) -> EntityUID;
> > 
> >     pub fn get(&self, attr: &str) -> Option<&AsPartialValue>;
> >    
> >     
> >     pub fn get_tag(&self, tag: &str) -> Option<&AsPartialValue>;
> > 
> >    pub fn ancestors -> Impl<Iterator....
> >    // whatever other things the evaluator could want to know.
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or the same with Result return types.
> > Basically you could implement this trait if you have a type that you know its representing an entity, whatever its internal structure is (it could be a serde_json::Value, your own struct, whatever), and the evaluator could use it to try to view it as one, without necessarily needing to allocate an actual owned 'Entity' structure.
> > At first, this api would be provided by the evaluator in fact aggressively converting to Entity, but later the conversion could be optimized away.
> > This wouldn't of course matter for use over the network, but could make it more seamless to integrate cedar directly to a rust codebase.
> > One might chose to then implement this manually, and the macros originally mentioned in the rfc would implement this too, instead of Into
> 
> This is a good idea. Do you mind writing up a RFC for it?

The RFC is here:
https://github.com/cedar-policy/rfcs/pull/93

PR #76, opened 2024-08-01
entity slice validation

mwhicks1:
Validating policies for entity slices using levels

[Rendered](https://github.com/cedar-policy/rfcs/blob/entity-slice-validation/text/0076-entity-slice-validation.md)

### DIFF
--- a/text/0076-entity-slice-validation.md
+++ b/text/0076-entity-slice-validation.md
@@ -78,6 +78,8 @@ Currently, Cedar provides no support for slicing. Users are required to either b
 
 Additionally, there are some application scenarios where the actors writing the policies and the people performing slicing are not the same people, and thus the slicers need to impose some kind of restrictions on policy authors.
 
+This RFC has a close connection to [RFC 74: Entity Slicing using Data Tries](https://github.com/cedar-policy/rfcs/pull/74), which also seeks to provide a sound foundation for entity slicing. We discuss the relationship between them in the Alternatives below; ultimately, we believe they are complementary and should coexist.
+
 ## Detailed design
 
 ### Entity dereferences
@@ -258,6 +260,14 @@ Level-based slicing is coarse-grained: *all* attributes up to the required level
 Implementing levels is a non-trivial effort: We must update the Rust code for the validator of policies and entities, and the Lean code and proofs. We also must provide code to perform level-based slicing.
 
 ## Alternatives
+
+### Alternative: Entity Slicing using Data Tries
+
+[RFC 74, Entity Slicing using Data Tries](https://github.com/cedar-policy/rfcs/pull/74) proposes the idea of an _entity manifest_ to ease the task of entity slicing. This manifest is a trie-based data structure that expresses which entity attributes and ancestors are needed for each possible action. This data structure is strictly more precise than the concept of level, meaning that it may require less entity data to be provided. For example, the first part of our Basic Example in this RFC (policies 1, 2, and 4 of TinyTodo), the determined entity manifest would indicate that for the `GetList` action, you would need to provide `resource.owner`, `resource.readers`, `resource.editors` and the ancestors of `principal`. This is less data than would be retrieved by the slicing algorithm proposed here: These are level-1 valid policies, so the generic algorithm would acquire the same data given in the manifest but also more, e.g., the ancestors of `resource` and the attributes of `principal`.
+
+Entity manifests are less effective as a _prescriptive_ mechanism that aims to limit the shape of allowable policies. For example, a Cedar policy and entity storage service might like to disallow uploading policies whose entity data is expensive to retrieve in the worst case. With levels, such a service can upper-bound this work by blocking schemas with a level greater than a threshold, say 3. Validation will then ensure that uploaded policies respect this upper bound. Manifests are unwieldy as a specification mechanism, since they are specialized to particular entity definitions and name particular types and attributes, of which there could be many. They'd also have to be updated as types evolved over time, while a level-based specification is less likely to change.
+
+We believe that ultimately level-based validation and entity manifests should coexist. Level-based validation is used to bound entity retrieval work and prevent pathological policies, while entity manifests are used to more efficiently define the needed entity data. We could imagine a deployment strategy in which we accept this RFC and perform generic entity slicing as described here, and then later implement entity manifests, and start performing slicing using those instead.
 
 ### Alternative: Level as a validation parameter, not in the schema
 
--- a/text/0076-entity-slice-validation.md
+++ b/text/0076-entity-slice-validation.md
@@ -96,7 +96,7 @@ Also note that by “dereference” above, we are only referring to the given op
 
 ### Entity roots
 
-The variables `principal`, `action`, and `resource` are *entity roots*. (Entity literals are *not* entity roots, and operations on such literals are limited, as described below.) They directly name entities from a request that could be dereferenced in a policy. A `context` is not an entity root directly; rather, any attributes it (transitively) contains that have entity type are considered roots. For example, consider the following schema:
+The variables `principal`, `action`, and `resource` are *entity roots*. (Entity literals are *not* entity roots, and operations on such literals are limited, as described below. Slots and the `Unknown` entity are treated as entity literals because slots will be replaced by literals and `Unknown` cannot be dereferenced.) They directly name entities from a request that could be dereferenced in a policy. A `context` is not an entity root directly; rather, any attributes it (transitively) contains that have entity type are considered roots. For example, consider the following schema:
 
 ```
 entity User {
--- a/text/0076-entity-slice-validation.md
+++ b/text/0076-entity-slice-validation.md
@@ -1,0 +1,430 @@
+# Entity Slice Validation
+
+## Related issues and PRs
+
+* Reference Issues:
+* Implementation PR(s):
+
+## Timeline
+
+* Started: 2024-08-01
+
+## Summary
+
+This RFC introduces "Entity Slicing Validation" (ESV), which consists of
+
+1. An additional validation check, and
+2. A generic entity slicing algorithm
+
+The two are connected by the concept of a "level," which defines the depth from root variables (like `principal` and `resource`) at which entity dereferencing operations may safely take place in policies. The validation check ensures that policies never dereference beyond the given level. The slicing algorithm determines what entities are needed to safely decide an authorization request when evaluating policies valid at a given level.
+
+## Basic example
+
+Consider the following [policies](https://github.com/cedar-policy/cedar-examples/blob/release/3.2.x/tinytodo/policies.cedar), taken from the [TinyTodo example application:](https://github.com/cedar-policy/cedar-examples/tree/release/3.2.x/tinytodo)
+
+```
+// Policy 1: A User can perform any action on a List they own
+permit (
+  principal,
+  action,
+  resource is List
+)
+when { resource.owner == principal };
+
+// Policy 2: A User can see a List if they are either a reader or editor
+permit (
+    principal,
+    action == Action::"GetList",
+    resource
+)
+when { principal in resource.readers || principal in resource.editors };
+
+// Policy 4: Admins can perform any action on any resource
+permit (
+  principal in Team::"Admin",
+  action,
+  resource
+);
+```
+
+We suppose (per the [schema](https://github.com/cedar-policy/cedar-examples/blob/release/3.2.x/tinytodo/tinytodo.cedarschema), not shown) that `principal` always has entity type `User`, and `resource` always has entity type `List`. These policies are valid at level 1, meaning that variables `principal`, `action`, `resource` and `context` are only *directly* dereferenced, i.e., any entities they reference are not themselves dereferenced.
+
+For example, in policy 1, the expression `resource.owner == principal` directly dereferences `resource` by accessing the contents of its `owner` attribute. In policy 2, expression `principal in resource.readers` directly dereferences `resource` to retrieve the contents of its `readers` attribute, and also directly dereferences `principal` to retrieve its ancestors, to see whether one might be equal to the contents of `resource.readers`. Policy 4 also directly dereferences `principal` to access its ancestors, to see if one is `Team::"admin"`.
+
+Because these policies are valid at level 1, applications submitting authorization requests only need to provide entity data for entities provided as variables, and nothing else. For example, say we want to evaluate the request (`User::"Aaron"`, `Action::"GetList"`, `List::"Objectives"`, `{}`). Then we would only need to provide entity data (direct attributes and ancestors) for `User::"Aaron"`, `Action::"GetList"`, and `List::"Objectives"`. There is no need to provide data for entities referenced from these entities. For example, suppose the `readers` attribute of `List::"Objectives"` is `Team::"interns"`; there is no need to provide data for `Team::"interns"` because we know it will never, itself, be dereferenced when evaluating these policies. There is also no need to provide entity data for other entities directly mentioned in policies, e.g., `Team::"Admin"`.
+
+Now suppose we wanted to extend our TinyTodo policies with the following:
+
+```
+// Policy 6: No access if not high rank and at location DEF, 
+// or at resource's owner's location
+forbid(
+    principal,
+    action,
+    resource is List
+) unless {
+   principal.joblevel > 6 && principal.location like "DEF*" ||
+   principal.location == resource.owner.location
+};
+```
+
+This policy does *not* validate at level 1. That is because the expression `resource.owner.location` is only valid at level 2: It requires dereferencing `resource` to retrieve the entity in the `owner` attribute, and then requires dereferencing that entity to obtain the contents of its `location` attribute. As a result, if we only provided the `principal` and `resource` entity data, as described above, evaluating policy 6 could produce an evaluation error since data may not be provided for `resource.owner`.
+
+All of the Cedar example policy sets validate with level-based validation, either at level 1 or 2; see the Appendix for details.
+
+## Motivation
+
+Currently, Cedar provides no support for slicing. Users are required to either be inefficient and produce their entire entity store as a slice, or derive their own ad-hoc slicing algorithm. We'd like to provide an out-of-the-box solution that addresses many user's needs. Entity slicing in general is a problem very tied to the specifics of application/deployment. It's likely impossible to have a fully general solution. This RFC is attempting to provide a solution for entity slicing that works for the 80% case, while acknowledging that it won't work for everyone. 
+
+Additionally, there are some application scenarios where the actors writing the policies and the people performing slicing are not the same people, and thus the slicers need to impose some kind of restrictions on policy authors.
+
+## Detailed design
+
+### Entity dereferences
+
+An "Entity Dereference" is any operation that requires directly accessing an entity’s data. The following is an exhaustive list of entity dereferencing operations
+
+1. `e1 in e2` dereferences `e1` (but not `e2`)
+2. `e1.foo` dereferences `e1` iff `e1` is an entity
+3. `e1 has foo` dereferences `e1` iff `e1` is an entity
+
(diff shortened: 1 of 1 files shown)


oflatt:
Nice idea, and much simpler than entity manifests.
We should add a note about how this is related to #74. Given a type-annotated entity manifest, calculating the level is easy

shaobo-he-aws (approved):
Pretty cool.

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,430 @@
...
+    context: {
+        admin: User,
+        building: { location: Long, ITDeptHead: User },
+        foo: { inner: User },
+        bar: { inner: User }
+    }
+};
+```
+
+Here, `principal`, `action`, and `resource` are entity roots (as usual), as are:
```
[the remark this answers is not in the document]
mwhicks1: My reason for treating `context` this way is that you sometimes need four entities in a request, and the `context` is the place to put the fourth one. For example:
```
permit(principal, action == Action::"view", resource) when {
  context.ticket.target == resource &&
  principal in context.ticket.viewers
};
```
In general, in Cedar when you want to write "there exists a Foo such that ..." you have to ask the requester to provide the Foo as an argument. Here the Foo is a ticket. And you need the ticket's attributes.

mwhicks1:
> Nice idea, and much simpler than entity manifests. We should add a note about how this is related to #74. Given a type-annotated entity manifest, calculating the level is easy

Thanks! I've updated the RFC to compare the two, landing on the position that we should ultimately have both.

text/0076-entity-slice-validation.md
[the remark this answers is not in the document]
mwhicks1: Another example is something like:
```
permit(principal, action == Action::"moveFile", resource) when {
  principal in resource.readers &&
  principal in context.target.writers
};
```
where `resource` is a file, and `context.target` contains the directory to move the file to.

D-McAdams:
Thanks for adding the cross-link to Entity Manifests in https://github.com/cedar-policy/rfcs/pull/74.
Agree that they are best as a pair. Entity Manifests give instructions on what to load, and Entity Slice Validation (this RFC) ensures the instructions are practical by capping the complexity of incoming policies. 

Also agree the alternatives in this RFC are less necessary when both techniques are used together.

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+// Fails because a Team-typed entity is on the LHS of `in`,
+//   but entity Team has no legal RHS expressions per the schema
+permit(principal,action == Action::"GetList",resource)
+when { resource.editors in Team::"admin" };
+
+// Fails because Team::"temp" is not a legal RHS expression for User
+permit(principal in Team::"temp",action,resource);
+```
+
+_Entity selection_: When collecting ancestors with `getentity` [during entity selection](https://quip-amazon.com/oXOuAij6Bse8/Validating-Entity-Selection#temp:C:aLCfd096a6e6e5849aaadec7cc6f), the application only gets ancestors per the expressions in the annotation. To do that, it would evaluate these expressions for the current request, and provide the ancestors present in the result, if present. For example, suppose our TinyTodo request was
```
khieta, line 383: Probably don't want Quip links in the public RFC

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+### Dereference levels
+
+During validation, we extend entity types with a *level*, which is either `∞` or a natural number. We call these *leveled entity types*. The level expresses how many entity dereferences are permitted starting from an expression having that type. Levels on types are only used internally; policy/schema writers never see them. When the level is `∞`, validation is unchanged from the status quo.
+
+### Specifying the level
+
+The maximum level that policies may validate at is specified in the schema. The benefit of this is that policy writers can see the expected form of policies in one place. For example, we could extend our schema above as follows:
+
+```
+level = 1;
```
khieta, line 136: I like this syntax (and prefer the info here as opposed to an argument to validation). Minor implementation question: if this syntax is used in one schema fragment, does it automatically apply to all other schema fragments? Or do the other schema fragments also have to explicitly use the same level?

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+
+### Alternative: Level as a validation parameter, not in the schema
+
+This RFC has suggested that the level should be specified in the schema. Alternatively, we could specify the level as a parameter to the validator itself, leaving schemas unchanged. The benefit of doing so is that entity slicing becomes an orthogonal concern: The schema specifies the structure of data, but nothing about what data is required/allowed for a request. The drawback is that policy writers cannot look in one place to know the limits on the policies they can write.  
+
+### Alternative: Per-entity levels, rather than a global level
+
+We might refine in-schema `level` to not apply to all entities, but rather to particular entity types. For example, for `User` entities (bound to `principal`) we might specify level 2 but for any other entity type we specify `level` as 0, as per the following schema:
```
khieta, line 278: Not a fan of this alternative. I think it will make the validator behavior more difficult for users to understand.

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+   
+   while current-level <= level || worklist.empty? {
+        // `t` will become our new worklist
+        let t : Set<EUID> = {};
+        // Clear our worklist
+        for euid in worklist {
+            if !es.contains(euid) {
+                es.insert(euid, get-entity(euid));
+            }
+            ; Set of all euids mentioned in this entities data
```
khieta, line 195: Nit: inconsistent comment syntax
```suggestion
            // Set of all euids mentioned in this entities data
```

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+   }
+
+    return es;
+}
+
+```
+
+### Soundness
+
+Leveled types permit us to generalize the soundness statement we use today. Currently, soundness assumes that expressions like `principal.manager` will never error, i.e., that the proper entity information is available. Now we assume dereferences are only available up to a certain level. In particular, we define a well-formedness judgment μ⊢v:τwhere μ is an entity store, v is a value, and τ is a possibly leveled type. This judgment is standard for values having non-entity type, e.g.,
```
khieta, line 212: Nit: spacing
```suggestion
Leveled types permit us to generalize the soundness statement we use today. Currently, soundness assumes that expressions like `principal.manager` will never error, i.e., that the proper entity information is available. Now we assume dereferences are only available up to a certain level. In particular, we define a well-formedness judgment μ⊢v:τ where μ is an entity store, v is a value, and τ is a possibly leveled type. This judgment is standard for values having non-entity type, e.g.,
```

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+There are two rules for entity literals of type `E::s`. The first says level `0` is always allowed. The second says a literal can have a level `n` only if `n` dereferences are possible starting from `E::s`:
+
+```
+mu |- E::s : E(0)
+
+mu(E::s) = (v,h) where v = { f1: v1, ..., fn: vn } and h is an ancestor set
+mu |- v1 : T1 ... mu |- vn : Tn
+where level(T1) >= n-1
```
khieta, line 231: Also need `level(Tn) >= n-1` right?. Also, I think you might want to use a different index for the attributes to avoid confusion with the level `n`

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+
+We believe that ultimately level-based validation and entity manifests should coexist. Level-based validation is used to bound entity retrieval work and prevent pathological policies, while entity manifests are used to more efficiently define the needed entity data. We could imagine a deployment strategy in which we accept this RFC and perform generic entity slicing as described here, and then later implement entity manifests, and start performing slicing using those instead.
+
+### Alternative: Level as a validation parameter, not in the schema
+
+This RFC has suggested that the level should be specified in the schema. Alternatively, we could specify the level as a parameter to the validator itself, leaving schemas unchanged. The benefit of doing so is that entity slicing becomes an orthogonal concern: The schema specifies the structure of data, but nothing about what data is required/allowed for a request. The drawback is that policy writers cannot look in one place to know the limits on the policies they can write.
```
khieta, line 274: I'd argue that the schema _does_ specify what data is required/allowed for a request since it specifies each action's expected principal/resource/context. So I think it makes sense to have level information in the schema too.

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+* `level({ f1: T1, ..., fn: Tn })`  = `m` where `m` = `min(level(T1),...,level(Tn))`
+* `level(Long)` = `level(String)` = `level(Boolean)` = `level(Set(T))` = (etc.) = ∞ for all other types
+
+Thus the soundness statement says that 
+
+1. If policy set `p` is validated against a schema `s` with level `l`, where  `l < ∞` .
+2. If entity store μ is validated against schema `s` 
+3. If request σ is validated against schema `s` 
+4. If μ ⊢ σ(principal): E(l) where E is the expected type of the principal for this particular action, and  similarly check the proper leveling of action, request, and context.
+    1. We prove separately that `slice` and σ produces a store μ that satisfies this property
```
khieta, line 248: I was a little confused about how this sub-point is relevant. Maybe elaborate more, or delete it?

khieta (approved):
Left a few small comments, but I like this proposal in general. I think it makes sense as a first step before RFC 74.

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,440 @@
...
+
+### Entity dereferences
+
+An "Entity Dereference" is any operation that requires directly accessing an entity’s data. The following is an exhaustive list of entity dereferencing operations
+
+1. `e1 in e2` dereferences `e1` (but not `e2`)
+2. `e1.foo` dereferences `e1` iff `e1` is an entity
+3. `e1 has foo` dereferences `e1` iff `e1` is an entity
+
+Note that testing equality *is not* a dereferencing operation. Therefore expressions like `e1 == e2`, `e1.contains(e2)` and `e3 in e2` do not dereference `e1` or `e2`.
```
cdisselkoen, line 93:
```suggestion
Note that testing equality *is not* a dereferencing operation. Therefore expressions like `e1 == e2`, `e1.contains(e2)` and `e1 in e2` do not dereference `e1` or `e2`.
```

text/0076-entity-slice-validation.md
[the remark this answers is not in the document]
cdisselkoen: Another argument for treating entities in the context different from entity literals in policies is that they might be written by different personas, if you have policy authors separate from app developers.  I'm unsure which of those two personas is responsible for the `context` contents.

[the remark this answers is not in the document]
john-h-kastner-aws: I agree with Mike here in terms of validity of the use case. At the same time, I don't like that a level 1 policy can sometimes contain `context.foo.bar.baz`. This feels likely to confuse users.

text/0076-entity-slice-validation.md
cdisselkoen: This is a great point.

I would propose that each fragment must declare its own level, and it's an error if the levels do not agree.  Alternately, if we take Alternative C (per-entity levels), it would be natural to allow different levels in each fragment, where the level in each fragment applies to all entity types declared in that fragment.

text/0076-entity-slice-validation.md
cdisselkoen: +1

text/0076-entity-slice-validation.md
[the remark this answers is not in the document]
mwhicks1: The dot notation applying to both records and entities is ... confusing.

text/0076-entity-slice-validation.md
mwhicks1: Proposal: When combining fragments, at most one can have a `level` specification, else it's an error. 

This works when you are importing a bunch of common and entity type defs into a single app, where the app specifies the level.

cdisselkoen: This makes more sense than my initial proposal when you consider the possibility that you have library-like fragments of entity definitions (or common-type definitions, etc). There's no reason library authors should (have to or want to) specify a level, it seems like purely a consideration for the app.

text/0076-entity-slice-validation.md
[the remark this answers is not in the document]
andrewmwells-amazon: Discussed this with the team. For now, we'll keep the RFC as-is and wait for more user input on specifying finer grained levels.

Some notes from the discussion for future reference:

Some users want to pass and dereference context entities. Treating these as entity literals wouldn't work for them.

Some users want level 1 to mean "I only have to pass in the P,A and R." If their schema doesn't allow any entities in the context then this RFC works for them. 

Some users may want to use "level 1" as above and do have entities (directly) in their context. This RFC would require these entities to be included in the slice. This loses some flexibility. E.g., someone could write `context.delegate == resource.owner"` and not pass `context.delegate`. We're open to extending levels to include different level behavior for all contexts / each actions context in the future if users want this.

text/0076-entity-slice-validation.md
mwhicks1: As a fix, I pulled out the soundness statement for `slice` on its own, below.

text/0076-entity-slice-validation.md
mwhicks1: Note: `e1 in e2` _does_ dereference `e1`, which is why I wrote `e3` there.

andrewmwells-amazon (approved):

aaronjeline (approved):

cdisselkoen (approved):

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,443 @@
...
+
+Also note that by “dereference” above, we are only referring to the given operation, not about any operations that might be executed to evaluate subexpressions. For example, while `e1 in e2` only dereferences `e1` (per rule 1 above), if `e2` is `principal.foo`, then of course evaluating `e2` dereferences `principal` (per rule 2 above).
+
+### Entity roots
+
+The variables `principal`, `action`, and `resource` are *entity roots*. (Entity literals are *not* entity roots, and operations on such literals are limited, as described below.) They directly name entities from a request that could be dereferenced in a policy. A `context` is not an entity root directly; rather, any attributes it (transitively) contains that have entity type are considered roots. For example, consider the following schema:
```
andrewmwells-amazon, line 99:
```suggestion
The variables `principal`, `action`, and `resource` are *entity roots*. (Entity literals are *not* entity roots, and operations on such literals are limited, as described below. Slots and the `Unknown` entity are treated as entity literals because slots will be replaced by literals and `Unknown` cannot be dereferenced.) They directly name entities from a request that could be dereferenced in a policy. A `context` is not an entity root directly; rather, any attributes it (transitively) contains that have entity type are considered roots. For example, consider the following schema:
```

text/0076-entity-slice-validation.md
hakanson: My initial reading of this also thought it was a typo.  Is readability better if you combine DOES and DOES NOT deference into the same list, so `e1 in e2` and `e3 in e2` examples are closer together?

Also, does the reflexive `e1 in e1` dereference `e1` or is that equality?

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,443 @@
...
+   principal.joblevel > 6 && principal.location like "DEF*" ||
+   principal.location == resource.owner.location
+};
+```
+
+This policy does *not* validate at level 1. That is because the expression `resource.owner.location` is only valid at level 2: It requires dereferencing `resource` to retrieve the entity in the `owner` attribute, and then requires dereferencing that entity to obtain the contents of its `location` attribute. As a result, if we only provided the `principal` and `resource` entity data, as described above, evaluating policy 6 could produce an evaluation error since data may not be provided for `resource.owner`.
+
+All of the Cedar example policy sets validate with level-based validation, either at level 1 or 2; see the Appendix for details.
+
+## Motivation
```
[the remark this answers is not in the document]
patjakdev: Also, I'm curious if there are real-world examples which motivated this proposal. If so, I think it would be useful to include them.

text/0076-entity-slice-validation.md
```diff
@@ -0,0 +1,443 @@
...
+
+We believe that ultimately level-based validation and entity manifests should coexist. Level-based validation is used to bound entity retrieval work and prevent pathological policies, while entity manifests are used to more efficiently define the needed entity data. We could imagine a deployment strategy in which we accept this RFC and perform generic entity slicing as described here, and then later implement entity manifests, and start performing slicing using those instead.
+
+### Alternative: Level as a validation parameter, not in the schema
+
+This RFC has suggested that the level should be specified in the schema. Alternatively, we could specify the level as a parameter to the validator itself, leaving schemas unchanged. The benefit of doing so is that entity slicing becomes an orthogonal concern: The schema specifies the expected type structure of data provided with a request, but not how much of that data is required. The drawback is that policy writers cannot look in one place to know the limits on the policies they can write.
```
patjakdev, line 277: > The drawback is that policy writers cannot look in one place to know the limits on the policies they can write.

I don't think this is so awful. The error message from the validator will tell the author what the limit is, if they hit it. This also allows for more flexibility in granularity of level validation in the future without having to change the Cedar language.

patjakdev (commented):
I generally find [RFC 74](https://github.com/cedar-policy/rfcs/pull/74) to be a more appealing approach to slicing. I think that were RFC 74 to be accepted, this one wouldn't have much utility given the cost to implement and (IMO) unlikeliness for it to apply well generally.

If you've got some real-world use cases where validation-time depth limitations would have solved performance problems for users in a way that RFC 74 wouldn't or couldn't, I'd definitely be open to changing my mind.

The one that springs to mind is maliciously authored policies, but I wonder if that wouldn't be something better guarded against in the Cedar library implementations rather than as part of the schema.

text/0076-entity-slice-validation.md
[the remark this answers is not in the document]
andrewmwells-amazon: I agree that #74 is likely to be more suitable for most consumers of Cedar. This RFC gives a less granular tool for describing the entity data that needs to be included in the slice. Sometimes, being more granular is better.

The usecase I find most compelling for this RFC is one where one party writes the entity slicing algorithm and another writes the policies (e.g., you offer a service where end-users can author Cedar policies). With #74, the entity slicing potentially needs to be changed whenever policies change (because we were passing in only the entities needed). With this RFC, we pass in some entity data that may be unused, but this means policies can change freely as long as they continue to only use the entity data at level `n`.

text/0076-entity-slice-validation.md
andrewmwells-amazon: After implementing this (https://github.com/cedar-policy/cedar/pull/1146) I am inclined towards making this a parameter.

Arguments for:
- In the non-JSON schema format, `level` can occur at the top of any namespace. This makes finding `level` annoying. We error if `level` occurs more than once, but we could just take it as a single parameter to `validate`.
- The JSON schema format has the same issue, but JSON isn't ordered, so it feels more native.
- Schema format remains unchanged and we can add `level` to the schema later if desired. In particular, adding the per-entity `level` as described in alternatives would require new syntax (something like `@level(1)`, as proposed in the alternative). Delaying this decision will let us hear if more people would prefer that and a `global level` or something similar.

Arguments against:
- Keep data in one place.
- Easier to change schema than to change code (but the entity slicing code would still need to change).

text/0076-entity-slice-validation.md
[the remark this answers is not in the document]
andrewmwells-amazon: "End-users writing policies" is (in vague terms) the real-world example that motivated this.

shaobo-he-aws (approved):

PR #80, opened 2024-08-13
Datetime Extension

apg:
Proposal for adding a datetime extension, co-authored with @khieta.

[Rendered](https://github.com/strongdm/cedar-rfcs/blob/datetime-rfc/text/datetime-extension.md)


emina:
Thank you for the carefully thought out RFC!  

I agree that having some support for dates/times would be useful, but unfortunately, this proposal is out of bounds of what's analyzable.

Specifically, we cannot support the proposed time representation and operators for two reasons.  

First, if we keep everything else the same, then it's not possible to enforce the well-formedness constraints on time components (year, month, day, hours, minutes, seconds) without using quantifiers, which would put the encoding into the undecidable fragment of SMT.  This is because Cedar lets you put arbitrary types into sets, and to ensure that every time value in a set is well-formed, we would have to use quantifiers. 

Second, even if we were able to ground the quantifiers by imposing restrictions on how times can be used (e.g., time values can't be included in sets, either directly or transitively as part of another value), enforcing and solving these ground time constraints would still be expensive.  For each occurrence of a datetime value, we would need to emit an assumption that says if the year field corresponds to a leap year, and if the month is February, then the month has 29 days, otherwise it has 28 days.  We would also need to emit an if-then-else condition that covers all the other months too.  Note that determining if a given year is a leap year itself requires checking multiple conditions.  Specifically, it is not enough to check if the year is divisible by four (see [here](https://learn.microsoft.com/en-us/office/troubleshoot/excel/determine-a-leap-year)).  

To retain analyzability, we can go the IAM route (see [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_Date)) and support date comparisons, perhaps with some duration operations as well.  The IAM operators are analyzable because we can internally represent time values as Unix epochs, which are unconstrained 64-bit integers, and the IAM operations on this representation reduce to long comparisons.  We are free to make the string representation of times richer than what IAM supports as long as the internal representation is longs, and the operators on times are expressible as simple operations on longs.

apg:
Thanks, @emina, for taking a look! A few ideas / questions in what follows... 

> Specifically, we cannot support the proposed time representation and operators for two reasons.
>
> First, if we keep everything else the same, then it's not possible to enforce the well-formedness constraints on time components (year, month, day, hours, minutes, seconds) without using quantifiers, which would put the encoding into the undecidable fragment of SMT. This is because Cedar lets you put arbitrary types into sets, and to ensure that every time value in a set is well-formed, we would have to use quantifiers.

I'm not tied to the proposed representation. I am hopeful we can define an abstract data type with useful, and practical time operations. If using a long instead of a "record like" is the way to get there, than so be it.

Please forgive the ridiculous questions and thoughts I'm replying with, as I'm more or less ignorant in Automated Theorem Proving, SAT, and SMT.

As specified in this document, the `datetime` constructor is the only way to set the fields of a datetime object. The `offset()` method does not make an in-place modification, but rather, instantiates a new datetime with an updated set of component values. If the constructor only produces well formed `datetime` objects, or "errors out" at runtime, then I don't see how this is any different than a long? 

> Second, even if we were able to ground the quantifiers by imposing restrictions on how times can be used (e.g., time values can't be included in sets, either directly or transitively as part of another value), enforcing and solving these ground time constraints would still be expensive. For each occurrence of a datetime value, we would need to emit an assumption that says if the year field corresponds to a leap year, and if the month is February, then the month has 29 days, otherwise it has 28 days. We would also need to emit an if-then-else condition that covers all the other months too. Note that determining if a given year is a leap year itself requires checking multiple conditions. Specifically, it is not enough to check if the year is divisible by four (see here).

Can we prove these operations under a simpler calendar model, and then let implementations choose the most appropriate calendar model (*cough* Gregorian) for their implementation? Most general purpose programming languages have robust date/time facilities, so it stands to reason that different implementations of Cedar will adopt them, rather than start from scratch.

I believe that the operations requested in the this API spec are valid under _any_ calendar that has years, months, days, hours, minutes, seconds, and whose measurements (instants in time) form a partial order.

> To retain analyzability, we can go the IAM route (see here) and support date comparisons, perhaps with some duration operations as well. The IAM operators are analyzable because we can internally represent time values as Unix epochs, which are unconstrained 64-bit integers, and the IAM operations on this representation reduce to long comparisons. We are free to make the string representation of times richer than what IAM supports as long as the internal representation is longs, and the operators on times are expressible as simple operations on longs.

The current IAM operations are insufficient to express many of the examples in this RFC. Adding additional functions and a duration type could improve the situation, but I think we'll still fall short. There wouldn't be a facility for policies that have periodic decision making abilities (e.g. "everyday between $x and $y"), which we've identified as important for our customers.

emina:
I'll try to clarify some of the concerns around the proposed API and explain better (hopefully :) why the current API is not analyzable.  To a first approximation, the best we can do is IAM plus a few duration operations (TBD).  Let's dig into why this is the case.

> Please forgive the ridiculous questions and thoughts I'm replying with, as I'm more or less ignorant in Automated Theorem Proving, SAT, and SMT.
> 
> As specified in this document, the `datetime` constructor is the only way to set the fields of a datetime object. The `offset()` method does not make an in-place modification, but rather, instantiates a new datetime with an updated set of component values. If the constructor only produces well formed `datetime` objects, or "errors out" at runtime, then I don't see how this is any different than a long?

These are great questions :). And your point is exactly right for the runtime implementation of the `datetime` constructor that we'd have in Lean, Rust, or Go.  The issue is the analysis-time encoding of various operators.  

To understand the issues here, it helps to have a mental model of how the analysis works.  The analyzer is a compiler:  it takes as input a set of Cedar policies, a property we want to check about them (e.g., are they equivalent?), and compiles these policies into the [input language of the SMT solver](https://smt-lib.org/language.shtml).  The solver then checks if the property holds and, if not, it returns a concrete input (a request and entities) that violate the property.  It helps to think of the solver as testing the encoded policies on the (infinite) set of all possible inputs.

Now, the language that the solvers support is very limited compared to a general purpose language.  And the practical subset of this language is even more limited than what the SMT standard suggests, because some operators in the SMT language are either undecidable (the solver can run forever trying to test the policies) or extremely inefficient to reason about (the solver will come back but not in any reasonable time).

So, what's expressible in Cedar is limited to what we can efficiently compile to this core functional language:  think, no loops or recursion, machine integers (aka bitvectors), some string operators, some set operators, some map operators (uninterpreted functions), structs, conditionals, and boolean operators.

If you're curious, [here](https://arxiv.org/abs/2403.04651) are some more details about the compilation process (Section 4).

Given this limited compilation target, we can't actually translate the `datetime` constructor itself, because this would require writing a parser for `datetime` strings in the SMT language.  This parser is hugely complicated, because it also needs to validate its input, and establish that the input string is a correct date time, enforcing the constraints on leap years, numbers of days in months etc.  Even if we could fully unroll all loops in this parser to express it in SMT, the resulting SMT expression would be analyzable in name only :).  In practice, there is a limit to how complicated and big of a formula a solver can solve in a reasonable amount of time.  

This is why we don't actually encode the parsing of Decimals or IPAddresses either.  Note that the strict validator creates a loophole for us to avoid parsing in the solver:   it rejects attempts to call the decimal constructor on a non-literal, e.g., `decimal(context.someStringRepresentation)`, and instead asks that arguments to these constructor are string literals.  We use the regular Rust/Lean parser to parse these strings, and then turn them into concrete values (bitvectors) representable in SMT.  Input (e.g., context) values of these types are represented as bitvector variables (unknowns), and the solver tries to find values for them that violate the desired property.  

Ok, so if we can't translate the datetime parser to SMT, how would we enforce the desired constraints on date values?  The only way to do is by emitting extra formulas that say the date components are within the right ranges.  Think of it as emitting a giant conditional expression that says "if all dates occurrences are well-formed, then the property we care about holds." 
 Now, we can't emit this expression for a general Cedar policy because of sets:  checking that every element of a set is well-formed conceptually requires iteration, which isn't supported by SMT.  And even if prohibit putting dates in sets, the resulting well-formedness constraints would still be too complex in practice. This is why we can't support the proposed representation while remaining in the decidable / practical fragment of the SMT language. 

> Can we prove these operations under a simpler calendar model, and then let implementations choose the most appropriate calendar model (_cough_ Gregorian) for their implementation? Most general purpose programming languages have robust date/time facilities, so it stands to reason that different implementations of Cedar will adopt them, rather than start from scratch.

Unfortunately, not for analysis.  We can definitely do this in general (in Lean), but for the analysis to work, we must be able to express each operation in the very limited SMT language.  

> The current IAM operations are insufficient to express many of the examples in this RFC. Adding additional functions and a duration type could improve the situation, but I think we'll still fall short. There wouldn't be a facility for policies that have periodic decision making abilities (e.g. "everyday between $x and $y"), which we've identified as important for our customers.

I totally understand that a richer set of operators would be desirable.  But if a datetime operation cannot be expressed (roughly) as a loop-free expression over integers, it's out of bounds of what's analyzable.  The IAM operators have the nice property of being expressible as < and <= over longs.  We can push this a bit further than IAM, but not to the point where we can support very complex calculations, like extracting the year or month from an epoch. 

I'd be happy to chat more offline (to save us some typing :) about whether there is a middle ground that we can design.  I'm hoping we can find a reasonable design that combines IAM++ and some of the alternatives discussed in the RFC.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,490 @@
...
+
+Here are the full validity constraints on datetime values:
+
+- `year` must be between 1900 and 2100
+- `month` must be between 1 and 12
+- `day` must be between 1 and...
+  - 30 for `month`=4,6,9,11
+  - 31 for `month`=1,3,5,7,8,10,12
+  - 28 for `month`=2 if `year` is not divisible by 4
+  - 29 for `month`=2 if `year` is divisible by 4
```
andrewmwells-amazon, line 193: I think this should at least be `y % 400 == 0 || (y % 4 == 0 && (y % 100 != 0))`

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,490 @@
...
+  resource
+) when {
+  context.afterBrexit &&
+  context.location.countryOfOrigin == 'GB' &&
+  resource.owner == 'EU'
+}
+```
+
+
+### Alt. C: Combine Unix Time with additional Pre-Computed context.
```
andrewmwells-amazon, line 378: This is my preference. In addition to the concerns Emina raised, I'd like to point out that different languages` time libraries aren't consistent. E.g., Java uses the Java Time-Scale (https://stackoverflow.com/questions/30984599/how-does-the-oracle-java-jvm-know-a-leap-second-is-occurring). UTC requires empirical observations of leap seconds (https://blogs.mathworks.com/loren/2021/05/06/what-is-a-leap-second-anyway/). IMO, Cedar shouldn't have an opinion about which of these you should use.

One can argue that a few seconds of authorization difference don't matter, and I agree this is usually true but I can think of cases like migrating existing authz code written in LangOfChoice to Cedar where these discrepancies will come up in shadow mode. Also I'm uncomfortable with an unbounded divergence between Cedar time and UTC / Java Time / whatever.

I think the right solution is to pass the date information via the context so that it matches the full-fledged date+time facilities of the language you're using.

text/0080-datetime-extension.md
apg: Correct, but the RFC puts a (I think? practical) constraint on the year to be between 1900 and 2100, in an attempt to simplify the quirks of the Gregorian calendar (e.g. leap years).

andrewmwells-amazon: I missed that. (But I also don't like it 😄)

text/0080-datetime-extension.md
andrewmwells-amazon: Sort of. What I'd like is for people to pass things like `context.day_of_week` (or better, `context.is_weekday`) and `context.month` in the context and add "native" (or maybe I should call it "extended"?) support just for the comparison operators.

text/0080-datetime-extension.md
apg: Ha! I didn't like it either, but, it simplified things! :)

text/0080-datetime-extension.md
emina: Yes, the first 10 operators would be analyzable because they are expressible as simple arithmetic operations over longs.  And since every long is a valid timestamp, we don't need to try and enforce any well-formedness constraints on them (which was the issue with the YYYYMMDDHHSS representation). 

I'll have to think more about the `DT.toDate()` operator.  If there is a definition of this operator that is expressed as arithmetic over longs, we can analyze it.   Defining this operator depends on what "midnight" means.  The usual definition is timezone-specific, so we can't support that one.  But if there is some alternative definition based on epoch arithmetic, it should be expressible in SMT.

andrewmwells-amazon: I think the `toDate` op can just be: `DT / (24*60*60)` which will give you the number of days since Jan 1 1970. When pretty-printing this from Rust, we could calculate the calendar day.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,490 @@
...
+```
+
+
+## Detailed design
+
+This RFC proposes supporting two new extension types: `datetime`, which represents a particular instant of time, up to second accuracy, and `duration` which represents a duration of time.
+To construct and manipulate these types we will provide the functions listed below.
+All of this functionality will be hidden behind a `datetime` feature flag (analogously to the current decimal and IP extensions), allowing users to opt-out if they do not want this functionality.
+
+- `datetime(s)` constructs a datetime object. Like with our other constructors, strict validation requires `s` to be a string literal, although evaluation/authorization support any string-typed expression. The string must have the form `"YYYY-MM-DDThh:mm:ssZ"` (UTC), or `"YYYY-MM-DDThh:mm:ss(+/-)hhmm"` (with Time Zone Offset).
```
andrewmwells-amazon, line 142: Thinking about this more, I'd vote for millisecond resolution. I think throttling is a fairly plausible use case.

text/0080-datetime-extension.md
apg: My thinking was actually `(DT - (DT / 24*60*60))` -- which gives you the "start of the day." With this you can then do something like:

`resource.createdAt > currentTime.toDate() && resource.createdAt < currentTime.toDate().add(duration(1, 'day'))`

Similar with the `toTime()` which would give you the amount of time since the day started... assuming UTC.

text/0080-datetime-extension.md
apg: I'm fine with millisecond precision!

apg:
Based on all the feedback so far (Thank you!!!!!), it seems like I should reposition this RFC to be Alt C, with the API discussed in the comments, and millisecond precision. 

I'll try to post a rewrite by the end of the week.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+- `datetime(string)` constructs a datetime value. Like with other extension function constructors, strict validation requires `string` to be a string literal, although evaluation/authorization support any string-typed expression. The string must be of one of the forms, and regardless of the timezone offset is always normalized to UTC:
+  - `"YYYY-MM-DDThh:mm:ssZ"` (UTC)
+  - `"YYYY-MM-DDThh:mm:ss.SSSZ"` (UTC with millisecond precision)
+  - `"YYYY-MM-DDThh:mm:ss(+/-)hhmm"` (With timezone offset in hours and minutes)
+  - `"YYYY-MM-DDThh:mm:ss.SSS(+/-)hhmm"` (With timezone offset in hours and minutes and millisecond precision)
+- `.add(duration)` returns a new `datetime`, offset by duration.
+- `.toDate()` returns a new `datetime`, truncating to the day, such that printing the `datetime` would have `00:00:00` as the time.
+- `.toTime()` returns a new `duration`, removing the days, such that only milliseconds since `.toDate()` are left. This is equivalent to `DT - DT.toDate()`
+
+Values of type `datetime` can be used with comparison operators:
```
emina, line 147: I'm not sure that we want to support operator overloading here.  

Note that we don't support operator overloading for other extension types (such as decimals).

We can keep all of these operators, but for consistency of the language design, I suggest that we make them into methods:

* `DT1.lessThan(DT2)` 
* ...

And same for datetime arithmetic:

* `DT1.sub(DT2)` or whatever we want to call this method.

Question:  can durations be negative?

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+
+- `duration(long, string)` constructs a duration value. The string argument must be one of `"days", "hours", "minutes", "seconds", "milliseconds"`. Strict validation requires `string` to be a string literal, although evaluation/authorization support any string-typed expression.
+- `.toMilliseconds()` returns a `long` describing the number of milliseconds in this duration. (the value as a long, itself)
+- `.toSeconds()` returns a `long` describing the number of seconds in this duration. (`.toMilliseconds() / 1000`)
+- `.toMinutes()` returns a `long` describing the number of minutes in this duration. (`.toSeconds() / 60`)
+- `.toHours()` returns a `long` describing the number of hours in this duration. (`.toMinutes() / 60`)
+- `.toDays()` returns a `long` describing the number of days in this duration. (`.toHours() / 24`)
+
+Values with type `duration` duration can also be used with comparison operators:
```
emina, line 170: Same comment as above:  for consistency, these operations on extension types should be methods (just like for Decimal).

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Unresolved questions
+
+### Should we consider local time at all?
+
+Writing programs that involve dates and time is tricky. Standardizing on UTC representations for all dates and times trades achieving evaluator correctness more quickly at the expense of decreased readability for policy authors and auditors.
+
+### Should we provide a `currentTime` function?
+
+The examples above expect `currentTime` (or `now`) to be passed in the context during the authorization request. An alternative would be to provide a `currentTime()` extension function that computes the current time at policy evaluation time. This has a drawback, when testing policies, in that the evaluator must be somehow extended to mock the time.
```
emina, line 367: Worth noting here that we cannot provide `currentTime` at all.  That's because `currentTime` is stateful; it's not a pure function.  That is, `currentTime() == currentTime()` could return false, and in fact `currentTime()` is likely to return different values during the same authorization call over multiple policies that use it.

text/0080-datetime-extension.md
apg: How hard would it be to support overloading? The ergonomic, and readability improvements are substantial... and they're just longs, really. :)

There's no reason durations can't be negative, I think?

text/0080-datetime-extension.md
apg: Yes, exactly! currentTime() would also have to consider leap seconds, which we should definitely punt on. :)

I can just drop this section if it's easier.

apg: (I'll add your comment about purity, and maybe move to out of scope, instead)

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+) when {
+  context.now.timestamp > datetime("20200131T23:00:00Z") &&
+  context.location.countryOfOrigin == 'GB' &&
+  resource.owner == 'EU'
+}
+```
+
+## Detailed design
+
+This RFC proposes supporting two new extension types: `datetime`, which represents a particular instant of time, up to millisecond accuracy, and `duration` which represents a duration of time.
```
mwhicks1, line 134: Why millisecond accuracy? Why not more or less? I wonder about precedents, use-cases, etc. that justify this.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+- `.toHours()` returns a `long` describing the number of hours in this duration. (`.toMinutes() / 60`)
+- `.toDays()` returns a `long` describing the number of days in this duration. (`.toHours() / 24`)
+
+Values with type `duration` duration can also be used with comparison operators:
+
+- `DT1 < DT2` returns `true` when `DT1` is before `DT2`
+- `DT1 <= DT2` returns `true` when `DT1` is before or equal to `DT2`
+- `DT1 > DT2` returns `true` when `DT1` is after `DT2`
+- `DT1 >= DT2` returns `true` when `DT1` is after or equal to `DT2`
+- `DT1 == DT2` returns `true` when `DT1` is equal to `DT2`
```
mwhicks1, line 176: This is a cut&paste error from above. Change `DT` to be `dur` and change explanations appropriately, e.g.,

- `dur1 < dur2` evaluates to `true` when `dur1` is shorter than `dur2`

mwhicks1: Also: Why have `==` here (and above, for DateTimes) but not `!=`? I'd suggest having neither or both. Presumably `==` (and `!=`) continue to mean syntactic (non)equality.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Out of scope
+
+- **Conversion between UTC and epochs:** This will be particularly difficult to model and verify in Lean (although it's technically possible, see [this paper](https://dl.acm.org/doi/abs/10.1145/3636501.3636958) which does something similar in the Coq proof assistant). Since it will likely require input-dependent loops, it is unlikely that this can be reasoned about efficiently with SMT.
+
+- **Conversion between UTC and other Time Zones:** Time Zones are a notoriously complex system that evolves rapidly. We avoid this complexity by offering `datetime.add(duration).` Policy authors that require "local time" can either provide an additional datetime in `context` or provide a `duration` to the `context` and call `.add()` to shift the time.
+
+- **Leap Seconds:** Cedar does not have a clock, and this proposal does not add one. Instead, we pass Unix Time through `context` or an entity and let the system / application handle leap seconds.
+
```
mwhicks1, line 187: Reword "Instead, we pass Unix Time through `context` or an entity" to "Instead, the application passes the current time through `context` or an entity attribute". 

If you want to say that a DateTime is equivalent in its representation to Unix Time you should say that somewhere else. Indeed, I think you should give the underlying representation here because it needs to be known for the entity JSON format.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+### Support for date/time in other authorization systems
+
+AWS IAM supports [date condition operators](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_Date), which can check relationships between date/time values in the ISO 8601 date format or Unix time. You can find an example of a IAM policy using date/time [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws-dates.html).
+
+[Open Policy Agent](https://www.openpolicyagent.org) provides a [Time API](https://www.openpolicyagent.org/docs/latest/policy-reference/#time) with nanosecond precision and extensive time zone support. During policy evaluation, the current timestamp can be returned, and date/time arithmetic can be performed. The `diff` (equivalent to our proposed `since` function) returns an array of positional time unit components, instead of a value typed similarly to our proposed `duration`.
```
mwhicks1, line 194: I don't see a proposed `since` function.

philhassey:
I agree that `now()` or `currentTime()` should be removed.  These are better supplied by the caller as part of the context.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Alternatives
+
```
mwhicks1, line 197: I wouldn't call these alternatives: they are workarounds. That's because they don't require changes to Cedar, they just present ways to use it as it exists now.

text/0080-datetime-extension.md
mwhicks1: In general, I'd like to see more technical details.

- How will you implement policy evaluation for these operators? Depends on their representation.
- How will we implement validation? Seems we need to change the way existing operators are treated. What's the effect there?
- What is the strategy for SMT/analysis?

text/0080-datetime-extension.md
mwhicks1: Note: If we do overloading here, we should really do it for `Decimal()` numbers too.

khieta: I don't think it would be too much effort to support (at least in the Rust/Lean implementation), we just don't have a precedent for it. If we decide to use `<` et al for date/time, then we should go back and do the same for decimal.

Also: Note that the decimal extension currently uses `.lessThan()`, so we'd either need to choose a new name for date/time or support overloading for that operation.

emina: I think we would have negative durations if we have something like "2024.08.20" - "2024.08.30". 

Or do we intend the operator to return the absolute value here?

If so, that's another argument for not overloading `-` for this operator.

text/0080-datetime-extension.md
andrewmwells-amazon: https://github.com/cedar-policy/rfcs/pull/80#discussion_r1718463176

text/0080-datetime-extension.md
andrewmwells-amazon: (Dropping milliseconds for readability)
`DT / (24*60*60)` should give you the number of days elapsed since epoch start (which will be at midnight). I think that's what `toDate` (or `toDays` or some other name) should do.

To get the "start of day" I think you want `epoch_start + (24*60*60)*(DT / (24*60*60))`

I think you maybe meant `(DT - (24*60*60)*(DT / 24*60*60))` which would give you the duration since "start of day." I'd be open to making that (or preferably "start_of_day") a function as well.

text/0080-datetime-extension.md
mwhicks1: Why not nanoseconds? That's what OPA/Rego supports. Also: What about other Date/Time libraries, e.g., available from Python.

text/0080-datetime-extension.md
khieta: Typo - replaced with the `-` operation in future iterations

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,367 @@
...
+```cedar
+permit(
+  principal is User,
+  action == Action::"view",
+  resource in Folder::"device_prototypes"
+)
+when {
+  principal.department == "HardwareEngineering" &&
+  principal.jobLevel >= 10 &&
+  (context.now.timestamp - principal.hireDate) > duration(365, "days")
```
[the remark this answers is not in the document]
aaronjeline: It seems this RFC is proposing operator overloading, which we've decidedly avoided. Do we think it's worth revising that decision?

mwhicks1:
I _think_ I like this proposal, but I am looking for more details. Spelling them out may reveal gotchas that are relevant to approval.

For example, I think what's going on is that a `datetime` and `duration` are represented as long integers under the covers, like Unix time. This means that evaluation of various operations is pretty simple, and encoding for SMT-based policy analysis is simple, too. But some details are important, e.g., if `duration` cannot be negative then x-y is really |x-y|. So I'd like to see those spelled out.

I'd also like to see the basic approach to validation spelled out. If you look at the recent proposals for [function macros](https://github.com/cedar-policy/rfcs/pull/61) or [embedded attribute maps](https://github.com/cedar-policy/rfcs/blob/main/text/0068-entity-tags.md), etc. you'll see these features written out carefully.

text/0080-datetime-extension.md
cdisselkoen: I think we should have negative durations and not implicitly the absolute value.

Cedar already supports some kinds of overloading, e.g., `in Group::"A"` and `in [Group::"A"]` are both valid.  And `has` works with both entities and records on its LHS.

cdisselkoen: I guess in summary, I think overloading here (and for `Decimal`, retroactively) is OK.

text/0080-datetime-extension.md
cdisselkoen: All Cedar values currently support `==` and `!=`.  I think it would be very confusing if we didn't support both of those for both duration and datetime.

text/0080-datetime-extension.md
apg: Whoops! Yeah! What I actually meant was `(DT - (DT % (24*60*60))` which drops the left over seconds. That one character typo made all the difference.. 😮‍💨 🌴.

text/0080-datetime-extension.md
apg: It feels to me that the order of the operands in `DT1 - DT2` may hold a lot of information about the intent of the policy. 

`(startDate - now) > 7days` vs `(now - startDate) > 7days` expresses two completely separate concepts. One asks "are you more than 7 days from starting?" The other "did you start at least 7 days ago?"

If you take absolute durations:

`abs(startDate - now) > 7days` and `abs(now - startDate) > 7days` say the same exact thing. You'd need an additional `startDate > now` or `now > startDate` to get the original intent.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,365 @@
...
+}
+```
+
+## Detailed design
+
+This RFC proposes supporting two new extension types: `datetime`, which represents a particular instant of time, up to millisecond accuracy, and `duration` which represents a duration of time.
+To construct and manipulate these types we will provide the functions listed below.
+All of this functionality will be hidden behind a `datetime` feature flag (analogous to the current decimal and IP extensions), allowing users to opt-out.
+
+- `datetime(string)` constructs a datetime value. Like with other extension function constructors, strict validation requires `string` to be a string literal, although evaluation/authorization support any string-typed expression. The string must be of one of the forms, and regardless of the timezone offset is always normalized to UTC:
```
andrewmwells-amazon, line 138: Maybe we should also include `datetime("epoch_start")` and `datetime("epoch_end")`?
A potential usecase would be half-open intervals. E.g., the same shape of policy could be used to express buying and renting. For buying it could be:
```
permit(principal = User::"Bob", action = Action::"view", resource = Movie::"UnReleasedFilm") when {
    context.curr_time >= resource.release_date && context.curr_time <= datetime("epoch_end")
};
```
While for renting you'd use the end of the rental period, e.g., `context.curr_time <= datetime("2024-08-21T20:07:23")`.

text/0080-datetime-extension.md
andrewmwells-amazon: I'd be happy with nanoseconds if you can think of a use case for them.

text/0080-datetime-extension.md
apg: This seems fine to me. Perhaps this is mostly useful when policies are templated, possibly in the context of #61, but why would a policy author include `context.curr_time <= datetime("epoch_end")` otherwise?

text/0080-datetime-extension.md
khieta: To clarify: `==` and `!=` will work for datetime & duration the same way they do for any other value. They are not being overloaded here -- I'll update the text to reflect this.

text/0080-datetime-extension.md
apg: *I* don't have a use case for nanoseconds (and probably not millis, either, for that matter), but if you do please present it so we can discuss. 

Similarly, if someone has a use case for Coordinated Mars Time, instead of Unix Time (NASA... any interest in Cedar?), let's discuss. :)

text/0080-datetime-extension.md
apg: That's an omission on my part. I can add that and fix the typo discovered. Thanks!

text/0080-datetime-extension.md
andrewmwells-amazon: Yeah, I'm envisioning templated / programmatically generated policies where it's easier to just give every policy a start and end time. In my example the RHS condition would always true so it would hopefully be dropped if the policies were written by hand.

text/0080-datetime-extension.md
andrewmwells-amazon: For milliseconds I can imagine something like "allow at most one access per 100ms" to prevent scraping. So `context.current_time > context.last_request_time + duration(100, "ms")`. I can't think of a use for nanoseconds.

text/0080-datetime-extension.md
apg: I have a pending commit that fixes this.

text/0080-datetime-extension.md
apg: Correct. But they are alternatives because we can not implement the proposal in this RFC and advocate for one of these alternatives instead... accepting the tradeoffs of doing so.

apg:
> I'd also like to see the basic approach to validation spelled out. If you look at the recent proposals for [function macros](https://github.com/cedar-policy/rfcs/pull/61) or [embedded attribute maps](https://github.com/cedar-policy/rfcs/blob/main/text/0068-entity-tags.md), etc. you'll see these features written out carefully.

@khieta can you help with the validation approach?

apg:
> For example, I think what's going on is that a `datetime` and `duration` are represented as long integers under the covers, like Unix time. 

Yes, this is detailed in the [Detailed Design](https://github.com/strongdm/cedar-rfcs/blob/datetime-rfc/text/0080-datetime-extension.md#detailed-design). 


> This means that evaluation of various operations is pretty simple, and encoding for SMT-based policy analysis is simple, too. But some details are important, e.g., if `duration` cannot be negative then x-y is really |x-y|. So I'd like to see those spelled out.

I can explicitly state that durations should be negatable, because I believe they should be. It does seems like that stance is not yet shared amongst everyone.

apg:
OK. I think most feedback that I'm aware of has been addressed by either @khieta or myself, and it should be "safe" to reevaluate.

I've also [implemented this API in Python](https://gist.github.com/apg/bc793d42f38d73a7d84ffa0fc309419f) to validate that the behavior makes sense. It caused me to realize that we'd have a bug with `-`, so that's helpful!

text/0080-datetime-extension.md
khieta: Update: we've added sections on operator overloading and negative durations to the "unresolved questions" section. Please move conversation to comments on those sections!

text/0080-datetime-extension.md
khieta: (Resolving this conversation due to overlap with other threads. Hopefully this will keep the review easier to manage.)

text/0080-datetime-extension.md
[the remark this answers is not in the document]
khieta: (Resolving this conversation due to overlap with other threads. Hopefully this will keep the review easier to manage.)

text/0080-datetime-extension.md
khieta: Update: we've added a section on operator accuracy to the "unresolved questions". Please move conversation to a comment on that section!

text/0080-datetime-extension.md
khieta: We've added more details and better signposting in the latest revision, which will hopefully answer these questions 🙂 Would appreciate a re-review when you get a chance!

text/0080-datetime-extension.md
khieta: I don't think I understand the utility here: if you want to time-box `context.curr_time`, wouldn't it make more sense to give specific values for the bounds rather than using `epoch_start` and/or `epoch_end`? 

I'm not sure that `epoch_start` and `epoch_end` are useful abstractions here: one is in 1970, and the other is the largest representable 64-bit value, the date of which will vary with our supported precision level.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,484 @@
...
+
+Values of type `datetime` have the following methods:
+
+- `.offset(duration)` returns a new `datetime`, offset by duration.
+- `.toDate()` returns a new `datetime`, truncating to the day, such that printing the `datetime` would have `00:00:00` as the time.
+- `.toTime()` returns a new `duration`, removing the days, such that only milliseconds since `.toDate()` are left. This is equivalent to `DT - DT.toDate()`
+
+Values of type `datetime` can also be used with the `-` operator.
+
+- `DT1 - DT2` returns the difference between `DT1` and `DT2` as a `duration`. (The inverse of `-` is `DT1.offset(duration)`)
```
mwhicks1, line 150: Proposal: Let's not allow arithmetic operations on `DT` objects but define a function instead, e.g.
```
DT1.durationUntil(DT2)
```
to return a duration that is equivalent to DT2 - DT1. The reasons I like this approach:
- Normally, the minus operator takes two objects of the same type and returns an object of that type, e.g., two `Long`s and returns a `Long`, or two `Decimal`s and returns a `Decimal`. It's strange to break that structure when overloading minus here, at least to me. It's equally strange to support `-` but not also `+`, `*` etc. That's a signal that it's not the right operator. Creating a fresh function addresses all of these issues.
- It's perfectly clear, with the proposed approach, that what's returned is a `duration`.
- This approach still allows negative `duration` objects, i.e., when DT2 is earlier than DT1.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,484 @@
...
+
+`DT1` may be negative (_i.e._ milliseconds _before_ `1970-01-01T00:00:00Z`) requiring that `-` be computed specially:
+
+```
+if DT1 < 0 && DT2 >= 0 then
+    result = -(DT1 - DT2)
+else
+    result = DT1 - DT2
+```
+
```
mwhicks1, line 162: `DT1` cannot be negative. There is no "negative Jan 1, 1971." Rather the _representation_ of a DT as a number can be negative. This is an implementation detail that should not be here. Rather, it should go in the "Representation" subsection below.

text/0080-datetime-extension.md
mwhicks1: Fair enough. In the past we've specifically highlighted a way you could do without the proposal as a "workaround", just to help the reader a little. That's different than an Alternative implementation/change. See for example this [workaround](https://github.com/cedar-policy/rfcs/blob/main/text/0068-entity-tags.md#workaround-records-with-optional-attributes) for entity attribute maps. Maybe just add the "workaround" prefix to these parts?

philhassey:
How does everyone feel about this signature for duration?

`duration(string)` where string is something like

`"5d 3h 2m 1s"`

This gives us the ability to define more precise durations in policy without having to add a bunch of them together.  It also keeps a single arg (which is convenient.). 

It's worth noting that with validation turned on, arguments to extension constructors are meant to be literals (see https://docs.cedarpolicy.com/policies/validation.html#:~:text=Errors%20due%20to%20incorrect,validate%20(regardless%20of%20mode). )

> However, by default the validator runs in a strict mode that forbids passing non-literal strings to extension function constructors; in this mode, the expression above will fail to validate with the error extension constructors may not be called with non-literal expressions. An expression like ip("XYZ") in a policy will fail to validate (regardless of mode).

This suggests that there's no room for doing things like this in validated Cedar:
`duration(context.expireDays,"days")`

And really the better way would be to have `context.expireDays` just be the exact duration already.  Or if you are writing it as policy text, then `duration("5d")` would be more terse anyways.

mwhicks1 (commented):
I'm still not sold on the use of `-` for DTs, and I suggested an alternative. I also suggested some small wording/structure adjustments. Overall I think this is looking really good. Well done @khieta and @apg !

cdisselkoen:
To @philhassey's toplevel comment (sorry, I can't thread, because it wasn't made as a comment on any particular part of the RFC text)

> How does everyone feel about this signature for duration?

I like this direction for all of the reasons given, and wonder if there is some standard for specifying durations as a string just like there are ISO standards for specifying datetimes as a string?

text/0080-datetime-extension.md
emina: Strongly agree with Mike about the `-` operator.  It’s not consistent with the usual meaning or type signature of `-`.  And I also don’t like the asymmetry of overloading just one arithmetic operator.

Let’s make this a method.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,484 @@
...
+#### Representation
+
+The `datetime` type is internally represented as a `long` and contains a Unix Time in milliseconds. This is the number of non-leap seconds that have passed since `1970-01-01T00:00:00Z` in milliseconds. Unix Time days are always 86,400 seconds and handle leap seconds by absorbing them at the start of the day. Due to using Unix Time, and not providing a "current time" function, Cedar avoids the complexities of leap second handling, pushing them to the system and application.
+
+Negative Unix Time values represent the number of milliseconds before `1970-01-01T00:00:00Z`. 
+
+### Durations of Time (`duration`)
+
+The `duration(long, string)` function constructs a duration value. The string argument must be one of `"days", "hours", "minutes", "seconds", "milliseconds"`. Strict validation requires `long` and `string` to be literals, although evaluation/authorization support any appropriately-typed expressions. Values of type `duration` have the following methods:
```
emina, line 185: Technically, strict validation doesn’t need to require the `long` and `string` in this case to be literals.  And that’s because it’s possible to encode the meaning of this constructor to SMT for arbitrary (i.e., unknown) values.

We require arguments to `ip` and `decimal` to be literals only because it would be too expensive / infeasible to encode their behavior (parsing) on arbitrary (unknown) strings.

That said, we may want to make these literals anyway for consistency with other operators.  And, admittedly, making them literals for the purpose of strict validation would help make the analysis more efficient.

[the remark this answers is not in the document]
khieta: I also like the look of a single string that supports more complex durations. ISO does indeed provide a format for this, although it's a little ugly 😅 https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm 

Example: `P3Y6M4D` is 3 years, 6 months, and 4 days.

There is also a format for time intervals: https://en.wikipedia.org/wiki/ISO_8601#Time_intervals

text/0080-datetime-extension.md
apg: I think the usefulness is only in generated policies, but also agree with @khieta that the user could provide a date far into the future, or far into the past just as easily. I'm :-1: on adding `epoch_start` and `epoch_end`.

text/0080-datetime-extension.md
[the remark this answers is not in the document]
apg: There are things about ISO 8601 intervals that complicate things:

1. .5Y is valid.
2. They are context sensitive, meaning `P1M` might mean 28, 29, 30, or 31 depending on how it's used.

I'd suggest we support: 

- `nD`: is `n * 86400000` milliseconds
- `nH`: is `n * 3600000` milliseconds
- `nM`: is `n * 60000` milliseconds
- `nS`: is `n * 1000` milliseconds
- `nm`: is `n` milliseconds

We can then eliminate the `P` and `T` leaders for the individual parts, as they don't provide much benefit to us.

- `5D3H15M10S500m` -> `443710500ms` if I did my math correctly.

We can say that it's inspired by ISO 8601 but unambiguous.

text/0080-datetime-extension.md
emina: I like `durationBetween`.

philhassey: `durationBetween` doesn't make it very clear in which order the subtraction happens.  I'd rather we had `.since` or `durationSince` .

text/0080-datetime-extension.md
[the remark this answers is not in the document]
khieta: This makes sense to me. I'm not a big fan of using (uppercase) M for minutes and (lowercase) m for milliseconds, but it's acceptable. Some alternatives are to 1) not support millisecond-level granularity, or 2) use multiple letters (e.g., `ms`) for milliseconds.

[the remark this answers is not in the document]
khieta: And third option: use `.` for milliseconds, so your example above would become `5D3H15M10.500S`

text/0080-datetime-extension.md
khieta: In my mind, `DT1.durationUntil(DT2)` maps to `DT2 - DT1`, `DT1.durationSince(DT2)` maps to `DT1 - DT2`, and `DT1.durationBetween(DT2)` maps to the absolute value of the difference.

Examples:
* `context.now.durationUntil(principal.startDate) > 0` - I'd expect the principal hasn't started yet
* `context.now.durationSince(principal.startDate) > 0` - I'd expect the principal already started
* `context.now.durationBetween(principal.startDate) > 0` - could be either

Given this, I think my preference is for `DT1.durationSince(DT2)` since that keeps things in the same order as `DT1 - DT2`. I'd also be open to supporting both `durationSince` and `durationUntil`.

apg: @khieta I agree, now that you point it out, that between implies absolute duration which is less desirable (at least in my opinion). :+1: for `durationSince`.

text/0080-datetime-extension.md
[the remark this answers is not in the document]
apg: I would definitely prefer the readability of lowercase, and don't mind `ms` at all. So, perhaps then:

* `nd`: is `n * 86400000` milliseconds
* `nh`: is `n * 3600000` milliseconds
* `nm`: is `n * 60000` milliseconds
* `ns`: is `n * 1000` milliseconds
* `nms`: is `n` milliseconds

`n` should be signed, since negative durations are a thing we're supporting.

I guess I'm not opposed to `n` being fractional, though it seems that this format is expressive enough, and easily readable enough that it shouldn't need to be. (Though, we _do_ have Decimal support so precision isn't as problematic...)

I added a duration parser (and switched the tests to use it) in my [python implementation](https://gist.github.com/apg/bc793d42f38d73a7d84ffa0fc309419f), and I like the way it feels to use.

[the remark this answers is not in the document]
emina: I’d also vote against fractional expressions.  Let’s start simple.  We can always add them later if needed.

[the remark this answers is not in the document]
apg: Updated the RFC with all of this!

text/0080-datetime-extension.md
mwhicks1: `durationSince` works for me. I prefer just having that, and not also having `durationUntil`, for simplicity and to hopefully avoid confusion.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,488 @@
...
+
+- Define `DT1.durationSince(DT2)` as the _absolute value_ of the difference between `DT1` and `DT2`.
+- Define `DT1.durationSince(DT2)` to be zero when `DT2 > DT1`.
+- Define `DT1.durationSince(DT2)` to be an error when `DT2 > DT1`.
+
+### Is millisecond precision enough? (Or too much?)
+
+The current proposal supports milliseconds. Do we want nanoseconds too? Or is just seconds sufficient? The ISO 8601 format does not specify a maximum precision, so we can technically allow any number of `S`s after the `.` in `YYYY-MM-DDThh:mm:ss.SSSZ`. Based on [this blog post](https://nickb.dev/blog/iso8601-and-nanosecond-precision-across-languages/), it appears that Javascript supports milliseconds (3 digits), Python supports microseconds (6 digits), and Rust and Go support nanoseconds (9 digits). Assuming nanosecond accuracy, the maximum (signed) 64-bit number (2^63 - 1) represents April 11, 2262. This date seems far enough out that any of these choices (milliseconds, microseconds, or nanoseconds) seems reasonable.
+
+Note that the backwards compatible option is to stick with milliseconds (or seconds). We can add precision later, but not remove it without a breaking change.
```
mwhicks1, line 488: In what sense is it backward compatible? The representation (as a `long`) would change its meaning, which means that data stored in the app would have to change. Are we expecting never to have to store a `long` (for duration or time) in the app? I was thinking "current time" or various deadlines would be stored that way.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,488 @@
...
+  action,
+  resource
+) when {
+  context.afterBrexit &&
+  context.location.countryOfOrigin == 'GB' &&
+  resource.owner == 'EU'
+}
+```
+
+## Unresolved questions
```
mwhicks1, line 440: Are these really "unresolved"? Maybe rename to "discussed questions" as a nicely formatted summary of the discussion? Might also be nice to summarize other topics here.

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,488 @@
...
+Currently in Cedar, these operations are only supported for `long`-typed values.
+For other extension types with similar operations, Cedar instead uses extension functions (e.g., `.lessThan()` for decimal values).
+
+This RFC proposes to _reverse_ this decision, and instead allow using builtin operators for extension functions, as appropriate.
+This will add some implementation complexity (at least in the primary [Rust implementation](https://github.com/cedar-policy/cedar)), but it will make policies that use these operations easier to read and easier to write.
+
+### Errors
+
+All the extension functions proposed in this RFC will throw a type error at authorization time if called with the wrong type of arguments.
+Additionally, the `datetime` and `duration` constructors will return an error if the input string does not match the expected format, or if the internal representation of the value (a 64-bit signed int) would overflow. `.offset(duration)` will return an error if the resulting datetime would overflow.
```
mwhicks1, line 224: Probably you should say somewhere that strict validation will require passing literals to `duration` and `datetime` constructors, and it will raise errors if those strings are malformed. Otherwise, validation is straightforward.

mwhicks1 (approved):
A few suggestions for final changes. Looks great!

text/0080-datetime-extension.md
khieta: Current time will need to be passed as a ISO date string with the current proposal. Users should not use the Unix `long` value directly.

Increasing precision is backwards compatible in the sense that all ISO date strings that were previously accepted will continue to be accepted in the future. But you are right that this isn't truly backwards-compatible: more precision will affect the underlying representation, which will impact when certain operators will overflow. My comment was sweeping this under the rug 🙂

andrewmwells-amazon: Note that there isn't a (nice) way to make it backwards compatible for analysis because extending the bitvector representation won't correspond to exactly to N extra decimal digits.

andrewmwells-amazon (approved):

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,470 @@
...
+
+The `datetime` type does not provide a way for a policy author to create a `datetime` from a numeric timestamp. One of the readable formats listed above must be used, instead.
+
+Values of type `datetime` have the following methods:
+
+- `.offset(duration)` returns a new `datetime`, offset by duration.
+- `.durationSince(DT2)` returns the difference between `DT` and `DT2` as a `duration`. (Note that the inverse of `durationSince` is `DT1.offset(duration)`).
+   An invariant for `DT1.durationSince(DT2)` is that when `DT1` is before `DT2` the resulting duration is negative.
+- `.toDate()` returns a new `datetime`, truncating to the day, such that printing the `datetime` would have `00:00:00` as the time.
+- `.toTime()` returns a new `duration`, removing the days, such that only milliseconds since `.toDate()` are left. This is equivalent to `DT - DT.toDate()`
```
cdisselkoen, line 148: > This is equivalent to `DT - DT.toDate()`

I think this should be rephrased using `.durationSince()`?

text/0080-datetime-extension.md
```diff
@@ -0,0 +1,470 @@
...
+}
+```
+
+## Discussion notes
+
+### Local time
+
+The current proposal has no direct support for local time, or time zones outside of UTC. Fundamentally, adding arbitrary time zone support would complicate the formal model, and ruin the project's SMT-based analysis goals.
+
+Policy writers can use the `datetime.offset()` method to simulate local time. Within the application, time zone data can be used to construct a `duration` that represents the time zone offset in hours and minutes, and provide that through `context` or another entity. Consider the policy below that checks if a principal's local time is between 09:00 and 17:00.
```
cdisselkoen, line 442: The discussion here ignores that above, `datetime()` is proposed to actually support a `+/-hhmm` offset.  This is worth mentioning in this discussion; while it doesn't amount to full support for local time, it does allow users to clearly specify times in (say) UTC-5.

cdisselkoen (approved):
Thanks for the team effort on this RFC.

text/0080-datetime-extension.md
apg: Ohhhh! Good catch, thanks!

text/0080-datetime-extension.md
apg: Updated: https://github.com/strongdm/cedar-rfcs/blob/datetime-rfc/text/0080-datetime-extension.md#local-time

mwhicks1 (approved):

apg:
@philhassey was questioning the special way to compute `durationSince` that's listed in the RFC, and it turns out that's completely wrong. I believe I convinced myself that that was necessary by having a bug in my reference implementation. Updating the RFC with that mistake removed.

emina (approved):

shaobo-he-aws (approved):
LGTM.

aaronjeline (approved):

khieta:
The final comment period (FCP) for this RFC is starting now, with intent to accept the RFC. The FCP will end 2024-09-10 at noon PT / 3pm ET / 8pm UTC. Please add comments, and especially any objections, if you have any. For more on the RFC process, see https://github.com/cedar-policy/rfcs.

PR #83, opened 2024-09-20
Type-aware Partial Evaluation

shaobo-he-aws:
[Rendered](https://github.com/cedar-policy/rfcs/blob/type-aware-pe/text/0083-type-aware-partial-evaluation.md)


shaobo-he-aws:
Closed in favor of #95

PR #84, opened 2024-10-18
Amend RFC 80

shaobo-he-aws:
This PR amends RFC 80 in two ways.

1. Introduce a new constructor for `datetime`.
2. Do not overload comparison operators. Instead, use methods like the `decimal` extension.


text/0080-datetime-extension.md
```diff
@@ -150,14 +155,14 @@ Values of type `datetime` have the following methods:
 - `.toDate()` returns a new `datetime`, truncating to the day, such that printing the `datetime` would have `00:00:00` as the time.
 - `.toTime()` returns a new `duration`, removing the days, such that only milliseconds since `.toDate()` are left. This is equivalent to `DT.durationSince(DT.toDate())`
 
-Values of type `datetime` can be used with comparison operators:
+Values of type `datetime` can be compared using the following functions:
+
+- `DT1.before(DT2)` returns `true` when `DT1` is before `DT2`
```
aaronjeline, line 160: I liked operator overloading, what is the justification for removing it? I think putting both of those in one PR is going to muddle things?

aaronjeline (changes requested):

text/0080-datetime-extension.md
emina: +1 to Aaron's comment.  We shouldn't have two operators that do the same thing.

andrewmwells-amazon: IIRC from the meeting, we sort of agreed to implement RFC 80 without operator overloading and then come back and add overloading for both datetime and decimal. At least personally, I don't think doing that requires amending the RFC because the end goal remains unchanged.

text/0080-datetime-extension.md
```diff
@@ -140,7 +140,12 @@ The `datetime(string)` function constructs a datetime value. Like with other ext
 - `"YYYY-MM-DDThh:mm:ss(+/-)hhmm"` (With timezone offset in hours and minutes)
 - `"YYYY-MM-DDThh:mm:ss.SSS(+/-)hhmm"` (With timezone offset in hours and minutes and millisecond precision)
 
-The `datetime` type does not provide a way for a policy author to create a `datetime` from a numeric timestamp. One of the readable formats listed above must be used, instead.
+The date and time must be valid. A valid date has `MM` between `1` and `12`, and `DD` between `1` and `31`, depending on the month/year. A valid time has `hh` between `0` and `23`, `mm` between `0` and `59`, and `ss` between `0` and `59`. In other words, a leap second is not representable. The hour offset is between `0` and `11`, even though an offset value greater than `11` can be handled with modulo operations.
```
andrewmwells-amazon, line 143:
```suggestion
The date and time must be valid. A valid date has `MM` between `1` and `12` (inclusive), and `DD` between `1` and `28-31` (inclusive), depending on the month/year. A valid time has `hh` between `0` and `23`, `mm` between `0` and `59`, and `ss` between `0` and `59`. In other words, a leap second is not representable. The hour offset is between `0` and `11`, even though an offset value greater than `11` can be handled with modulo operations.
```

andrewmwells-amazon, line 143: It's still not clear to me what the issue with representing a leap second is. As long as we can map the leap second into "Cedar time," what issues arise?

text/0080-datetime-extension.md
shaobo-he-aws: I just wanted to call out that the implementation will not do operator overloading. But rather another and later PR will do it.

text/0080-datetime-extension.md
shaobo-he-aws: I think it's kind of hard to map time with leap seconds into Cedar time, if the latter refers to Unix timestamps. We need to know whether the leap second is legal (i.e., exists). And it complicates the arithmetic. For instance, `end_of_leap_second.durationSince(start_of_leap_second)` should evaluate to 0 or 1s?

text/0080-datetime-extension.md
emina: I'm fine with not supporting overloading (now or ever).

But in that case, we should amend this RFC to take out all mentions overloading in the main body and in examples, and maybe add it as an `Alternative` at the bottom.

This is important because we'll also need to write documentation for Datetime, and in the docs, we have to commit to some way of representing comparison.

If overloading is coming at some undefined later point, I think it merits a separate RFC.

shaobo-he-aws: The updated version removes operator overloading, which will be described in its own RFC.

aaronjeline: I'm saying leave operator overloading in the RFC. (As strongdm has already implemented it that way, and we want that way to be standard.)

cdisselkoen: I agree with @aaronjeline.  We should leave operator overloading in the RFC.  That doesn't mean it has to be in the same _PR_ that the other datetime operations are added in; for all I care, we could have a separate PR per datetime operation if the implementer thought that made sense organizationally.  But when we stabilize the datetime feature and consider the RFC complete, it should include operator overloading.  (Among other reasons, because StrongDM has done it that way in the Go implementation.)

emina: If StrongDM has already implemented overloading, then I agree with should do the same.

My main objection is to having _both_ overloading and methods.  I think that's confusing and unnecessary.  I would have been ok with either, but existing implementation takes precedence.

On a related note, we should write a separate RFC for supporting overloading for Decimals and deprecating / removing the comparison methods.  As a user, I'd find it weird / counterintuitive that I can compare Dates with operators but not Decimals.

shaobo-he-aws: To sum up, there is no clear winner between operator overloading and non-overloading, as indicated by the table below.

```
|             | overloading                   | non-overloading |
|-------------|-------------------------------|-----------------|
| consistency | N (or Y with breaking change) | Y               |
| elegancy    | Y                             | N               |
```

To make the implementation both consistent and elegant, we should make breaking changes to the `decimal` extension, which will be another RFC, as Emina pointed out. If that's the case, I don't think it hurts to disallow operator overloading for this RFC and propose a new RFC to allow it for all extensions.

aaronjeline: Again, I disagree. Leaving aside decimal for a second, we already agreed to support operator overloading, and we want to support it eventually, and someone else has already implemented it based on that decision, so why remove it from the RFC?
If we think having both overloaded operators and methods is confusing duplication, then we shouldn't add methods.

john-h-kastner-aws: I agree with Aaron on this. We accepted the RFC as written, an implementation exists according to that spec, and we intended to support it as written eventually, so removing it from the RFC is silly.

It's fine to merge a partial implementation of the RFC while creating an issue to finish the implementation.

text/0080-datetime-extension.md
```diff
@@ -33,7 +33,7 @@ permit(
 when {
   principal.department == "HardwareEngineering" &&
   principal.jobLevel >= 10 &&
-  context.now.timestamp.durationSince(principal.hireDate) > duration("365d")
+  context.now.timestamp.durationSince(principal.hireDate).longerThan(duration("365d"))
```
john-h-kastner-aws, line 36: Do we intended to even release a version of the language with this methods? Would we want to keep the methods around after introducing operator overloading? I guess I don't understand the scope the work required to support operator overloading, but I don't like adding functions we don't actually want just to put off some development work.

shaobo-he-aws: It seems currently the implementation burden is something like this: https://github.com/cedar-policy/cedar/pull/1276/commits/f7c5711103ee493c9770ee93a415d0885e65735e. If it looks reasonable to you, we can just go ahead to implement operator overloading.

text/0080-datetime-extension.md
apg: In Unix time, isn't the answer 0? It replays the second, effectively ignoring it.

shaobo-he-aws: Yes. But it's still impossible to tell if an instant representing a leap second is valid or not.

shaobo-he-aws:
Closed now given new info.

PR #89, opened 2024-11-25
`datetime` extension: Adjust `duration` to account for zero-amount quantities

adpaco-aws:
@emina noted in [this comment](https://github.com/cedar-policy/cedar-spec/pull/471) that the specification for `duration` might be wrong, as the set of positive integers does not include zero. But we should be able to specify zero-amount units and durations.

[Rendered](https://github.com/USERNAME/REPONAME/blob/BRANCHNAME/text/FILENAME.md)


john-h-kastner-aws (approved):

cdisselkoen (approved):

apg:
Good catch. This was definitely intended to be positive or zero--cedar-go supports durations of 0.

PR #94, opened 2025-01-28
Set additional restrictions on datetime string formats

adpaco-aws:
Update the RFC for the `datetime` extension to set additional restrictions on datetime string formats.

These restrictions are based on [this comparison table](https://github.com/cedar-policy/cedar-spec/pull/519#issuecomment-2613547897) which made clear some discrepancies between Cedar implementations and the common restrictions in popular datetime frameworks.

[Rendered](https://github.com/adpaco-aws/rfcs/blob/datetime-restrictions/text/0080-datetime-extension.md)

### DIFF
--- a/text/0080-datetime-extension.md
+++ b/text/0080-datetime-extension.md
@@ -225,7 +225,10 @@ For example: `context.now.offset(duration("-3d"))` expresses "three days before 
 ### Errors
 
 All the extension functions proposed in this RFC will throw a type error at authorization time if called with the wrong type of arguments.
-Additionally, the `datetime` and `duration` constructors will return an error if the input string does not match the expected format, or if the internal representation of the value (a 64-bit signed int) would overflow. `.offset(duration)` will return an error if the resulting datetime would overflow.
+Additionally, the `datetime` and `duration` constructors will return an error if the input string does not match the expected format, or if the internal representation of the value (a 64-bit signed int) would overflow.
+The `datetime` constructor will also return an error if the input string refers to a non-existent calendar date (e.g., `"2025-02-31"`).
+Unlike some datetime libraries that perform [overflow normalization](#overflow-handling), this extension explicitly rejects such invalid dates.
+`.offset(duration)` will return an error if the resulting datetime would overflow.
 
 As noted above, strict validation will require passing literals to the `duration` and `datetime` constructors, and it will raise an error if those strings are malformed. Otherwise, validation is straightforward.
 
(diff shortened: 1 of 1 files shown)


emina (approved):
I agree with this more restricted interpretation of datetime strings.  

I'm curious to hear from @apg about the impact on the Go implementation.

apg:
The Go implementation doesn't have restrictions other than number of digits, but I don't see a problem with this proposal. The _major_ problem here is that, regardless of being able to accept up to `day = 31`, `2025-02-31` is never a valid date. The function that we use in Go would normalize that to `2025-03-03`, and will similarly normalize `2025-02-41` to `2025-03-13`. 

If we're going to specify valid ranges for the numeric components, we should probably also specify how overflow is handled. I'd argue that the behavior in the Go implementation makes more sense than erroring.

emina:
> If we're going to specify valid ranges for the numeric components, we should probably also specify how overflow is handled. I'd argue that the behavior in the Go implementation makes more sense than erroring.

Can you say a bit more about why you think that it makes more sense to give an interpretation to invalid date-time strings like `2025-03-03` rather than reject them? 

Interpreting these strings tricky because there is no obvious correct interpretation.  For example, why is `2025-03-03` the right choice for overflow behavior for `2025-02-31` instead of, say, `2025-02-28` (which can be thought of performing a floor function on dates)?

I find it more intuitive to reject invalid strings with an error.  If a policy contains `2025-02-31`, I'd view this as an error.  That also seems to be the consensus (at least for invalid minutes) among various libraries Adrian tested (see [here](https://github.com/cedar-policy/cedar-spec/pull/519#issuecomment-2613547897)).

It's worth noting that, under the proposed restriction, we wouldn't have to wait until evaluation time to detect malformed dates in policies:  the validator would detect and reject policies that contain `2025-02-31`.  This seems desirable to me.

emina:
@adpaco-aws Can you run a few more tests on invalid strings, like `2025-02-31`, and see what the various implementations do?

adpaco-aws:
> @adpaco-aws Can you run a few more tests on invalid strings, like `2025-02-31`, and see what the various implementations do?

Yes, I've run a couple more tests:
 * `2025-02-31` which is the one you requested
 * `2024-02-29` which is a valid date since 2024 was a leap year

Here's the new rows in the table for your convenience:

| Test Case | Description | Rust | Lean | Go | Python | TypeScript |
|-----------|-------------|------|------|-----|---------|------------|
| 10 | Valid leap year: `2024-02-29`  | ✅ | ✅ | ✅ | ✅ | ✅ |
| 11 | Invalid leap year: `2025-02-31` | ❌ | ❌ | ✅ | ❌ | ✅ |

The `2025-02-31` test case is rejected in Rust, Lean and Go. In TypeScript, it follows the behavior that @apg describes and results in `2025-03-03`.

> I'd argue that the behavior in the Go implementation makes more sense than erroring.

To me, it makes more sense to reject the date. In these cases, the validation can be done offline so there isn't a technical challenge as with leap seconds. The alternative would be to handle the overflow, and again there are multiple choices to do so like @emina pointed out. I'm afraid that, regardless of how it's handled, the invalid date will be ignored and left as is.

The wrapping behavior, in particular, seems too risky to accept in an authorization system: Let's assume a policy giving access to a principal until that date. In practice, the principal will have access to those resources until later than expected. I'd expect that to be a problematic situation in many cases.

That said, I wish my car insurance worked like that. I purchased 6 months of insurance on 2024-08-30 and it will expire on 2024-02-28. I certainly didn't expect this truncation neither...

adpaco-aws:
I've added some content to specify how non-existent calendar dates are handled. In summary, the current proposal is to reject them.

emina (approved):

PR #95, opened 2025-02-05
RFC for simple type-aware partial evaluation.

emina:
[Rendered](https://github.com/cedar-policy/rfcs/blob/emina/tpe/text/0095-type-aware-partial-evaluation.md)

### DIFF
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -13,7 +13,7 @@
 
 ## Summary
 
-This RFC proposes a simple _type-aware partial evaluator_ (TPE) for Cedar. Cedar’s current partial evaluator (CPE) is untyped and as a result, it can produce an ill-typed residual when applied to a well-typed policy and partial input. The lack of type awareness also prevents CPE from safely performing basic optimizations such as reducing `true && context.isPrivate` to `context.isPrivate`. TPE addresses these limitations with a simplified design that handles key use cases for partial evaluation such as permission queries.
+This RFC proposes a simple _type-aware partial evaluator_ (TPE) for Cedar. Cedar’s current partial evaluator (CPE) is untyped and as a result, it can produce an ill-typed residual when applied to a well-typed policy and partial input. The lack of type awareness also prevents CPE from safely performing basic optimizations such as reducing `true && context.isPrivate` to `context.isPrivate`. TPE addresses these limitations with a simplified design that handles key use cases for partial evaluation such as [permission queries](#permissions-queries-with-tpe).
 
 ## Basic example
 
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -246,7 +246,7 @@ structure PartialEntityData where
   tags : Option (Map String Value)   -- Tags are fully known or fully unknown.
 ```
 
-In JSON, we describe partial entities as a list of JSON objects whose known parts are specified using the [Cedar JSON format for entities](https://docs.cedarpolicy.com/auth/entities-syntax.html#entities).
+In JSON, we describe partial entities as a list of JSON objects whose known parts are specified using the [Cedar JSON format for entities](https://docs.cedarpolicy.com/auth/entities-syntax.html#entities). Note that the ancestor hierarchy of an entity must be either unknown or fully specified.  So, the "parents" field of an entity cannot include an entity whose "parents" are unknown.
 
 For example, this partial entities list for our [basic schema](#basic-example) specifies the ancestors for the entity `Document::"Manual"` and leaves the attributes unknown:
 
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -343,7 +343,7 @@ entity UnknownIP {
 };
 ```
 
-Next, we change the type of `context` for the `delete` action to use the new entity type for `srcIP`:
+Next, we change the type of `context` for the `Delete` action to use the new entity type for `srcIP`:
 
 ```
 action Delete appliesTo {
@@ -407,7 +407,7 @@ TPE returns the following residual that shows the value of Alice's source IP is 
 ```
 permit (principal, action, resource)
 when {
-  UnknownIP:"AliceIP".isInRange(ip("1.1.1.0/24"))
+  UnknownIP:"AliceIP".value.isInRange(ip("1.1.1.0/24"))
 };
 ```
 
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -407,7 +407,7 @@ TPE returns the following residual that shows the value of Alice's source IP is 
 ```
 permit (principal, action, resource)
 when {
-  UnknownIP:"AliceIP".value.isInRange(ip("1.1.1.0/24"))
+  UnknownIP::"AliceIP".value.isInRange(ip("1.1.1.0/24"))
 };
 ```
 
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -448,9 +448,22 @@ There are four main drawbacks to this RFC:
 
 ## Alternatives
 
-An alternative approach would be to build on CPE's existing interface for specifying unknowns, while adding type awareness. Like TPE, this approach would require policies and unknowns to be typed, so users would still lose the ability to partially evaluate untyped policies.
+We considered two alternatives to TPE, each adding different levels of type awareness to CPE:
+1. A lightweight approach that adds partial type awareness to CPE's algorithm.
+2. A comprehensive approach that adds full type awareness to both CPE's algorithm and interface.
 
-The key advantage of this approach would be a smoother transition for current CPE users, especially those with advanced use cases not directly supported by TPE. The interface would also be more flexible, allowing unknowns to be specified anywhere in the request and entity data.
+### Adding partial type-awareness to CPE
 
-But this flexibility comes at the cost of significantly more complex implementation, models, and proofs. This added complexity would support advanced functionality that most users may not need. If we do find that a CPE-like interface is needed in the future, we can implement it as a [transformation layer](#future-improvements) on top of TPE.
+A lightweight approach would modify CPE's algorithm to take advantage of operator-specific type information. For example, CPE could safely drop `true` from the expression `true && resource.owner == principal` by tracking that the equality operator always produces boolean results (or errors).
+
+This approach would be easier to implement than TPE while enabling more optimizations than CPE. But like CPE, it could still return ill-typed residuals, and it would miss optimizations that require full type awareness. For example, `true && context.isPrivate` cannot be simplified without a schema because `context.isPrivate` might not be boolean.
+
+
+### Adding full type-awareness to CPE
+
+A more comprehensive approach is to build on CPE's existing interface for specifying unknowns, while adding full type awareness to the underlying algorithm. Like TPE, this approach would require policies and unknowns to be typed, so users would lose the ability to partially evaluate untyped policies.
+
+The key advantage would be a smoother transition for current CPE users, especially those with advanced use cases not directly supported by TPE. The CPE-style interface would remain more flexible, allowing unknowns to be specified anywhere in the request and entity data.
+
+However, this approach would require significantly more complex implementation, models, and proofs to support advanced functionality that most users may not need. If we do find that a CPE-style interface is needed in the future, we can implement it as a [transformation layer](#future-improvements) on top of TPE.
 
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -261,6 +261,8 @@ For example, this partial entities list for our [basic schema](#basic-example) s
 
 There are two key differences between the TPE and CPE interfaces for partial entities. First, TPE allows mixing known and unknown parts of entity data; for example, an entity can have known attributes but unknown ancestors. This is important for applications that implement role or relation-based access control using Cedar's hierarchy, where entities typically have a few simple attributes but potentially large ancestor sets. In such cases, TPE can attempt to authorize requests using just the attributes, avoiding the cost of loading the full ancestor set. Second, entity attributes must be either unknown or fully specified. TPE doesn't support partially known attributes, following the same design principle as with `context` data.
 
+TPE treats absent entities as present entities with unknown contents. In other words, omitting an entity from the partial entity store is equivalent to including it but leaving all of its parts as unknown.  This is convenient for partially evaluating conditions such as `User::"Alice" in principal.friends` without having to determine which literals are referenced in the policies and then map them to unknown contents.
+
 ### Permissions queries with TPE
 
 TPE provides native support for basic permission queries:
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -329,7 +329,7 @@ when {
 };
 ```
 
-The residual says that Alice can perform the delete action if her source IP address is in the given range. In contigent authorization, an application would examine this residual, take further steps to determine Alice's IP address, and re-authorize against the residual to reach a concrete authorization decision.
+The residual says that Alice can perform the delete action if her source IP address is in the given range. In contingent authorization, an application would examine this residual, take further steps to determine Alice's IP address, and re-authorize against the residual to reach a concrete authorization decision.
 
 #### Contingent authorization with entity-based unknown values
 
--- a/text/0095-type-aware-partial-evaluation.md
+++ b/text/0095-type-aware-partial-evaluation.md
@@ -1,0 +1,456 @@
+# Simple Type-Aware Partial Evaluation
+
+## Related issues and PRs
+
+- Reference Issues: N/A
+- Implementation PR(s):
+
+## Timeline
+
+- Started: 2025-01-30
+- Accepted: TBD
+- Stabilized: TBD
+
+## Summary
+
+This RFC proposes a simple _type-aware partial evaluator_ (TPE) for Cedar. Cedar’s current partial evaluator (CPE) is untyped and as a result, it can produce an ill-typed residual when applied to a well-typed policy and partial input. The lack of type awareness also prevents CPE from safely performing basic optimizations such as reducing `true && context.isPrivate` to `context.isPrivate`. TPE addresses these limitations with a simplified design that handles key use cases for partial evaluation such as permission queries.
+
+## Basic example
+
+Consider the following Cedar policies:
+
+```
+// Users can view public documents.
+permit (
+  principal,
+  action == Action::"View",
+  resource
+) when {
+  resource.isPublic
+};
+
+// Users can view owned documents if they are mfa-authenticated.
+permit (
+  principal,
+  action == Action::"View",
+  resource
+) when {
+  context.hasMFA &&
+  resource.owner == principal
+};
+
+// Users can delete owned documents if they are mfa-authenticated
+// and on the company network.
+permit (
+  principal,
+  action == Action::"Delete",
+  resource
+) when {
+  context.hasMFA &&
+  resource.owner == principal &&
+  context.srcIP.isInRange(ip("1.1.1.0/24"))
+};
+```
+
+These policies are well-typed with respect to the following schema:
+
+```
+entity User;
+
+entity Document  = {
+  "isPublic": Bool,
+  "owner": User
+};
+
+action View appliesTo {
+  principal: [User],
+  resource: [Document],
+  context: {
+    "hasMFA": Bool,
+  }
+};
+
+action Delete appliesTo {
+  principal: [User],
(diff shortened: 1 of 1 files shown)


text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,456 @@
...
+permit (principal, action, resource)
+when {
+  true &&
+  true &&
+  true &&
+  context.srcIP.isInRange(ip("1.1.1.0/24"))
+};
+```
+
+The residual says that Alice can perform the delete action if her source IP address is in the given range. In contigent authorization, an application would examine this residual, take further steps to determine Alice's IP address, and re-authorize against the residual to reach a concrete authorization decision.
```
cdisselkoen, line 332:
```suggestion
The residual says that Alice can perform the delete action if her source IP address is in the given range. In contingent authorization, an application would examine this residual, take further steps to determine Alice's IP address, and re-authorize against the residual to reach a concrete authorization decision.
```

text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,456 @@
...
+  },
+  {
+    "uid": { "type": "Document", "id": "SuperSecret" },
+    "attrs": {
+      "isPublic": false,
+      "owner"   : { "type": "User", "id": "Alice" }
+    },
+    "parents": [ ]
+  }
+]
```
cdisselkoen, line 402: Nit/technicality: wouldn't we need an entry for `UnknownIP::"AliceIP"` in the entity data (with unknown attributes)?  Does this RFC propose that completely omitting an entity from the entity data is allowed, and equivalent to specifying all of its data as unknown?  Are there footguns that could result?

cdisselkoen (approved):
I'm in support of this change.

text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,456 @@
...
+
+The residual is also not fully simplified, including the trivial constraint `true` that is the result of partially evaluating the scope constraint `action == Action::"PickUp"`. Ideally, `true && expr` should be simplified to just `expr`, but CPE cannot perform this simplification safely. The reason is that CPE lacks type information and cannot guarantee that `expr` will always evaluate to a boolean value. If `expr` could potentially evaluate to a non-boolean value, then simplifying `true && expr` to `expr` would change the error behavior of the original policy. To preserve the original policy's semantics, including potential type errors, CPE must conservatively leave expressions like `true && expr` unsimplified. This limitation leads to complex residuals that are harder to apply for solving use cases like permission queries.
```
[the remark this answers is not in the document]
emina: We have the `true && context.isPrivate` in the [summary](https://github.com/cedar-policy/rfcs/blob/emina/tpe/text/0095-type-aware-partial-evaluation.md#summary). 

We could repeat something like it here but I worry the example might get harder to follow.  

Note also that if we modified CPE to make type inferences like this, we would end up re-implementing parts of the Cedar type checker.  For example, we could also safely assume that `e1 < e2` is always either error or boolean, so it’s safe to simplify `true && e1 < e2` etc.  So, CPE would be trying to get closer to TPE.

text/0095-type-aware-partial-evaluation.md
emina: Yes, I was thinking we would allow an entity to be fully omitted from the entity store, which would be treated equivalently as specifying that all of the entity’s data is unknown.  This would be useful / convenient for evaluating policy conditions have constraints like `User::”Alice” in principal.friends`. 

No footguns immediately come to mind, but I’ll keep thinking about it.

text/0095-type-aware-partial-evaluation.md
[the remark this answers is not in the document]
cdisselkoen: Yes, but just adding a hardcoded list of operators which are known to always return error or boolean, is a lot less than full typechecking and would be a lot less implementation work than this RFC.  I'm just pointing out this alternative.  I'm still in favor of this RFC in general.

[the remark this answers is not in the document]
emina: Sounds good! I'll work on adding a section to the "Alternatives" that explains this option.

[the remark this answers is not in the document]
emina: New alternative added: https://github.com/cedar-policy/rfcs/blob/emina/tpe/text/0095-type-aware-partial-evaluation.md#alternatives

text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,469 @@
...
+
+## Timeline
+
+- Started: 2025-01-30
+- Accepted: TBD
+- Stabilized: TBD
+
+## Summary
+
+This RFC proposes a simple _type-aware partial evaluator_ (TPE) for Cedar. Cedar’s current partial evaluator (CPE) is untyped and as a result, it can produce an ill-typed residual when applied to a well-typed policy and partial input. The lack of type awareness also prevents CPE from safely performing basic optimizations such as reducing `true && context.isPrivate` to `context.isPrivate`. TPE addresses these limitations with a simplified design that handles key use cases for partial evaluation such as permission queries.
```
shaobo-he-aws, line 16: permission queries can be an unknown concept to broad audience. Maybe we should briefly explain it here.

emina: We explain it in detail later in the RFC; I can add a link here to the later section (so that we can keep the summary short).

text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,469 @@
...
+
+structure PartialEntityData where
+  attrs : Option (Map Attr Value)    -- Attributes fully known or fully unknown.
+  ancestors : Option (Set EntityUID) -- Ancestors fully known or fully unknown.
+  tags : Option (Map String Value)   -- Tags are fully known or fully unknown.
+```
+
+In JSON, we describe partial entities as a list of JSON objects whose known parts are specified using the [Cedar JSON format for entities](https://docs.cedarpolicy.com/auth/entities-syntax.html#entities).
+
+For example, this partial entities list for our [basic schema](#basic-example) specifies the ancestors for the entity `Document::"Manual"` and leaves the attributes unknown:
```
shaobo-he-aws, line 251: Nit: the `tags` field is currently [optional](https://github.com/cedar-policy/cedar/blob/e563cc0c859a4ffbc23a20388590bc3907a82219/cedar-policy-core/src/entities/json/entities.rs#L60). Absence of `tags` means that it's empty. If we want to mark `tags` optional, we may have to use `null`.

text/0095-type-aware-partial-evaluation.md
emina: Done!

text/0095-type-aware-partial-evaluation.md
emina: Ah, we don't need `null` for this in the interface (though we may in the implementation).  That's because we can tell from the schema whether an entity type has `tags` or not.  If it is declared to have `tags`, then omitted `tags` field means it's unknown---empty tags are just an empty map.  If not, then `tags` must be absent anyway.

shaobo-he-aws: Say the schema is like this,

```
entity A tags Long;
```

And JSON entities,

Case 1
```
{ 
  "uid" : ...,
   "tags": []
}
```

Case 2
```
{ 
  "uid" : ...
}
```

Case 1 means `tags` is known and empty. Case 2 means `tags` is unknown according to the proposal but the current deserializer treats it as case 1. So if we want to keep the proposal, it looks like a breaking change.

text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,469 @@
...
+  }
+]
+```
+
+TPE returns the following residual that shows the value of Alice's source IP is needed to reach a decision:
+
+```
+permit (principal, action, resource)
+when {
+  UnknownIP:"AliceIP".value.isInRange(ip("1.1.1.0/24"))
```
shaobo-he-aws, line 410: Nit: `:` -> `::`

shaobo-he-aws (approved):
Looks great!

text/0095-type-aware-partial-evaluation.md
emina: Ah, interesting.  I just assumed that the deserializer requires the `tags` field to be present (and empty) when it's declared in the schema.  

Do we allow other kinds of absent fields to be interpreted as empty?  

For example, if an entity declares no attributes, can you just omit `attrs` or are you supposed to provide the empty object?

shaobo-he-aws: IIUC, the `tags` field is different from all other fields. In other words, users have to provide fields `parents` and `attrs` anyways. @cdisselkoen to double check.

emina: In that case, I'd vote for making it behave like `parents` and `attrs`.  If `tags` are present in the schema, then the field should be present in JSON.  The schema-based parser can enforce this though, of course, the regular one can't.  But that's okay for TPE, because we assume that the schema is present and can use schema-based parsing.

emina: (To be clear, throughout all this, I'm assuming we'd use schema-based parsing of entities for TPE.)

cdisselkoen: The reason it behaves differently was to avoid a breaking change in Cedar 4.  `parents` and `attrs` have always been required.  Making `tags` required would be a breaking change, all existing JSON entity data would no longer be valid.  Treating `tags` how we do (omitted = empty) provides backward compatibility with all JSON entity data that was valid in Cedar 3.

emina: @cdisselkoen I'm thinking that we'd still maintain backward compatibility with schema-based parsing if we require `tags` to be present _if_ they are declared in the schema.  That would be backward compatible since no schemas could have had `tags` declared before :)  Or maybe I'm missing some subtlety with schema-based parsing too?

(I totally agree that `tags` need to be optional for regular entity parsin.)

cdisselkoen: Cedar 4.0.0-4.3.1 allow `tags` to be absent even if they are declared in the schema; this is equivalent to an empty tags map.  Note that an empty tags map is perfectly legal even if the schema declares a type for tags for that entity.  We should not break backward compatibility with the behavior of Cedar 4.0.0-4.3.1.

emina: My thinking is that TPE would use schema-based parsing because we have and need the schema.  The regular parser would be unaffected (it wouldn't need to handle partial inputs).

cdisselkoen: My comment applies to schema-bsaed parsing.

text/0095-type-aware-partial-evaluation.md
luxas: This also came to mind when I read this proposal, I wasn't sure at first if the omission was intentional or not. Maybe worth calling out explicitly in the proposal (right before the `Permissions queries with TPE` section) that there might be references to entities that are not given, and they are then treated as "fully unknown".

luxas:
Really excited to see this 💯 🎉 !! This would help Kubernetes integration quite a bunch 👍 (see: https://github.com/awslabs/cedar-access-control-for-k8s/issues/11).

In particular, I proposed to @micahhausler we'd use partial eval to be able to use only one unified action (e.g. "just" `k8s::Action::"create"`, not both that and `k8s::admission::Action::"create"` like now), first with partial data (without any populated object) and later with object populated. 

This would rely directly on the special "feature" described in this proposal that instead of mixing unknown and known attributes within one entity, unknown attributes are factored out to its own typed entity to be populated later. 

This is enough for this use case (I think), we can embed this semantic directly into the schema at design time, no need for the CPE to TPE conversion layer.

I have one question: Will we remove the `unknown` function as a result, or make it result in an error in future versions of Cedar, when this is implemented? Today, one can e.g. populate an attribute value with the `__extn` JSON semantic; I guess we'd want the TPE to error if it saw partially populated attributes like that in the future.

Finally, did you consider already the interaction between this feature and entity slicing, and/or what the practical Rust interface would look like for reauthorizing a residual?

shaobo-he-aws:
> Finally, did you consider already the interaction between this feature and entity slicing, and/or what the practical Rust interface would look like for reauthorizing a residual?

My 2 cents. TPE should be compatible with the existing methods to perform entity slicing (i.e., entity manifest based and level based), since they all require validation.

I can't say for the Rust implementation. The Lean model asks the user to provide fully concrete requests and entities (i.e., those who do not contain any unknowns) to reauthorize a residual. And these requests and entity stores must be consistent with the partial requests and entities, which TPE checks. I imagine that Rust will provide a simpler interface but it will be tricky because there do not exist named unknowns in TPE, preventing us to adopt a similar interface to the existing PE that is based on bindings.

text/0095-type-aware-partial-evaluation.md
emina: Sounds good; will do!

emina:
> Really excited to see this 💯 🎉 !! This would help Kubernetes integration quite a bunch 👍 (see: [awslabs/cedar-access-control-for-k8s#11](https://github.com/awslabs/cedar-access-control-for-k8s/issues/11)).

Thank you for reading and commenting!  This is super helpful to know :) 

> I have one question: Will we remove the `unknown` function as a result, or make it result in an error in future versions of Cedar, when this is implemented? Today, one can e.g. populate an attribute value with the `__extn` JSON semantic; I guess we'd want the TPE to error if it saw partially populated attributes like that in the future.

That's a good question.  Since `unknown` was part of the current experimental PE feature, I think that it's safe to remove it without officially breaking backward compatibility.  But I defer to the Cedar team on that one. 

> Finally, did you consider already the interaction between this feature and entity slicing, and/or what the practical Rust interface would look like for reauthorizing a residual?

This should work with entity slicing, with minor adjustments.  We'd need to adjust the entity slicing algorithm to work on residual expressions as well as typed expressions (its current input).  Residual expressions are closely related to typed expressions so this is definitely doable.

luxas:
Another thing I came to think about, could we add a concrete example of how to authorize partial ancestors too?

Consider this schema and policy:
```
namespace partialancestors {
    entity Organization;
    entity Team in [Organization];
    entity User in [Team, Organization];
    entity Document;
    
    action "get" appliesTo {
        principal: User,
        resource: Document,
    };
}
```
```
permit (
    principal in partialancestors::Team::"foo",
    action == partialancestors::Action::"get",
    resource is partialancestors::Document
) when {
    principal in partialancestors::Organization::"org" &&
    principal in partialancestors::Team::"intermediate" &&
    principal in partialancestors::Team::"toplevel"
};
```
Then, I perform a request for principal `User::"lucas"`, which is known to have two ancestors: `Organization::"org"` and `Team:"a"`, but any other team membership is unknown (i.e. whether the user is part of team b)

Given entities 
```
[
    {
        "uid": {
            "type": "partialancestors::User",
            "id": "lucas"
        },
        "attrs": {},
        "parents": [
            {
                "type": "partialancestors::Team",
                "id": "a"
            },
            {
                "type": "partialancestors::Organization",
                "id": "org"
            }
        ]
    },
    {"uid": { "type": "partialancestors::Team", "id": "lucas"}},
    {"uid": { "type": "partialancestors::Organization", "id": "org"}}
]
```

it is known that the first two conditions are met, but not whether the third one holds. If the requester wants to indicate that parents of lucas are in fact partially unknown, I guess one would perform a schema rewrite such that there is an "intermediate parent layer" in the schema, e.g.
```
entity UserPartialParents in [Team, Organization];
entity User in [Team, Organization, UserPartialParents];
```

then still keep lucas' parents known, but refer to an entity that is unknown (not in the entity store), like this:
```
[
    {
        "uid": {
            "type": "partialancestors::User",
            "id": "lucas"
        },
        "attrs": {},
        "parents": [
            {
                "type": "partialancestors::Team",
                "id": "a"
            },
            {
                "type": "partialancestors::Organization",
                "id": "org"
            },
            {
                "type": "partialancestors::UserPartialParents",
                "id": "lucas"
            }
        ]
    },
    {"uid": { "type": "partialancestors::Team", "id": "lucas"}},
    {"uid": { "type": "partialancestors::Organization", "id": "org"}}
]
```

Now, Cedar cannot evaluate `User::"lucas" in partialancestors::Team::"toplevel"`, because `partialancestors::UserPartialParents::"lucas"` does not exist, and _could_ have a parent called `partialancestors::Team::"toplevel"` (in a transitive closure).

Was this kind of rewrite what you had in mind for ancestors/parents?

emina:
> Another thing I came to think about, could we add a concrete example of how to authorize partial ancestors too?

> Was this kind of rewrite what you had in mind for ancestors/parents?

That is an interesting question.  We haven't considered supporting partial ancestors, since this is not supported by the current PE either.  The transformation you suggest seems like it should work, but I'll need to think more about it.

emina:
> The transformation you suggest seems like it should work, but I'll need to think more about it.

Coming back to this idea, the suggested transformation wouldn’t work.  That’s because the partial evaluator, like the regular evaluator, represents the entity hierarchy as a map from an entity to its _ancestors_ (rather than parents).  That is, the input to the (partial) evaluator is the transitive closure of the hierarchy graph.  In JSON, we let you specify just the direct parents for convenience—so you don’t have to compute the transitive closure and pass it in.  But we do compute the transitive closure over the JSON graph before passing the final map from entities to ancestors to the (partial) evaluator.  This way, the (partial) evaluator can check `A in B` in constant time, simply by testing if `B` is a member of `A`’s ancestors set (and without having to perform a reachability check over the graph). 

So, this actually reveals a restriction that we need to put on the JSON partial input format:  entities with unknown parents must _not_ appear in the “parent” field of any other entity. 

With this in mind, there is a more aggressive transformation that would let us express something like this. 

First, change the schema to add `UnknownUserAncestors` like this:

```
entity UnknownUserAncestors in [Team, Organization];
entity User in [Team, Organization] {
  unknownAncestors: UnknownUserAncestors
};
```

Next, transform the policy as follows:
```

permit (
    principal,
    action == partialancestors::Action::"get",
    resource
) when {
    // We don’t know if Lucas is in Team “foo” or “topLevel”
    principal.unknownAncestors in partialancestors::Team::"foo" && 
    principal in partialancestors::Organization::"org" &&
    principal in partialancestors::Team::"intermediate" &&
    principal.unknownAncestors in partialancestors::Team::"toplevel"
};
```

Finally, invoke the partial evaluator with the following partial input:

```
[
    {
        "uid": {
            "type": "partialancestors::User",
            "id": "lucas"
        },
        “attrs”: {
            “unknownAncestors” : { “type” : “UnknownUserAncestors”, “id” : “LucasUnknownAncestors” }
         },
        "parents": [
            {
                "type": "partialancestors::Team",
                "id": "a"
            },
            {
                "type": "partialancestors::Organization",
                "id": "org"
            }
        ]
    },
   // Provide the parents for partialancestors::Organization::”org”, partialancestors::Team::”a” and their parents 
   // too, if any. 
]
```

Given this input, TPE will return the following residual:

```
permit (
    principal,
    action,
    resource
) when {
    UnknownAncestors::”LucasUnknownAncestors” in partialancestors::Team::"foo" &&
    UnknownAncestors::”LucasUnknownAncestors” in partialancestors::Team::"toplevel"
};
```

The alternative to all this would be for the partial evaluator to work directly on the parent relation, rather than the ancestor relation like the evaluator.  Then, it would be ok for an entity to have a parent whose parents themselves are unknown.  But it would be very difficult to establish the correctness of this algorithm with respect to the concrete evaluation semantics (which works on the transitive closure).  

@shaobo-he-aws See this discussion :).  It will be relevant for the proofs.  Specifically, we will need to require well-formed partial entities to enforce the constraint that if an entity UID with unknown ancestors cannot appear in the ancestor set of another entity.

luxas:
I guess this RFC has de-facto been approved, as it's formally specified (with no surprises that I saw/heard there) and being implemented at the moment. Let's merge it to signal approval 😄?

emina:
The RFC is now entering the final comment period (FCP) that will end on May 23 at 12 PM Pacific.

shaobo-he-aws (approved):

text/0095-type-aware-partial-evaluation.md
```diff
@@ -0,0 +1,471 @@
...
+```
+abbrev PartialEntities := Map EntityUID PartialEntityData
+
+structure PartialEntityData where
+  attrs : Option (Map Attr Value)    -- Attributes fully known or fully unknown.
+  ancestors : Option (Set EntityUID) -- Ancestors fully known or fully unknown.
+  tags : Option (Map String Value)   -- Tags are fully known or fully unknown.
+```
+
+In JSON, we describe partial entities as a list of JSON objects whose known parts are specified using the [Cedar JSON format for entities](https://docs.cedarpolicy.com/auth/entities-syntax.html#entities). Note that the ancestor hierarchy of an entity must be either unknown or fully specified.  So, the "parents" field of an entity cannot include an entity whose "parents" are unknown.
```
luxas, line 249: If you have the energy, you might want to put the rewrite that we discussed on the PR comments in an appendix, so it's easy for people to see/understand/discover it from the proposal. 😄

PR #98, opened 2025-05-19
Generalized Cedar Templates

alanwang67:
[rendered](https://github.com/alanwang67/rfcs/blob/main/text/0098-generalized-templates.md)


shaobo-he-aws:
It seems that there will be a language of type annotations. Can you specify the details of the language (e.g., the syntax of Cedar types, if it's related to a schema)? Can you also specify the grammar of the template annotations? And what if there are name conflicts (e.g., `template(?name: Bool, ?name: Long)`)?

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,275 @@
...
+
+#### Cedar Template 
+```
+template(?folder: FS::Folder) =>
+permit(
+  principal == ?principal,
+  action,
+  resource in ?resource)
+when {
+    ?resource.created_by == resource.created_by
```
shaobo-he-aws, line 68: It seems `created_by` is not specified in the schema and `?folder` is not used.

alanwang67: Thanks for pointing that out didn't catch that earlier. I have fixed example 1.

alanwang67:
> It seems that there will be a language of type annotations. Can you specify the details of the language (e.g., the syntax of Cedar types, if it's related to a schema)? 

The types provided by the user should be schema types since compared with AST types we are able to more precisely specify the types of Set and Record. In the implementation, I have a function that provided with a schema translates a schema type into a type checker type to perform validation. 

>  Can you also specify the grammar of the template annotations?

The grammar of template annotations is as follows. `Type` is taken directly from the cedar documentation for schema types.
```
Template := 'template' '(' ( TypeAnnotation { ',' TypeAnnotation } ) ')' '=>'

TypeAnnotation := ‘?’ IDENT ‘:’ Type

Type := Path | SetType | RecType
```

> And what if there are name conflicts (e.g., `template(?name: Bool, ?name: Long)`)?

If there are name conflicts, the template should not be allowed. Each new typed slot requires a distinct name. However, within the when/unless clause the same typed slot can be used multiple times.

> And should users provide values or expressions when binding template variables?

Users should provide values when binding template variables. I think it will be confusing if a user provides an expression and the expression doesn't evaluate to what the user expects it to when linking with a template. In the implementation, the type of the values provided by the user should be of `RestrictedExpression`.

shaobo-he-aws:
@alanwang67 thank you for your reply. Could you please incorporate your response to the RFC? Furthermore, I have another question that is if bind-time validation is too expensive. In other words, we validate linked policies and reject them if type checking fails? I feel like creation-time validation only buys us when the policies are very complicated and templates do not show up often.

alanwang67:
> Furthermore, I have another question that is if bind-time validation is too expensive. In other words, we validate linked policies and reject them if type checking fails? I feel like creation-time validation only buys us when the policies are very complicated and templates do not show up often.
> 
> My rationale is that type annotation may be too much a burden for both Cedar users and implementors. Bind-time validation save us from it at the cost of potential repeated validation work. 

I think with bind time validation we would still want some type of validation to be performed at creation time. If we didn't it would be possible to have templates where there is no possible substitution for the slots that results in a valid linked template. The lack of type annotations could also be a problem for analysis. To perform analysis without type annotations, we would likely have to consider all possible combinations of the types of the slots that will allow for link time validation to succeed. This means that what we could say about our policies would have to be generalized over these combinations which could potentially result in a less meaningful analysis. If that isn't a concern, then the question of eliminating type annotations comes down to how much of a burden type annotations are for Cedar users.

shaobo-he-aws:
> If we didn't it would be possible to have templates where there is no possible substitution for the slots that results in a valid linked template.

I don't think it a fundamental blocker because users will be likely inspect the policies when bind-time validation fails and correct the aforementioned issue.

> The lack of type annotations could also be a problem for analysis.

Good point. I wonder if type-inference could help us to reduce the combinations.

alanwang67:
> Good point. I wonder if type-inference could help us to reduce the combinations.

I am not sure if we would be able to infer the exact type. Take for example, the following template:

```
permit(principal, resource, action) when {
    ?slot.attribute
}
```
If multiple entities have this attribute in their schema with type `bool`, then any value that are of those entity types can be substituted and would result in a valid linked template. Analysis will have to consider all of these different entity types.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Timeline
+
+- Started: 2025-05-13
+- Accepted: TBD
+- Stabilized: TBD
+
+## Summary
+
+Cedar templates are quite restrictive. There are only two slots provided (```?principal``` and ```?resource```) and they are limited to appearing in the scope of the policy. This results in users' of Cedar needing to come up with their own [solution](https://github.com/cedar-policy/rfcs/pull/3#issuecomment-1632645305) to handling templates that can not be expressed in Cedar currently. In this RFC, we propose to generalize Cedar templates to support:
```
chaluli, line 19: Should be just "users" without apostrophe?

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+- Stabilized: TBD
+
+## Summary
+
+Cedar templates are quite restrictive. There are only two slots provided (```?principal``` and ```?resource```) and they are limited to appearing in the scope of the policy. This results in users' of Cedar needing to come up with their own [solution](https://github.com/cedar-policy/rfcs/pull/3#issuecomment-1632645305) to handling templates that can not be expressed in Cedar currently. In this RFC, we propose to generalize Cedar templates to support: 
+
+1. Allowing ```?principal``` and ```?resource``` in the condition of the policy only if they also appear in the scope as well.  
+2. Allowing for an arbitrary number of user defined slots to appear in the ```when/unless``` clause of the template contingent on the type being explicitly annotated. This would also include slots that are not entity types. 
+
+Both of these additions, do not seem to require substiantial changes to the existing type checking code. However, that may change upon diving deeper into the implementation.
```
chaluli, line 24: Broken sentence.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+
+1. Allowing ```?principal``` and ```?resource``` in the condition of the policy only if they also appear in the scope as well.  
+2. Allowing for an arbitrary number of user defined slots to appear in the ```when/unless``` clause of the template contingent on the type being explicitly annotated. This would also include slots that are not entity types. 
+
+Both of these additions, do not seem to require substiantial changes to the existing type checking code. However, that may change upon diving deeper into the implementation. 
+
+For policies without schemas generalized templates will be supported. Correct type annotations will need to be supplied by a user of Cedar and a link time check will be performed to make sure that instantiations of the template are values of that type. However, note that without a schema it is still possible for a program that passes the link time type check to exhibit runtime errors.
```
chaluli, line 26: Maybe strengthen / restructure these sentences for clarity?
Perhaps:
Generalized templates will be supported for policies without schemas. Templates will not be validated, but type checking will be performed during link-time. While this ensures that the provided values match the type of slot they are filling, it does not ensure that the policy won't experience a run-time type error.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+
+Both of these additions, do not seem to require substiantial changes to the existing type checking code. However, that may change upon diving deeper into the implementation. 
+
+For policies without schemas generalized templates will be supported. Correct type annotations will need to be supplied by a user of Cedar and a link time check will be performed to make sure that instantiations of the template are values of that type. However, note that without a schema it is still possible for a program that passes the link time type check to exhibit runtime errors. 
+
+## Basic example
+
+### Example 1 
+
+By allowing ```?principal``` and ```?resource``` in the condition of the policy if it also appears in the scope, it allows us to express templates where we want to access attributes of a slot. In this example since we are able to allow ```?resource``` in the condition of the policy we are able to express the constraint that we can only view resource's who's owner is the same as the owner of the resource that we instantiate for the typed hole.
```
chaluli, line 32: Perhaps "it allows us to ..." ==> "we can write templates that access attributes of a `?principal` or `?resource` slot."

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+action Access appliesTo {
+    principal: Company::Employee,
+    resource: Company::InternalServices,
+    context: { date: datetime }
+}
+```
+
+#### Template 
+```
+template(?date: datetime) =>
```
chaluli, line 130: I think this is a perfectly good example. But others might argue that this scenario doesn't require templates at all.

For instance, it could be achieved by requiring each principal have a `start_date` attribute. then changing the when clause to `context.date > principal.start_date`.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+  action == Action::"Access",
+  resource is Company::InternalServices
+) when {
+    context.date > ?date 
+};
+```
+
+### Example 4
+
+Suppose there are certain professors with dual appointments under two different departments. You want to give students of their group access to both the resources in the first department and the resources in their second department.
```
chaluli, line 142: Perhaps: "... under two different departments. You ..." ==> "... under two different departments who ..."

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+) when {
+    context.date > ?date 
+};
+```
+
+### Example 4
+
+Suppose there are certain professors with dual appointments under two different departments. You want to give students of their group access to both the resources in the first department and the resources in their second department. 
+
+Note: Although we can create two seperate templates and instantiate them individually, by using generalized templates we are able to make sure that these two templates stay in sync. Error states where only one part of the template is applied but not the other would not occur. Generalized templates allows us to subsume the functionality described in [#7](https://github.com/cedar-policy/rfcs/pull/7) and with typed slots we can for the most part express the motivating example for template-groups.
```
chaluli, line 144: I think this sentence is an error: "By using generalized templates we are able to make sure that these two templates stay in sync." I think I get the intuition, though. There is only one template using generalized templates? Thus you don't have to do any work to ensure the templates stay in sync.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+    resource in ?department2) &&
+    context.date < principal.graduationDate 
+};
+```
+
+## Motivation
+
+This is a follow up on the discussion on [#3](https://github.com/cedar-policy/rfcs/pull/3). 
+
+The main motivation to support generalized templates is to allow users to take advantages of templates: (1) prevention against injection attacks (2) ability to change a group of policies overtime. At the same time however, we still want to balance validation, and analysis on templates.
```
chaluli, line 181: I feel that this sentence needs restructured. It's a bit hard to follow.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+
+The main motivation to support generalized templates is to allow users to take advantages of templates: (1) prevention against injection attacks (2) ability to change a group of policies overtime. At the same time however, we still want to balance validation, and analysis on templates. 
+
+In it's previous form, generalized templates were proposed to allow ```?principal``` and ```?resource``` in the condition if it was in the scope and disallowing arbitrary slots. However, this addition would not be able to capture the use cases of examples #2, #3, and #4. 
+
+In this proposal, by having the Cedar user explicitly supply types for slots that are not ```?principal``` and ```?resource``` we are able to express all of the examples listed above while still supporting analysis on templates and allowing for link time validation to be more performant (we can just check that the supplied values are inhabitants of the types of the slots rather than running the type checker on the instantiated template when the slots are filled in). Types are also a form of documentation that can help with reading templates in the future.
```
chaluli, line 185: I think also, it's good to point out it still supports validation as well. And can be supported for non-validated policies, too. So win-win.

text/0098-generalized-templates.md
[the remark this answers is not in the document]
alanwang67: Clarified it in the RFC, type annotations are optional for slots when not using a schema. In Cedar, if we don't use a schema there are no guarantees about the runtime behavior of your policies. However, there are no immediate downsides that I see to optionally allow link time type checking of slots.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+#### Cedar Template 
+```
+permit(
+  principal == ?principal,
+  action,
+  resource in ?resource)
+when {
+    ?resource.owner == resource.owner
```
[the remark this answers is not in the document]
alanwang67: The only case we would miss is when ```in``` is used in the scope, since if we used ```==``` then the slot and the variable would be equivalent. Though, if we wanted to eliminate ```?resource``` we can potentially express it using typed slots as follows if we know the exact type of ```?resource```.
```
template(?bound1: FS::Folder)
permit(
  principal == ?principal,
  action,
  resource)
when {
    resource in ?bound1 && ?bound1.owner == resource.owner 
};
```

alanwang67:
> Might look better with the ?'s removed, as they are unnecessary:

They are mainly kept their for uniformity and as a identifier that they are slots, without ?'s you can potentially have conflicts with the naming. For example, let's say you have an entity type in the global namespace as slot. If you then use this within your policy with the ```is``` operator, it is unclear which is which without the ?. 

> It seems like if ?principal or ?resource were given type annotations, then they shouldn't be required to appear in the scope.

```?principal``` and ```?resource``` are a special type of slot. They are must be in the scope whether or not they are given type annotations. 

> I'm not understanding why we wouldn't want to be able to put the custom slots into the scope, can you explain that more?

As for user defined slots, I'm not exactly sure what types of use cases would motivate having them also appear in the scope. There are also some reasons related to policy indexing behind why ```?principal``` and ```?resource``` are the only slots that can appear in the scope.

**Comment by philhassey** [comment]
> ?principal and ?resource are a special type of slot. They are must be in the scope whether or not they are given type annotations.

I think if they were given type information, maybe they wouldn't have to be special?  Without type info, I understand they need to be in the scope to be analyzed.

> As for user defined slots, I'm not exactly sure what types of use cases would motivate having them also appear in the scope.

Here's an example

```
permit ( principal in ?role, action, resource in ?group );
```

I suppose these could be just ?principal and ?resource, but it seems convenient to be able to name them whatever I want.

**Comment by alanwang67** [comment]
> I think if they were given type information, maybe they wouldn't have to be special? Without type info, I understand they need to be in the scope to be analyzed.

I think if they didn't need to appear in the scope then it would fall into the use case of user defined slots except you wouldn't be able to name them ```?principal``` and ```?resource```.

**Review by chaluli — APPROVED** [review]

**Review by cdisselkoen — APPROVED** [review]
In favor of this change

**Review by shaobo-he-aws — APPROVED** [review]

**Comment by emina** [comment]
And one minor comment on the proposal:  I don't think we should treat `?principal` and `?resource` variables differently than other template variables.  In my view, it's more confusing to have to explain the corner case of "these two slots cannot appear in conditions" than to just allow them to appear in conditions.  Users who find these variable names too confusing to see in conditions don't have to use them---they would have the option of defining their own.

**Comment by shaobo-he-aws** [comment]
> In particular, the motivating examples for this RFC fall roughly into three categories:

Good point. I agree that this RFC needs more convincing examples on discretionary policies.

 > The cost of implementing this RFC is not limited to the cost of changing Cedar, which is already non-trivial. This RFC will also impose a high implementation burden on any policy management service that wants to stay up to date with new Cedar versions. Specifically, services that support the current template model assume that `?principal` and `?resource` are the only template variables, which means that template instantiation records are simple structures with a small fixed number of slots to fill, and the way the slots are filled is also fixed (with just entity values). A service that is built under these assumptions will need to migrate potentially thousands of customers and millions of policies to a new storage model for records---or complicate the service code to allow some in-between state where they support both old and new instantiation records. This is an invasive change that requires a lot of work, and in my view, the benefits of this RFC aren't high enough to offset the cost.
 
 I don't think there will be a huge burden. My naive understanding of the workflow of templates is that a frontend sends a request to a backend running Cedar that instantiates a template and adds it to the policy set. The request should contain the bindings (from template variables to values), presumably in JSON. And the whole process should not rely on any persistent storage. Furthermore, the new set of bindings should be a superset of existing (`?principal` or `?resource`) bindings, which means backward compatibility shouldn't be an issue as well.

**Comment by apandays** [comment]
I think the reason for supporting this feature is primarily around discretionary permissions. 

A good example is a customer using Cedar policies to manage access to a data warehouse based on user groups and attributes. They create a policy for each user group that grant access to a subset of rows based on resource attributes 

For example, a policy might look like 

```
permit (
        principal in UserGroup::"EU accountants", 
        action, 
        resource 
) when {

          principal.location = "EU" and 
          resource.dept = "finance" and 
          ["invoice", "tickets"].contains resource.type and  
          ......<other conditions>
}

The customers have hundreds of such policies in a policy store, and uses partial evaluation to generate a SQL query to get only authorized rows.

cdisselkoen:
We (maintainers) don't have a formal FCP process anymore for RFCs, but we'd still like to do an informal "last call for comments" on this one. Note this one already has an in-progress implementation on the Rust side (cedar-policy/cedar#1732 and cedar-policy/cedar#1738). There's been quite a bit of engagement on this one, and there have also been a few updates since the last comments. We are looking for comments on the latest version.

In particular, for those who have previously been against (@emina @blubben, I think?), does the latest version satisfy your concerns, or not? For those who have previously been in favor, do you have any responses to the last few comments above? Looking particularly at the high-level discussion that happened July 9-24

philhassey (changes requested):
I think it would be nice to remove all the arbitrary constraints surrounding `?principal` and `?resource` and allow people to name and use whatever variables they want.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,277 @@
...
+    resource,
+) when {
+    1 + ?bound1
+};
+```
+
+In addition, there is some subtlety involved when handling type declarations involving primitive types. The main issue stems from entity types being allowed to shadow primitive type. Without a schema to distinguish whether or not this is the case, we opt to be more permissive. We interpret primitive types as both entity types and Cedar primitive types and typechecking passes if one of the cases typechecks. 
+
+### Syntax 
+The proposed syntax looks as follows ```template(?bound1: type1, ?bound2: type2) =>``` and types specified must be valid schema types. Schema types are chosen over AST types because they allow us to more precisely specify the type for sets and records. Each new typed slot requires a distinct variable name. However, within the ```when/unless``` clause the same typed slot can be used multiple times. The grammar of template declarations is as follows. ```Type``` is taken directly from the Cedar documentation for schema types. The template syntax (```template(?bound1: type1, ?bound2: type2) =>```) must go after all the declarations.
```
patjakdev: Why is the `=>` required? Can't we just assume that a policy statement follows the `template(...)` statement?

chaluli (commented):
I support this RFC! In general, I'm in agreement with @philhassey in that I think it would be a nicer experience if `?principa.`, `?resource`, `?action`, and `?context` did not have any special treatment / are reserved.

I understand its in part to be backwards compatible and be opinionated about name choices (to potentially avoid confusion, but I feel like the ? in front of the slot names would already be distinguishing). As a compromise, we could issue a warning during template parsing if they are using one of the reserved keywords in an unexpected way (e.g., `?principal` within the resource scope, etc.).

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,277 @@
...
+
+action AccessDocument appliesTo {
+    principal: FS::Person,
+    resource: FS::Document,
+};
+```
+
+#### Template
+```
+permit(principal, action == Action::"AccessDocument", resource in ?resource) when {
```
luxas: As others pointed out above, it feels like it'd be clearer to force the `template` syntax everywhere a template is used (even though there would not be other arguments), it'd make it clearer than this policy just being an "implicit" template by just using the special `?resource` and `?principal`. I understand why they are special, but I'm thinking if there's a simple way to make it more explicit.

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,277 @@
...
+permit(principal, action == Action::"AccessDocument", resource in ?resource) when { 
+    ?resource in principal.owned_folders
+};
+```
+
+### Example 3
+
+Suppose there are certain professors with dual appointments under two different departments who want to give students of their group access to both the resources in the first department and the resources in their second department. 
+
+Note: Although we can create two seperate templates and instantiate them individually, we would need to make sure they stay in sync. With generalized templates, since there is only one template we do not need to do so. Error states where only one part of the template is applied but not the other would not be able to occur. Generalized templates allows us to subsume the functionality described in [#7](https://github.com/cedar-policy/rfcs/pull/7) and with typed slots we can express the motivating example for template-groups.
```
luxas:
```suggestion
Note: Although we can create two separate templates and instantiate them individually, we would need to make sure they stay in sync. With generalized templates, since there is only one template we do not need to do so. Error states where only one part of the template is applied but not the other would not be able to occur. Generalized templates allows us to subsume the functionality described in [#7](https://github.com/cedar-policy/rfcs/pull/7) and with typed slots we can express the motivating example for template-groups.
```

text/0098-generalized-templates.md
```diff
@@ -0,0 +1,277 @@
...
+
+Aside from the optional type annotation for ```?principal``` and ```?resource``` slots, there are some other differences between them and user defined slots. For one, ```?principal``` and ```?resource``` slots must appear in the scope of the template and can optionally be in the condition of the template, however user defined slots can only appear in the condition clause. Because of this difference ```?principal``` and ```?resource``` slots can only be filled in with entity types and user defined slots can be filled in with any ```SchemaType```.
+
+### What is allowed and not allowed?
+1. Every user defined slot needs to have a type associated with it 
+2. Disallow ```?action``` and ```?context``` slots
+3. Annotated ```?principal``` and ```?resource``` types must be an entity type 
+4. All slots within the template declaration (```template()```) must be used in the template
```
luxas: Given this, it'd be nice to somehow "allowlist" the use of `?resource` and `?principal` too somehow.

luxas (commented):
Great proposal! 🎉 

Left a couple of comments, mostly curious about some of the rationale, and how much further this could be generalized. But I'd be happy to see this in one form or the other 😄

cdisselkoen:
Keeping context on this thread: the state of this RFC is that the RFC needs updates to adopt @philhassey & others' suggestion that `?principal` and `?resource` have no special treatment (the suggestion was also endorsed by @blubben above and agreed to at a Cedar maintainers meeting). I'll be moving off of the Cedar team soon and won't have the bandwidth to make this update myself, so I'm leaving this comment in order to preserve the state of discussion.

PR #99, opened 2025-06-26
Make isInRange Operator Variadic for ipaddr Type

tpaulus:
[Rendered](https://github.com/tpaulus/cedar-rfcs/blob/main/text/0099-ip-is-in-range-list.md)

### DIFF
--- a/text/0099-ip-is-in-range-list.md
+++ b/text/0099-ip-is-in-range-list.md
@@ -113,6 +113,9 @@ Cedar policies are designed to be easy to read and understand by both technical 
 
 ## Detailed design
 
+If any of the arguments passed to `.isInRange` are not of the `ipaddr` type, a type error should be returned and evaluation of this policy be skipped. This is similar to the behavior of the in operator where the evaluator always throws a type error if there exists an ill-typed element even if certain previous element evaluators to true.
+
+
 This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with Cedar to understand, and for somebody familiar with the implementation to implement. This should get into specifics and corner-cases, and include examples of how the feature is used. Any new terminology should be defined here.
 
 // TODO
--- a/text/0099-ip-is-in-range-list.md
+++ b/text/0099-ip-is-in-range-list.md
@@ -1,4 +1,4 @@
-# isInRangeList Operator for ipaddr Type
+# Make isInRange Operator Variadic for ipaddr Type
 
 ## Related issues and PRs
 
@@ -13,7 +13,7 @@
 
 ## Summary
 
-This RFC proposes the addition of a new method-style operator, `.isInRangeList`, to Cedar's `ipaddr` extension type. This operator will take a single `ipaddr` value (the receiver) and check for its inclusion within a `$Set<ipaddr>$` (the argument). It evaluates to `true` if the receiver IP address or range is contained within any of the IP address ranges provided in the argument set, and `false` otherwise. This feature is designed to dramatically simplify policies that must verify a request's source IP against an allowlist of allowed CIDR blocks, improving policy readability, scalability, and maintainability.
+This RFC proposes modifying the existing `.isInRange` operator to Cedar's `ipaddr` extension type. The operator currently is unary, taking a single IP Address; but expanding it to be variadic will enable simplification of policies which check for containment in more than one IP Address block. This maintains backwards compatibility, while simplifying policies that must verify a request's source IP against an allowlist of allowed CIDR blocks, improving policy readability, scalability, and maintainability.
 
 ## Basic example
 
@@ -27,14 +27,14 @@ permit (
   resource
 )
 when {
-  ip(context.network.source_ip).isInRange(ip("198.51.100.0/24")) ||
-  ip(context.network.source_ip).isInRange(ip("203.0.113.0/24")) ||
-  ip(context.network.source_ip).isInRange(ip("192.0.2.0/25")) ||
-  ip(context.network.source_ip).isInRange(ip("2001:db8:a001::/48"))
+  context.network.source_ip.isInRange(ip("198.51.100.0/24")) ||
+  context.network.source_ip.isInRange(ip("203.0.113.0/24")) ||
+  context.network.source_ip.isInRange(ip("192.0.2.0/25")) ||
+  context.network.source_ip.isInRange(ip("2001:db8:a001::/48"))
 };
 ```
 
-The proposed `.isInRangeList` operator simplifies this logic into a single, intuitive function call, making the policy's intent immediately clear.
+The proposed change simplifies this logic into a single function call making the policy's intent clearer.
 
 ```
 // Policy to allow access only from corporate IP ranges (proposed method)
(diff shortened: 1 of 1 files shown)


text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,95 @@
+# isInRangeList Operator for ipaddr Type
```
shaobo-he-aws, line 1: An alternative name would be `isInRanges`.

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Timeline
+
+- Started: 2025-06-26
+- Accepted: TBD
+- Stabilized: TBD
+
+## Summary
+
+This RFC proposes the addition of a new method-style operator, `.isInRangeList`, to Cedar's `ipaddr` extension type. This operator will take a single `ipaddr` value (the receiver) and check for its inclusion within a `$Set<ipaddr>$` (the argument). It evaluates to `true` if the receiver IP address or range is contained within any of the IP address ranges provided in the argument set, and `false` otherwise. This feature is designed to dramatically simplify policies that must verify a request's source IP against an allowlist of allowed CIDR blocks, improving policy readability, scalability, and maintainability.
```
shaobo-he-aws, line 16: We should specify the case where a list element is not an ip address. I assume that it should lead to a type error. Besides, the behavior should be akin to that of the `in` operator where the evaluator always throws a type error if there exists an ill-typed element even if certain previous element evaluators to true.

shaobo-he-aws (commented):
I'm in favor of this extension.

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,95 @@
...
+  ip(context.network.source_ip).isInRangeList(
+    [
+      ip("198.51.100.0/24"),
+      ip("203.0.113.0/24"),
+      ip("192.0.2.0/25"),
+      ip("2001:db8:a001::/48")
+    ]
+  )
+};
+```
```
emina, line 56: Augh, we actually can't support this method as proposed:  it's not analyzable. 

To see why, consider the expression `ip(context.network.source_ip).isInRangeList(context.ranges)`. 

In this case, the set of ranges, `context.ranges`, is not a set literal, and it is therefore unbounded.  That is, we don't know its contents or size by just looking at the policy, so we'd have to encode a loop to check if this holds.  Such loops aren't analyzable:  https://github.com/cedar-policy/rfcs/blob/main/archive/rfc/0021-any-and-all-operators.md

But there are two alternatives that would be analyzable:

1) Have the validator reject policies in which the argument to `isInRangeList` is not a set literal.  There is a precedent for this:  the validator rejects calls to extension type constructors where the arguments aren't literals.

2) Instead of introducing a new method, overload the `in` operator and, again, have the validator reject policies if the LHS of `in` is an ip address and the RHS is not either an ip address or a set literal containing ip addresses.

So, this example would be:

```
permit (
  principal,
  action == Action::"accessAdminDashboard",
  resource
)
when {
  ip(context.network.source_ip) in 
    [
      ip("198.51.100.0/24"),
      ip("203.0.113.0/24"),
      ip("192.0.2.0/25"),
      ip("2001:db8:a001::/48")
    ]
};
```

We could also support writing the following (in addition to the existing `isInRange` method, which we have to keep for backwards compatibility):

```
ip(context.network.source_ip) in ip("198.51.100.0/24")
```

Both of the above overrides are unambiguous at the type level, though they would add some extra branching to the evaluation of the `in` operator.

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,95 @@
...
+
+```
+// Policy to allow access only from corporate IP ranges (current method)
+permit (
+  principal,
+  action == Action::"accessAdminDashboard",
+  resource
+)
+when {
+  ip(context.network.source_ip).isInRange(ip("198.51.100.0/24")) ||
```
emina, line 30: Note that this example would be rejected by the validator:  the argument to `ip` has to be a string literal.

This would work though, assuming that `context.network.source_ip` is of type IPAddr:

```
permit (
  principal,
  action == Action::"accessAdminDashboard",
  resource
)
when {
  ip(context.network.source_ip).isInRange(ip("198.51.100.0/24")) ||
Comment
Note that this example would be rejected by the validator:  the argument to `ip` has to be a string literal.

This would work though, assuming that `context.network.source_ip` is of type IPAddr: `context.network.soruce_ip.isInRange(ip("198.51.100.0/24")) || ...`.

File: text/0099-ip-is-in-range-list.md [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+
+```
+// Policy to allow access only from corporate IP ranges (proposed method)
+permit (
+  principal,
+  action == Action::"accessAdminDashboard",
+  resource
+)
+when {
+  ip(context.network.source_ip).isInRangeList(
```
emina, line 47: Same as above:  can we change the example `context.network.source_ip.isInRangeList(...)` so it validates?

text/0099-ip-is-in-range-list.md
tpaulus: I like that too! I've updated the RFC with option 4.

text/0099-ip-is-in-range-list.md
tpaulus: Fixed

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,135 @@
...
+* __Increased Language Surface Area:__ Any new feature adds to the complexity of the language. However, the proposed operator's logic is a minimal change, making it an intuitive and easily understood addition rather than a complex new concept.
+
+* __Implementation Effort:__ This is a substantial feature that requires coordinated development effort across the cedar core repository (parser, validator, evaluator) and the cedar-spec repository for the formal model. This is a standard and expected cost for any significant language enhancement, similar to the effort required for features like the is operator.   
+
+## Alternatives
+
+### The Status Quo (Chained `||`)
+This is the current method for checking against multiple IP ranges. As detailed extensively in the Motivation section, this approach is not scalable and suffers from poor readability and maintainability, especially as the number of IP ranges grows. It is the problem this RFC aims to solve.
+
+## Unresolved questions
```
emina, line 134: I think we can consider this resolved :)

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,135 @@
...
+## Motivation
+Making the `.isInRange` operator variadic is motivated by several key factors that align with Cedar's goals of expressiveness and readability.
+
+### Smaller JSON Formatted Policies
+The current method of chaining `.isInRange()` calls with `||` operators creates a large and object tree that can quickly exceeded the stack depth and recursion limit of systems. The use of a variadic operator would allow for the reduction in not only the policy objet's size.
+
+### Improved Policy Readability and Authoring
+Cedar policies are designed to be easy to read and understand by both technical and non-technical stakeholders. Chaining `.isInRange()` calls with `||` operators creates syntactically complex expressions that obscure the simple intent of "is this IP in our allowlist?". As the list of IP ranges grows, the policy becomes increasingly unwieldy, increasing cognitive load and the likelihood of authoring errors, such as typos or misplaced logical operators. The proposed operator replaces a complex boolean expression with a single, declarative function call, significantly improving the clarity and conciseness of the policy.
+
+## Detailed design
```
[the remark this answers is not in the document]
emina: Alternatively, we can require `.isInRange` to have at least one argument:

Implementing this RFC requires extending the parser, evaluator, typechecker, and symbolic compiler to allow .isInRange to take one or more arguments. Since .isInRange is an extension function, there is no need to change the AST, since the AST nodes for extension functions already take as input a list of arguments.

The evaluator should be modified so that it no longer forces .isInRange to have exactly one argument. Instead, the evaluator checks that at least one argument is provided and that all provided arguments are of type ipaddr. If not, the evaluation errors. Otherwise, the evaluator checks if the receiver IP address is in the range of any argument IP addresses.

The typechecker and symbolic compiler should be modified similarly to the evaluator.

emina (approved):

shaobo-he-aws (approved):
I'm favor of this RFC although being concerned about the implementation burden.

shaobo-he-aws:
> @shaobo-he-aws Can you say more about the implementation burden? Are you worried about the parser / CST -> AST conversion on the Rust side?

Yes, the changes on the Rust side will be non-trivial compared to the scope of the RFC. But the effects are likely justified if we plan to have more variadic extension functions in the future.

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+## Motivation
+Making the `.isInRange` operator variadic is motivated by several key factors that align with Cedar's goals of expressiveness and readability.
+
+### Smaller JSON Formatted Policies
+The current method of chaining `.isInRange()` calls with `||` operators creates a large and object tree that can quickly exceeded the stack depth and recursion limit of systems. The use of a variadic operator would allow for the reduction in not only the policy objet's size.
+
+### Improved Policy Readability and Authoring
+Cedar policies are designed to be easy to read and understand by both technical and non-technical stakeholders. Chaining `.isInRange()` calls with `||` operators creates syntactically complex expressions that obscure the simple intent of "is this IP in our allowlist?". As the list of IP ranges grows, the policy becomes increasingly unwieldy, increasing cognitive load and the likelihood of authoring errors, such as typos or misplaced logical operators. The proposed operator replaces a complex boolean expression with a single, declarative function call, significantly improving the clarity and conciseness of the policy.
```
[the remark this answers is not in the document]
chaluli: Doesn't this depend on how `isInRange` semantics are defined?

If you define it so that it evaluates from left to right and returns once any argument would evaluate to true it retains the same semantics as using disjunction, right?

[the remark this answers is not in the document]
cdisselkoen: You could define it that way, but that would be unlike how any other function calls work (and how I think folks expect function calls to work in general), which is that all arguments are eagerly evaluated before passing them to the function

adpaco-aws (approved):

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,134 @@
...
+
+The evaluator should be modified so that it no longer forces `.isInRange` to have exactly one argument. Instead, the evaluator checks that at least one argument is provided and that all provided arguments are of type ipaddr. If not, the evaluation errors. Otherwise, the evaluator checks if the receiver IP address is in the range of any argument IP addresses.
+
+The typechecker and symbolic compiler should be modified similarly to the evaluator.
+
+## Drawbacks
+
+* __Increased Language Surface Area:__ Any new feature adds to the complexity of the language. However, the proposed operator's logic is a minimal change, making it an intuitive and easily understood addition rather than a complex new concept.
+
+* __Implementation Effort:__ This is a substantial feature that requires coordinated development effort across the cedar core repository (parser, validator, evaluator) and the cedar-spec repository for the formal model. This is a standard and expected cost for any significant language enhancement, similar to the effort required for features like the is operator.
```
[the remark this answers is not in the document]
tpaulus: Updated in [efcc2d6](https://github.com/cedar-policy/rfcs/pull/99/commits/efcc2d6da59093587f6662515a2efc45fc3e0fc8)

text/0099-ip-is-in-range-list.md
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+]
+```
+
+## Motivation
+Making the `.isInRange` operator variadic is motivated by several key factors that align with Cedar's goals of expressiveness and readability.
+
+### Smaller JSON Formatted Policies
+The current method of chaining `.isInRange()` calls with `||` operators creates a large and object tree that can quickly exceeded the stack depth and recursion limit of systems. The use of a variadic operator would allow for the reduction in not only the policy objet's size.
```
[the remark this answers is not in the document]
tpaulus: Addressed in [efcc2d6](https://github.com/cedar-policy/rfcs/pull/99/commits/efcc2d6da59093587f6662515a2efc45fc3e0fc8).

text/0099-ip-is-in-range-list.md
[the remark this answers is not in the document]
tpaulus: @cdisselkoen's note about eager evaluation has been added in [efcc2d6](https://github.com/cedar-policy/rfcs/pull/99/commits/efcc2d6da59093587f6662515a2efc45fc3e0fc8).

PR #101, opened 2025-08-06
Proposal for policy store specification

ossdhaval:
[Rendered](https://github.com/ossdhaval/cedar-rfcs/blob/main/text/0101-policy-store-specification.md)

### DIFF
--- a/text/0101-policy-store-specification.md
+++ b/text/0101-policy-store-specification.md
@@ -22,26 +22,19 @@ Create a specification that recommends a JSON descriptor file to store Cedar pol
 
 ## Motivation
 
-
-_Why are we doing this? What use cases does it support? What is the expected outcome?_
-
-_Please focus on explaining the motivation so that if this RFC is not accepted, the motivation could be used to develop alternative solutions. In other words, enumerate the constraints you are trying to solve without coupling them too closely to the solution you have in mind._
-
-------
-
 The proposal in this RFC targets the following constraint:
 
 ### Lack of a recommended method to co-locate and bind policies, schema, and other information
 
 Cedar does not prescribe any method or system that co-locates and binds together the information about the schema, and the policies that use that schema. In Cedar, a policy's binding with its schema happens at the time of evaluation. This approach gives flexibility to combine any schema with any policy at evaluation time. But in practice, more often than not, the policies are targeted to a specific schema, which they use every time. 
 
-In light of this fact, not having a prescribed method of co-location and binding the policy, schema and other information has the following drawbacks:
+The following drawbacks stem from not having a prescribed method of co-location and binding the policy, schema, and other information:
  
-* Versioning: in the audit logs to create a cryptographic chain of custody, it's critical to know the exact version of the policies at evaluation time. Formally binding the policies and schema makes the audit logs more clear
-* Productivity: Developers will have to separately map all the assets individually. Federated identity protocols like SAML and OpenID typically bundle entity metadata, which simplifies both development and operations
-* Risk: The lack of a formal binding increases the risk of error, e.g., using the wrong schema or schema version
+* **Versioning**: in the audit logs to create a cryptographic chain of custody, it's critical to know the exact version of the policies at evaluation time. Formally binding the policies and schema makes the audit logs more clear
+* **Productivity**: Developers will have to separately map all the assets individually. Federated identity protocols like SAML and OpenID typically bundle entity metadata, which simplifies both development and operations
+* **Risk**: The lack of a formal binding increases the risk of error, e.g., using the wrong schema or schema version
 
-This lack of a prescribed method for co-locating and binding the policy, schema and other information together, creates more pain points for large organizations. Large organizations invariably see a multitude of policies and corresponding schemas. To maintain order, organizations will establish their own conventions and structures for storage policies and schemas. Invariably, this will result in inconsistent storage structures across different organizations or within different teams in the same organization.
+This lack of a prescribed method creates even more pain for large organizations, as they have to deal with a multitude of policies and corresponding schemas. To maintain order, organizations will establish their own conventions and structures for storage policies and schemas. Invariably, this will result in inconsistent storage structures across different organizations or within different teams in the same organization.
 
 Such inconsistency in methods used by organizations will create the following systemic issues in the Cedar ecosystem:
 
@@ -53,7 +46,7 @@ Such inconsistency in methods used by organizations will create the following sy
 
 This RFC proposes to create a new "policy store" specification. The specification will recommend a file descriptor format that contains all information required to evaluate an authz request. 
 
-Sample policy-store-descriptor.json structure--only an illustrative hypothetical sample--check below [design](detailed-design) for the details.
+See the sample content of a descriptor file below. This is only an illustrative example. Check [design](detailed-design) section for more details.
 
 ```
 {
@@ -74,7 +67,7 @@ Sample policy-store-descriptor.json structure--only an illustrative hypothetical
 ## Benefits
 
 ### Easier for End Users
-Having one one file versus three lowers the transaction costs for end-users of Cedar. As working with policies and schema is quite common, the cumulative productivity benefit of a small optimizations can add up! 
+Having one file versus three(policies, schema, etc) lowers the transaction costs for end-users of Cedar. As working with policies and schema is quite common, the cumulative productivity benefit of a small optimizations can add up! 
 
 ### Strong versioning
 
@@ -96,9 +89,11 @@ A robust Cedar third party developer ecosystem will help drive broader adoption 
 
 ## Detailed design
 
-* [Field descriptions](#field-descriptions) below define each field in the JSON structure. 
-* The [JSON schema](#json-schema-for-the-policy-store) provides an accurate structure of the policy store. 
-* Refer to the policy store [examples](#policy-store-examples) to help visualize the policy store in real-life usage.
+Content structure of the policy store descriptor file is detailed in three sections below.
+
+* [Field descriptions](#field-descriptions) section describes each field in the policy store's JSON structure. 
+* The [JSON schema](#json-schema-for-the-policy-store) section provides an accurate structure of the policy store's JSON structure. 
+* Refer to the policy store file [examples](#policy-store-examples) to help visualize the policy store in real-life usage.
 
 ### Field Descriptions
 
@@ -125,7 +120,7 @@ The structure below loosely follows policy store JSON structure, minus some stru
 
 ##### Trusted issuers
 
-The `trusted_issuers` field is OPTIONAL. It contains information about a domain that mints JWT tokens. A store can have many trusted issuers, each who issue different kinds of tokens. Specifying the trusted issuers enables Cedar-based PDPs to establish the sources of truth who can supply evidence to satisfy policies. The signed tokens from trusted issuers are essential to establish a cryptographic chain of custody that compliments policy store versioning, proven delivery of the policy store, and secure deliver of decision logs. 
+OPTIONAL. It contains information about a domain that mints JWT tokens. A store can have many trusted issuers, each who issue different kinds of tokens. Specifying the trusted issuers enables Cedar-based PDPs to establish the sources of truth who can supply evidence to satisfy policies. The signed tokens from trusted issuers are essential to establish a cryptographic chain of custody that compliments policy store versioning, proven delivery of the policy store, and secure deliver of decision logs. 
 
 ###### Claims to schema mapping
 
--- a/text/0101-policy-store-specification.md
+++ b/text/0101-policy-store-specification.md
@@ -42,9 +42,9 @@ This lack of a prescribed method creates even more pain for large organizations,
 
 Such inconsistency in methods used by organizations will create the following systemic issues in the Cedar ecosystem:
 
+- Reduces PDP interoperability, for example makes it harder to move a policy store from Janssen Project Cedarling to Amazon Verified Permissions and vice versa. 
 - Impedes the development and deployment of third-party tools that need to access policies and schema. For instance, third-party policy analysis tools, visualization tools, or workflow integration tools like n8n. A uniform method of storing information across teams and organizations will reduce the configuration and customization burden on third-party tools.  
-- Adds overhead to achieve interoperability. For instance, a merger between two companies that use different storage methods. Tools and apps written by one organization will not be able to be deployed without the additional effort to unify policy and schema storage methods.
-- Makes it difficult for policy administrators who work with different teams of the same organisation if the teams have their policies and schema stored differently. 
+- Reduces portability, for instance, a merger between two companies that use different storage methods. Tools and apps written by one organization will not be able to be deployed without the additional effort to unify policy and schema storage methods. Also makes it difficult for policy administrators who work with different teams of the same organisation if the teams have their policies and schema stored differently. 
 
 ### Proposed solution
 
@@ -149,6 +149,7 @@ policy-store-root/
 - **Issuer files**: Must have `.json` extension and contain trusted issuer configurations
 - **Schema file**: Must be named `schema.cedarschema` and contain the Cedar schema
 - **Metadata file**: Must be named `metadata.json` and contain policy store metadata
+- **Manifest file**: Must be named `manifest.json` and contain data about files in the policy store
 - **Archive file**: Must have `.cjar` suffix
 
 #### File Content Requirements
@@ -157,7 +158,8 @@ policy-store-root/
 - **Template files**: Each file must contain exactly one Cedar template with an `@id()` annotation  
 - **Entity files**: Each file contains a JSON array of entity definitions or a single entity definition
 - **Schema file**: Contains the Cedar schema in `.cedarschema` format
-- **Metadata file**: Contains policy store metadata in JSON format (see Metadata Format section) 
+- **Metadata file**: Contains policy store metadata in JSON format (see Metadata Format section)
+- **Manifest file**: Contains an inventory of all files in the policy store for validation and integrity checking 
 
 ### Metadata Format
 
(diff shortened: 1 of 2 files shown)


text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,428 @@
...
+
+Create a specification that recommends a JSON descriptor file to store Cedar policies, schemas, default entities, and other relevant information.
+
+## Motivation
+
+The proposal in this RFC targets the following constraint:
+
+### Lack of a recommended method to co-locate and bind policies, schema, and other information
+
+Cedar does not prescribe any method or system that co-locates and binds together the information about the schema, and the policies that use that schema. In Cedar, a policy's binding with its schema happens at the time of evaluation. This approach gives flexibility to combine any schema with any policy at evaluation time. But in practice, more often than not, the policies are targeted to a specific schema, which they use every time.
```
emina, line 29: Note that policy evaluation is independent of schemas in Cedar.  That is, you don't need any schema at all to evaluate policies.  You only need schemas to validate or analyze policies---or, if you are (optionally) using schema-based entity parsing, to parse entities.

So, I think we need to rephrase the motivation here to clarify this:  there are no required dependencies on schemas at authorization (run) time.

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,428 @@
...
+Why should we *not* do this? Please consider:
+
+There are no apparent drawbacks to creating this specification. Primarily owing to the facts below:
+
+- Implementation cost: None, as this specification does not require any code changes in the Cedar language
+- Migration of applications using Cedar: Apps using Cedar can choose not to adopt this specification. It is voluntary. If they choose to adopt, the apps can take a gradual path to adopt this specification. Release of this specification does not force compliance or timelines on applications using Cedar. 
+
+## Alternatives
+
+Not applicable
```
emina, line 418: The alternative is to maintain status quo and not provide a policy store specification as part of Cedar.

text/0101-policy-store-specification.md
nynymike: Yes, we are doing "schema-based entity parsing" in the Cedarling based on the rules for a trusted issuer, as mentioned in the docs [here](https://docs.jans.io/stable/cedarling/cedarling-policy-store/#claim-mapping).

text/0101-policy-store-specification.md
patjakdev: My feedback is that I don't really understand why the community needs an authoritative specification for the "policy store" concept. The specification you've created seems fine for your specific use case, but I don't really see a need for the Cedar team to dictate (and maintain into the future) such a specification.

Are you looking to interoperate with policy stores generated by third parties or something?

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,428 @@
...
+    - `trusted_issuers`: A mapping of trusted issuer IDs to their metadata.
+      - `<issuer_id>`: unique hash to identify a trusted issuer
+        - `name`: Name of the trusted issuer.
+        - `description`: A human-readable description of the issuer
+        - `openid_configuration_endpoint`: OpenID Connect discovery endpoint URL for this issuer.
+        - More fields can be added to describe the trusted issuer and token metadata
+    - `schema`: base64-encoded Cedar schema used by policies in this store. There can be only one schema per policy store.
+    - `default_entities`: A mapping of entity IDs to base64-encoded JSON entity data. 
+
+##### Trusted issuers
```
RoyalCookie, line 121: I think attempting to defined all these new concepts in the initial RFC might be a mistake, I'd expect some pushback to simply introducing the Policy Store concept to Cedar.

But I think every AVP/OIDC/OAuth specific attribute outlined in the RFC  requires it's own discussion so it might be more productive to simplify the scope of this RFC.

Limit it specifically to introducing the `Policy Store` concept where only the core attributes are defined such as `description`, `policy_content`, `creation_date?`, but allows for additional user defined  attributes perhaps?

text/0101-policy-store-specification.md
cdisselkoen: You can pass them in `entities` if you do not pass a schema.  (And if you don't have action groups, I believe you don't need to pass them in either `entities` or schema)

text/0101-policy-store-specification.md
Swolebrain: I completely agree we should leave out identity-related portions of the specification for the time being.

There's a tremendous amount of complexity in coming up with a general-enough abstraction to support a majority of oidc providers, due to the fact that they all deviate from the oidc spec in slightly different ways. For example:

* The concept of "groups" is modeled differently. Okta models it as a "Groups" claim that can only be present in identity tokens. Ping models it as a "groups" claim that is present in both identity and access tokens. Cognito models it as a "cognito:groups" claim that can be present in either token. Auth0 models something similar as a "permissions" claim. 
* Different levels of support for JWE token encryption
* Different treatments for `client_id` claim. The one example I can remember is that Auth0 sometimes uses an `azp` claim (depending on which type of "auth0 application" the token comes from) instead of `client_id` to represent the client id.
* Can't find the details right now but if my memory doesn't fail me, there's also different formats for some standard claim values (where one idp will want a space-separated string another one might want a string array).

Beyond that, this proposal doesn't leave room to support other identity sources, like for example SAML, or any future specs that might come out (perhaps for MCP's or what not).

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,428 @@
...
+    "policy_stores": {
+        "9e96b204911d1c3": {
+            "name": "Acme Analytics Web Application",
+            "description": "Acme's premier tool for counting narwhals, doves and gorns.",
+            "policies": {"a270f688133cd00": "base64-content",...},
+            "trusted_issuers": {...},
+            "schema": "base64-content"
+            "default_entities": {"1694c954f8d9": "base64-content",...}
+        }
+    }
```
Swolebrain, line 63: While I think the idea of standardizing on a policy store format is fantastic, and I thank you for your contribution, I have several objections to this format. I will try to organize them by theme below.

I will use the term **builder** to refer to people like us, who build applications using Cedar, and use the term **end user** to refer to the users using those applications.

### 1. Use case support
There are at least two major use cases that are left out, not possible to support by the format outlined in this proposal: multi-tenancy and dynamic policy stores (as well as use cases which have both attributes). This format is good for IaC-driven policy stores but it is not good for dynamic policy stores. 

To summarize, I will use as an example a use case that requires both: dynamic policies and multi-tenancy. Let's say you're a builder working on ClickDown.com, an imaginary productivity app that allows you to manage tasks, create and share docs, manage content, and even live chat with coworkers. You're using Cedar to model the authz model of ClickDown. With Cedar, when building a multitenant app like ClickDown, the best practice is to have one policy store per tenant. This allows for permission isolation and avoids accidental collisions. Since your application is very successful and you market with the freemium model, your app has 80 million users. That makes this JSON file be somewhere north of 2.4 terabytes (assuming 30kb per policy store, which is actually low). 

Additionally, each tenant of ClickDown has many end users. Admin end-users of the ClickDown app have the ability to create other users. Since you user Cedar, there's a lot of flexibility in what you can do in your "Create user" flow. You provide your tenant admins with a flow such that they can select all the actions and resources the new user is allowed to act on. One or more cedar policies are created in the backend. So when you create a new user `ClickDown::User::"123123"`, there are associated policies like `permit(principal == ClickDown::User::"123123", action...);` generated specific to that user. Under this use case (which is common), each tenant will have policy stores that will grow arbitrarily. One tenant by itself might have 100,000 users in it, with 100,000 policies. It's impractical to version the whole policy store when one single policy was created or updated.

### 2. Format serialization

If a policy store descriptor format is to exist, one of its uses should be to serialize dynamic policy stores for the purposes of disaster recovery. Serializing a multi-gigabyte file in json is needlessly difficult. Since you can't load it all into memory (because you run oom) you need to write a streaming serializer. This needlessly increases the complexity of encoding this format.

It's much easier to compare the git `HEAD` to `HEAD~1` and only deploy the files that were changed.

De-serializing the format to compare it to what's already deployed has similar streaming complications.

### 3. Support for templates

If a policy store descriptor format is to exist, it should support policy templates and template linked policies.

### My proposal

I would recommend that instead of this, the policy store format should:

1. Encompass a single policy store
2. Store units in the most atomic manner possible, so they can be versioned independently and thus deployed independently.

One structure could be like so:

```
└── root/
    ├── policyStoremetadata.json                          # Export metadata (cedar version, name, description, etc)
    ├── schema.cedarchema                 # Schema definition for the policy store
    ├── static_policies/                       # Directory for standalone policies
    │   ├── policy_1.cedar                     # Individual Cedar policy files, filename should be policy id
    │   ├── policy_2.cedar
    │   └── policy_n.cedar
    ├── templates/                             # Directory for templates and their linked policies
    │   ├── template_1/                      # Folder should be template id
    │   │   ├── template.cedar         
    │   │   ├── link_1.json                    # Batch of template variable combinations representing template linked policies
    │   │   ├── link_2.json                    
    │   │   └── link_n.json
    │   ├── template_2/
    │   │   ├── template.cedar
    │   │   └── link_1.json
    │   └── template_n/
    │       └── template.cedar
    └── manifest.json   
```

Note there's a `manifest.json` file which can be used to list all the contents of a policy store. It would end up being something reminiscent of an `info.plist` file in an iOS app or a progressive web app's `manifest.json` file. I haven't thought through this in detail, but one pro of having a manifest is that it _should_ be easier to do static analysis validation on the manifest. For example, if the manifest says that the policy store should perform validation at authorization time, then there must be a schema file present, it must be parseable, and all the policies named in the manifest must be validated against the schema.

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,428 @@
...
+## Benefits
+
+### Easier for End Users
+Having one file versus three(policies, schema, etc) lowers the transaction costs for end-users of Cedar. As working with policies and schema is quite common, the cumulative productivity benefit of a small optimizations can add up! 
+
+### Strong versioning
+
+* For forensic analysis of an audit log, you must know the exact version of the policies and schema in use at the time of the decision. 
+* Versioning provides an opportunity for review and governance. For example, the CISO may want to review the policies of a certain application and then review all subsequent changes. Versioning aligns with existing change management and CI/CD pipelines.
+* Strong versioning aligns with cloud native declartive syntax goals, making the deployment of new code and rollback to old code easier.
```
Swolebrain, line 76: Using a single file makes these 3 versioning concerns difficult to support:

* For forensic analysis, under the context of multi-tenancy and dynamic policies, if tenant1 changes one policy, that creates a whole new version of the policyStores file. So tenant 700 sees a different policy store version in audit logs even though they've changed no policies
* If a policy store has a thousand policies and only one policy is changed, now my IAC deployer needs to download the whole policy store, diff it against the file, and change what's necessary to change. That, or put write pressure on the DB re-writing everything. Neither is preferrable over updating only what needs to be updated, which can be accomplished by versioning at a more granular level.
* See above, it's much easier to deploy changes and roll back if you know what specific things changed in each policy store.

text/0101-policy-store-specification.md
Swolebrain: There's one alternative I've outlined above. In general, alternatives are always applicable. For instance with the alternative I outlined, we make it more difficult to cryptographically sign the state of a particular policy store, but we can support more use cases. Another alternative is to specify the cedar schema in a way that's driven by the API spec. It can be versioned and signed in that way. We have a reference implementation for how an OpenAPI spec can be made to express a cedar schema here: https://github.com/cedar-policy/cedar-authorization

Regarding the mapping of issuers, I believe what Phil meant is that we have oidc hardcoded into this spec without a way to specify other types of identity sources, like SAML or any future ones.

text/0101-policy-store-specification.md
nynymike: > And while it is supported by Amazon Verified Permissions, it is limited to one Trusted Issuer at the moment.
> 

So if we could define the spec in a way that supports AVP's one Trusted Issuer, but open the possibility that we'll need to make policies about more then one issuer... that would be a win right?  

Is it really so future looking to say that we may need to write policies about tokens from more then one issuer? I would say we're already seeing these requirements in the AI world.

text/0101-policy-store-specification.md
nynymike: I actually quite like this design... I even like the use of templates... for those who don't want it, don't use it.

text/0101-policy-store-specification.md
nynymike: I don't actually agree that every policy change requires a new policy store version. If you have 700 policies, you edit 50 of them, and then submit a PR for approval, and once approved... it gets released as a new policy store version.

If your app has dynamic policies, then a policy store version won't make sense for you. I think the individual policy version is just as good as long as you have a way to create a chain of custody for the audit log.

We will try to flush out these examples in the revision...

ossdhaval:
Thank you everyone for your time and input. We have revised the RFC to address the comments. Please follow the link in the PR description to read the updated version. 

#### Major changes

- Moving from a single JSON file structure to a developer-friendly directory format. For suggesting this idea, @Swolebrain has been added as a contributor. :-)
- Each policy, template, and entity is stored as an individual, human-readable file. No base64-encoded strings. Greatly improving version control, IDE support, and collaborative workflows
- Design allows for multiple trusted issuers
- For deployment or distribution, the proposal introduces a standardized .cjar archive format that preserves the structured directory
- The RFC details file conventions, extensible metadata, and open questions while explicitly outlining drawbacks and alternatives


And many more changes.

Looking forward to your inputs on the revised RFC.

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,595 @@
...
+
+## Summary
+
+Create a specification that recommends a structured directory format to store Cedar policies, schemas, default entities, and other relevant information. The specification supports both a developer-friendly directory structure for development and editing, and a compressed archive format for distribution and deployment. Each policy and template is stored as an individual human-readable file, enabling better version control, IDE support, and collaborative development workflows.
+
+## Motivation
+
+The proposal in this RFC targets the following constraint:
+
+### Lack of a recommended method to co-locate and bind policies, schema, and other information
```
benkehoe, line 31: I'm struggling with why this makes sense to include as part of Cedar proper.

> policies are targeted to a specific schema

This might be addressable with a much more narrow proposal, like a well-known policy annotation for linking the policy to a schema version.

> In the audit logs to create a cryptographic chain of custody, it's critical to know the exact version of the policies and the schema used at evaluation time. For example, when using schema-based entity parsing.

This is a very use case-specific assumption, that the policies as collected and schema are accessed via a consistent snapshot for authorization decisions, and that schema is used at authorization time. Cedar itself does not define either of these things.

The rest of the motivation deals with interoperability of authorization systems that are built _using_ Cedar, as opposed to the interoperability of Cedar itself. It doesn't make sense to me, for example, that Cedar would define a format that includes trusted issuers, which are not represented anywhere within Cedar itself.

My take here is that a better path might be establishing a separate effort (within or alongside the Cedar project) to define (optional) standards for authorization systems built on top of Cedar. A policy store spec is one aspect. Another could be, for example, a standardized way to provide entities in a call to an AuthZEN-compliant authorization endpoint.

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,595 @@
...
+- **Claim mapping configurations**: Mapping JWT claims to Cedar entity attributes for normalization
+- **Role and permission mappings**: How to extract user roles from token claims
+- **Principal mapping**: Which Cedar entity types this token can represent as a principal
+- **Issuer-specific metadata**: Custom fields needed for integration with proprietary identity systems
+
+This extensibility allows the specification to support diverse identity provider ecosystems while maintaining a common foundation for interoperability.
+
+#### Default Entities
+
+Default entities are stored as individual JSON files in the `entities/` directory. These files capture static resource entities, such as organization information that doesn't frequently change.
```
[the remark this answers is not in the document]
nynymike: Ok, hence the `default_` part of the claim :-)

text/0101-policy-store-specification.md
```diff
@@ -0,0 +1,595 @@
...
+1. **Directory Structure Format**: A folder-based approach where each component (policies, templates, schema, entities) is stored as individual files within a structured directory hierarchy
+2. **Archive Format**: A compressed archive (zip file) with the file extension `.cjar`, containing the same directory structure for easy distribution and deployment
+
+See the sample directory structure below:
+
+```
+policy-store/
+├── metadata.json                    # Policy store metadata
+├── manifest.json                    # Inventory of files
+├── schema.cedarschema               # Cedar schema file
```
blubben, line 64: I'd suggest making the schema a subdirectory of its own. It can be convenient to break a schema up by namespace (and sharing schema fragments across projects.)

ossdhaval:
Hello everyone, 

This RFC has been updated to address the feedback comments. 

**Highlight of the changes:**

- Introduced the new `manifest.json` file that provides an inventory of all files in the policy store for validation and integrity checking
- Added samples to cover action groups
- Expanded details on `Default entities`
- Added JSON Schema for trusted issuers and entity files

Please share your views for the updated RFC.

nynymike:
We made some changes to address the feedback, like simplifying the `trusted_issuer` metadata a little and adding another example to the entity.  I hope we can make progress on this RFC soon.

nynymike:
https://aws.amazon.com/blogs/aws/amazon-bedrock-agentcore-adds-quality-evaluations-and-policy-controls-for-deploying-trusted-ai-agents/
image

Governance needs to be interoperable. While CNCF community contributors like Amazon are developing interesting services, their claims to audit and compliance teams are hard to accomplish without specific versioned policies. We need to know what version of policy is in the execution environment, and how it got there. And we need continuous evaluation according to an assessment procedure, without which, the policy is merely a recomendation. Lack of a versionable binary like this makes Cedar impossible to govern in an interoperable way.

nynymike:
Duly noted. I'm in the process of proposing a "GovOps" working group at OpenSSF, which could be a home for interoperability related standards.

text/0101-policy-store-specification.md
windley: Late to the party here... 

I don't know that we need a specification that binds Cedar to a specific format, but I do see value in having well-thought-out formats for how to package and move schema and policies (which is what I see this being about primarily).

nynymike:
I'm proposing this as an AuthZen Standard. See draft [here](https://htmlpreview.github.io/?https://github.com/nynymike/AuthZen_Policy_Store/blob/main/draft-schwartz-authzen-policy-store.html)

benkehoe:
I think AuthZen makes a ton of sense for this.

ossdhaval:
We are closing this RFC in light of the proposed AuthZen Standard. 

Thank you everyone for chipping in. Your inputs have helped us shape the original idea. :pray:

PR #104, opened 2025-08-20
Cedar syntax for entity data

benkehoe:
[Rendered](https://github.com/benkehoe/cedar-rfcs/blob/entity-literal-syntax/text/0104-entity-data-syntax.md)


text/0000-entity-literal-syntax.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Drawbacks
+
+There are several reasons not to develop a custom syntax:
+
+### Multiple formats can raise cognitive burden
+
+Adding another custom format raises the bar for what customers need to know. They may ignore one format at first, but eventually they may need to learn both. For example, if one team uses the JSON format and another uses the custom format, but then the teams merge, they will each end up having to read/update the other's format. They may also fight about which format to move to.
+
+Mitigating this problem is that it's easy and predictable to convert between the two formats, since the syntactic concepts line up very closely. The features you lose when converting from the new format to the JSON one would be 1/ any comments, and 2/ any use of intermingling of `action`, `entity`, and `type` declarations, since they must all be in their own sections in the JSON. Otherwise the features line up very closely and the only difference is syntax. We would expect to write conversion tools as part of implementing the new syntax (which are easily achieved by parsing in the new format and pretty-printing in JSON).
```
cdisselkoen: I agree that easy conversion between the formats is a sufficient mitigation for this. And that everything said in the Motivation section outweighs the drawback of increased cognitive burden described here.

cdisselkoen (commented):
I like this proposal, and I definitely think we should accept some version of this proposal.

text/0000-entity-literal-syntax.md
```diff
@@ -0,0 +1,224 @@
...
+* Consistency with policy and especially schema syntax
+* Easy for parsers to differentiate from policy and schema
+* Optionally, allow for reusable record (and maybe primitive) literals
+* Optionally, if use cases exist for it, reusability of the syntax for inline literals in schemas and policies
+
+Two potential options for syntax are listed below.
+
+### Option 1: instance declarations
+
+The schema syntax already defines a syntax for entity types where the values of parents and attributes are types, and actions are already sort of entity literals; replacing these with values provides the basis for this option. We would select a new keyword, using `instance` as an example (`entity` would be nice, but it's taken). This should make it easy to reuse the parsing for schema syntax (and make syntax highlighting straightforward), and the parser could identify when they were incorrectly used in schemas and vice versa.
```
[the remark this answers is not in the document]
mwhicks1: I prefer Option 1, for the reasons given: It looks most natural to me, given the existing Cedar schema and policy syntax.

As for the keyword `instance`: I like leaving open the possibility that entity data could be in the same syntactic input stream as schema data, and a separate keyword allows both to be safely parsed. In a sense, this mixture is already happening: Enumerated entity types appear in schemas, as do actions, and these define particular entity instances implicitly.

But it might be easier not to allow this for now, since there are ramifications about the combination. For example, you can now write
```
entity Color enum ["Red"];
```
and with mixing you could instead write
```
instance Color::"Red" = {};
```
The former indicates that there are no other `Color` entities than `Color::"red"` whereas the latter doesn't say that. You get none of the validity checking of the latter than when you do of the former, e.g., if you have a typo in your policy that refers to `resource == Color::"RED"` the validator flags the mistake.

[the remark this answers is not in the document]
benkehoe: Relevant to combining schema (which can be defined only once) and entity data (there can be many instances), would we want a way to define multiple instances for a type without repeating the `entity <type> instance` prefix? e.g.
```
entity SomeType instances [
  "id1" ["parent"] {},
  "id2" {}
];
```

mwhicks1 (commented):
This is looking good! I really like the direction. Before finally approving, I'd love to see a revision of the proposal that:
1. Proposes the change (Option 1?) in the main body, and puts alternatives in an appendix, for posterity.
2. Has a *Detailed Design* section that fills in the details, e.g., around the particular context-free grammar we want to use, or at least more examples that unambiguously flesh out the details. [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) is probably more than you need, but it's a good guide.

text/0000-entity-literal-syntax.md
```diff
@@ -0,0 +1,224 @@
...
+  Namespace::SomeType::"entity_id"([Namespace::OtherType::"parent_entity_id"], {
+      attribute: "value"
+    }, {
+      tagName: "value"
+    }),
+  //subsequent entities
+]
+```
+
+This would be more explicit that the content of the file is a set of entities and only a set of entities (like with JSON). It's a little less familiar/similar to schema syntax than Option 1, and might be harder to do useful syntax highlighting, but it is relatively close to how extension types like `datetime` are written. It opens up the possibility of composition within policies and schemas, but that is only relevant if that would ever be needed. This syntax doesn't allow for reusable records.
```
philhassey: It'd be nice to see examples of how datetime and all the other extensions would be represented.  There's also some questions around how to represent datetimes that are before 0000-00-00.  (Though I think we might want to not allow those to be represented here.)

text/0000-entity-literal-syntax.md
```diff
@@ -0,0 +1,224 @@
...
+        },
+        "attrs": {},
+        "parents": []
+    }
+]
+```
+
+a more compact syntax could be
+
+```
```
Swolebrain: Can you walk me through a real world use case for this?

One use case I can imagine based on what I've seen some Cedar users do is to export the top n levels of an org hierarchy, since those rarely change. The top n levels of the org hierarchy is something that customers commonly cache in their application layer.

At build time, you can pull this information from your LDAP server, for instance, and then commit it to the repo in cedar entity format. This would make it more human-legible than having it in JSON (but you're not likely to be able to support comments here since they'd get wiped by the import process).

But if you did this, you probably would want to convert the entities back to json (or protobuf when it's available) to use in authz decisions because of performance. 

Given that, why wouldn't you store the entities in regular JSON (as they are in the DB, not cedarJSON) or YAML  instead of converting them to cedarJson or to cedar entity instance format? Wouldn't that achieve readability too?

Whats the use case you're imagining?

benkehoe:
> This is looking good! I really like the direction. Before finally approving, I'd love to see a revision of the proposal that:
> 
> 1. Proposes the change (Option 1?) in the main body, and puts alternatives in an appendix, for posterity.
> 2. Has a _Detailed Design_ section that fills in the details, e.g., around the particular context-free grammar we want to use, or at least more examples that unambiguously flesh out the details. [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md) is probably more than you need, but it's a good guide.

@mwhicks1 will do

text/0000-entity-literal-syntax.md
patjakdev: I guess the broader point is that, so long as `Entity::new()` continues to allows pass extension _method_ calls to be passed as attribute values (as opposed to just extension _function_ calls like `datetime(...)`), it'll be possible to construct entities that can't be serialized in this format.

patjakdev:
I'm strongly in favor of this RFC and would love for it to be accepted and implemented.

One request which might come into play when choosing between the implementation options: I would actually love to be able to have a single, human-readable document (i.e. not JSON) that describes all my entities AND policies. It would also be great to get template instantiations in there, but I won't hold my breath on that one.

At first glance, I can't see anything in any of the options that would prevent such a thing but I wanted to put my two cents in in case it changes the desirability of any of the options.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+
+Path := IDENT {'::' IDENT}
+Name := STR
+EntityRefOrRefs := EntityRef | '[' [EntityRefOrRefs] ']'
+EntityRef := Path '::' STR
+
+Record := '{' [KeyValues] '}'
+KeyValues := Key ':' Value [',' | ',' KeyValues]
+Key := IDENT | STR
+Value := // TODO: scoped down from policy grammar, also need some bits from schema grammar
```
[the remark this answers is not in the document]
cdisselkoen: IMO we can accept the RFC without nailing down yet some of these details, and then during the implementation phase make the final minor calls (eg what identifiers to reserve). Or the RFC can just propose something (e.g., use the same list as the schema grammar) and we can revise after accepting if necessary.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+
+Tags := Record
```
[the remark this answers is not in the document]
cdisselkoen: I don't think this is the job of the grammar, personally.

cdisselkoen (approved):
I think this RFC is in a good place, and I'm happy to approve the current version.

text/0104-entity-data-syntax.md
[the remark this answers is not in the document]
benkehoe: That makes sense, I think everyone knows what value literals should look like.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+        "parents": []
+    }
+]
+```
+
+with this proposal, it could be represented compactly as follows:
+
+```
+namespace PhotoApp {
+  entity User instance "alice" in [UserGroup::"alice_friends", UserGroup::"AVTeam"] {
```
philhassey: In this example `UserGroup::"alice_friends"` should resolve to `PhotoApp::UserGroup::"alice_friends"` (as best I can tell.). I'm not so sure we want to require resolution as part of parsing these files.  

Either we should require resolved namespaces for the entity types, or there needs to be a way to indicate (somehow) that this is referring to an entity type within this namespace.  Since this is just a data format, it's entirely possible that no `UserGroup` is instantiated within the namespace or at the global level.  So there's really no way to know "guess" what the data intends.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+          name: "alice"
+      }
+  };
+
+  entity Photo instance "vacationPhoto.jpg" {
+      private: false,
+      account: Account::"ahmad"
+  };
+
+  entity UserGroup instances [
```
philhassey: I'm not sure that having `instances` is really bringing much value.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+namespace PhotoApp {
+  entity User instance "alice" in [UserGroup::"alice_friends", UserGroup::"AVTeam"] {
+      userId: "897345789237492878",
+      personInformation: {
+          age: 25,
+          name: "alice"
+      }
+  };
+
+  entity Photo instance "vacationPhoto.jpg" {
```
philhassey: The format could be simplified down to this if we wanted something even more terse.

```
Photo::"vacationPhoto.jpg" in [SomeThing::"whatever"] {
   private: false,
   account: Account::"ahmad",
}
```

text/0104-entity-data-syntax.md
benkehoe: Namespace resolution for entity types is already part of the schema syntax definition, and it should work identically here. If we need to improve that, it should still be the same across both.

text/0104-entity-data-syntax.md
benkehoe: It's a relatively small thing but the verbosity of, for example, 50 instances with no attributes would be significantly lower with `instances` (and the suggestion of declaring instances as part of a type declaration wouldn't work without it). I'm not strongly attached to it, but I think it would be convenient (and human-readable convenience is the explicit purpose of this syntax, JSON is expected for machine-constructed entity data).

text/0104-entity-data-syntax.md
benkehoe: Terser formats were in the original options. A challenge that kind of syntax brings is that it's not necessarily easy for a parser for the other Cedar syntaxes to distinguish. You want the syntax to be unambiguous for two reasons: First, clear error messages when inevitably someone puts a file into the wrong API. Second, it allows (as discussed above) the potential for mixing the content in a single file, schema + policies + entities for, say, a dump of the content of an auth database. The current syntax even allows for instances to be defined as part of the entity type declaration.

text/0104-entity-data-syntax.md
philhassey: Since this data format isn't coupled with a schema, there's no way to know for sure if `UserGroup` exists within the PhotoApp or global namespace.

Also, since Cedar doesn't require that everyone do things in a "schema-validatable" way, it is possible for people to do things with data that is not "valid" -- so a `UserGroup` could exist in both.

text/0104-entity-data-syntax.md
philhassey: > Second, it allows (as discussed above) the potential for mixing the content in a single file, schema + policies + entities for, say, a dump of the content of an auth database.

That does sound compelling.

text/0104-entity-data-syntax.md
benkehoe: I'm not saying the resolution of the ambiguity is through schema, I'm saying that schema _also_ has that ambiguity. With this schema:
```
namespace PhotoApp {
  entity User in UserGroup;
}
```
it's also ambiguous whether `UserGroup` is in `PhotoApp` or the global namespace. The way we do resolution for this ambiguity in schemas would be reused to resolve ambiguity in entity instances.

Similarly whatever changes come with #105 (if approved) would apply equally to the entity data syntax.

text/0104-entity-data-syntax.md
benkehoe: It's why I want us to consider the syntax for this and #98 jointly, to make sure it makes sense together.

text/0104-entity-data-syntax.md
mwhicks1: I have concerns about `instances` too. I like that up until now in our non-JSON schema syntax the list form `[...]` just has entity types or entity IDs in the ... part. Note that set literals are also written with `[ ... ]`, e.g., you have
```
entity List instance "1" in Application::"TinyTodo" {
  editors: Team::"1_editors", 
  ...
  tasks: [ { 1, "write abstract", "done" }, { 2, "write outline", "todo" }, ... ]
}
```
In sum: a `[ ... ]` can contain a list of _types_ or a list of _values_. This proposal puts _definitions_ inside a list, which feels incongruous to me, and potentially confusing (for human readers and for parser code, potentially). A more congruous analogy would be our  use of `namespace { ... }` to prepend a bunch of `;`-separated _definitions_ with the namespace. Maybe an alternative for `instances` using `{ ... }` with `;`-separate instance definitions would work better. But then that might create confusion with the single-instance case, `entity Foo instance "bar" { ... }` as opposed to `entity Foo instances { "bar" = { ... }, "baz" = { ... }}`.

Ultimately I agree with Phil that this not providing much value over its cost. I see the point that when you write `entity UserGroup instances [ A, B, C, ... ]` you are saving writing `entity UserGroup instance` for each of B, C, ... separately. As you say, if A, B, and C, ... don't have attributes this is a lot of avoided keywords. But I am wondering how often you'd really want to define entities at all with no attributes, rather than simply not put them in your file.

My proposal would be to drop the `instances` feature at least for now, and consider adding it later.

text/0104-entity-data-syntax.md
mwhicks1: I propose an alternative syntax that is both more terse and should still work (I think/hope) with our other syntaxes:
```
entity Photo::”vacationPhoto.jpg” in [Album::”vacations”,Album::"1998"] = {
  private: false,
  account: Account::"Alice"
};
```
Adding the word “instance” between the entity type and the ID makes it harder to read. I find it much easier to scan over the file looking for `entity Photo::”vacationPhoto.jpg”` compared to `entity Photo instance ”vacationPhoto.jpg”`. 

The enumerated entity syntax should parse fine alongside this syntax (e.g., if entities were in the same file as the schema), e.g.,
```
entity Status enum ["Done", "Open" ];
entity List = {
  name: String,
  description: String,
  status: Status
};
entity List::"example" = {
  name: "submit proposal",
  description: "complete the RFC and submit for review",
  status: Status::"Open"
};
```
Also, I think there is a nice symmetry between the entity ID and entity type syntax in setting it up this way. For example, 
```
entity User in [Application];
entity User::”Alice” in [Application::”TinyTodo”];
```

text/0104-entity-data-syntax.md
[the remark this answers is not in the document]
mwhicks1: If we drop the `instances` option as I've proposed above, and change the entity syntax as I've proposed, the grammar would be
```
Entities := {NamespaceOrEntityDeclaration}
NamespaceOrEntityDeclaration := Namespace | EntityDeclaration
Namespace := 'namespace' Path '{' {EntityDeclaration} '}'
EntityDeclaration := 'entity' EntityRef ['in' EntityRefOrRefs] [['='] Record] ['tags' Tags] ';'
...
```
where the other parts (`EntityRef`, values, etc.) are the same. This is pretty simple setup, I think.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Drawbacks
+
+There are several reasons not to develop a custom syntax:
+
+### Multiple formats can raise cognitive burden
+
+Adding another custom format raises the bar for what customers need to know. They may ignore one format at first, but eventually they may need to learn both. For example, if one team uses the JSON format and another uses the custom format, but then the teams merge, they will each end up having to read/update the other's format. They may also fight about which format to move to.
+
+Mitigating this problem is that it's easy and predictable to convert between the two formats, since the syntactic concepts line up very closely. The features you lose when converting from the new format to the JSON one would be 1/ any comments, and 2/ any use of intermingling of `action`, `entity`, and `type` declarations, since they must all be in their own sections in the JSON. Otherwise the features line up very closely and the only difference is syntax. We would expect to write conversion tools as part of implementing the new syntax (which are easily achieved by parsing in the new format and pretty-printing in JSON).
```
mwhicks1: Another benefit of dropping `instances` is that it simplifies conversion between JSON and natural syntax: There is a more one-to-one connection, since JSON has no "instances" concept.

mwhicks1 (commented):
I am a fan of this proposal, but I would like to suggest two more changes get made:
1. Update the syntax from `entity Foo instance "Bar" ...` to be `entity Foo::"Bar" ...`. I think this stylistically more similar to our other syntaxes and will combine better with them, while also being easier to read/scan.
2. Drop the `instances` option. I don't think the benefits are worth the costs.
If these changes are made I'll be ready to approve!

benkehoe:
@mwhicks1 

> 1. Update the syntax from `entity Foo instance "Bar" ...` to be `entity Foo::"Bar" ...`. I think this stylistically more similar to our other syntaxes and will combine better with them, while also being easier to read/scan.

`entity Foo::"Bar"` for entity instance declarations is ambiguous in its syntax with entity type declarations, so it has a strong potential for reducing the usefulness of error messages. Originally I had `instance Foo::"Bar"` so that entity instance declarations were unambiguous, but people didn't like that.

> 2. Drop the `instances` option. I don't think the benefits are worth the costs.

I'm not sure what the costs are, but overall I'm fine with leaving it out, especially if we stay with the `entity ... instance` syntax that would allow us to add `instances` later if its warranted.

mwhicks1:
> @mwhicks1
> 
> > 1. Update the syntax from `entity Foo instance "Bar" ...` to be `entity Foo::"Bar" ...`. I think this stylistically more similar to our other syntaxes and will combine better with them, while also being easier to read/scan.
> 
> `entity Foo::"Bar"` for entity instance declarations is ambiguous in its syntax with entity type declarations, so it has a strong potential for reducing the usefulness of error messages. Originally I had `instance Foo::"Bar"` so that entity instance declarations were unambiguous, but people didn't like that.

Why is it ambiguous? An entity type will never have the ID part, and an entity instance will never not have it. The grammar would allow

```
DECL ::= 'entity' EntityRef '=' ... | 'entity' Path '=' 
```

> > 2. Drop the `instances` option. I don't think the benefits are worth the costs.
> 
> I'm not sure what the costs are, but overall I'm fine with leaving it out, especially if we stay with the `entity ... instance` syntax that would allow us to add `instances` later if its warranted.

Even if we go with the syntax I've proposed, I think that a future in which there are multiple instances (e.g., using braces rather than brackets per my thread) will be possible.

benkehoe:
It's ambiguous when something goes wrong. If we take

```
entity Foo::bar = {
  key: "value"
}
```
is the error that strings are not allowed there in entity type declarations, or that quotes were forgotten around the entity id in an entity instance declaration?

mwhicks1:
> It's ambiguous when something goes wrong. If we take
> 
> ```
> entity Foo::bar = {
>   key: "value"
> }
> ```
> 
> is the error that strings are not allowed there in entity type declarations, or that quotes were forgotten around the entity id in an entity instance declaration?

That's a fair point. Two further thoughts:

1. The uncertainty only arises when something goes wrong, and you are in a mixed-syntax situation (both entity instances and entity schema in one file together and intermixed). In the typical case when we know you are intending to write just one or the other, the error message can be made clear.
2. If we do allow mixed syntax but go with `instance` there is a similar scenario where someone writes
```
entity Foo::"bar" = { key: "value" };
```
instead of
```
entity Foo instance "bar" { key: "value" };
```
Instead of being told they should have written the above, they'll be told that `Foo::"bar"` is not an entity type. I have a feeling that there are other circumstances in the current grammar where a parse error does not have an unambiguous explanation, but I'm not certain.

My leaning remains toward the more terse syntax, but I won't die on that hill if others prefer `instance` instead.

benkehoe:
> The uncertainty only arises when something goes wrong, and you are in a mixed-syntax situation (both entity instances and entity schema in one file together and intermixed). In the typical case when we know you are intending to write just one or the other, the error message can be made clear.

People have expressed the desire to have intermingled content, for example to be able to take a snapshot of the state of a policy store, and also inevitably people will plug the wrong file into the wrong API and there too you want good error messages. I think it'd be great to have a single grammar for Cedar and use it everywhere, so that error messages could be like "you've got an entity instance declaration in here, but the API is expecting only entity types".

In the example of `entity Foo::"bar" = { key: "value" };` we'd presumably stop parsing after the first or second quote mark. The error message might be "entity instance ids are not allowed in entity type declarations", which I think would be sufficient to set the user back on track, and a "did you mean entity `entity Foo instance "bar"`?" might also be possible if we parse the whole identifier.

mwhicks1:
It seems that our disagreement boils down to different priorities for supporting usability. You are worried that error messages will be hard to follow in some circumstances, whereas I feel that the readability advantage is worth this risk. 

At this point I think we may need others to weigh in.

As a last plug in favor of my alternative, reconsider this nice symmetry between schema (1):
```
entity Photo in [Album,Application] = {
  private: bool,
  account: Account
};
```
and entities that are elements of that type (2):
```
entity Photo::”vacationPhoto.jpg” in [Album::”vacations”,Application::"PhotoFlash"] = {
  private: false,
  account: Account::"Alice"
};
```
The use of the `instance` keyword inhibits following your nose from the schema definition to the value definition because entity IDs appear as you'd expect for every corresponding use of an entity type except after the `entity` keyword, which breaks up said entity ID with `instance` in the middle (3):
```
entity Photo instance ”vacationPhoto.jpg” in [Album::”vacations”,Application::"PhotoFlash"] = {
  private: false,
  account: Account::"Alice"
};
```
Maybe this is a small thing, but I find it easier to quickly understand (2) over (3) esp. after being trained to read (1), and in the case I might be scanning a large file.

benkehoe:
As we ask for additional input, I think there are three options:

1. `entity Foo instance "bar"`
2. `entity Foo::"bar"`
3. `instance Foo::"bar"`

Personally I'm fine with 1 or 3 (including 3 with any keyword other than `entity`).

mwhicks1 (approved):
I’m good with option 3 (using `instance` for each instance, rather than `entity`). In any case, I’ll approve now.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,275 @@
...
+
+```
+instance SomeType1::"entity_id_2" {
+  attribute: "value"
+};
+
+instance SomeType1::"entity_id_3";
+```
+
+Parents are defined like in entity type declarations, but use entity identifiers rather than types. Tags are similar.
```
john-h-kastner-aws, line 173: Even better, this is exactly as in action definitions.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,275 @@
...
+EntityInstanceDeclaration := 'instance' EntityRef ['in' EntityRefOrRefs] [['='] Record] ['tags' Tags]
+
+Path := IDENT {'::' IDENT}
+EntityRefOrRefs := EntityRef | '[' [EntityRefOrRefs] ']'
+EntityRef := Path '::' STR
+
+Record := '{' [KeyValues] '}'
+KeyValues := Key ':' Value [',' | ',' KeyValues]
+Key := IDENT | STR
+Value := // TODO: scoped down from policy grammar, also need some bits from schema grammar
```
john-h-kastner-aws, line 199: It's probably worth being explicit here that this will match the extension function behavior specified for the JSON format in  https://github.com/cedar-policy/rfcs/pull/110

Specifically, extensions method calls like `ip("127.0.0.1").isLoopback()` are not allowed. This is notable because the Cedar Rust implementation actually does allow this in the JSON, so the implementer should be careful to not just inherit the Rust json parsing behavior here.

text/0104-entity-data-syntax.md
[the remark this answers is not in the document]
john-h-kastner-aws: Yeah, entity validation already handles this. If you don't validate entities, then Cedar's actually  fine with heterogeneous tags.

text/0104-entity-data-syntax.md
```diff
@@ -0,0 +1,275 @@
...
+```
+
+## Potential future additions
+
+Not addressed in this proposal but open for future improvement:
+
+* Annotations
+* Intermixing of policies, instances, and/or schema in a single file
+* Deduplication
+  * The ability to define a record once and use it in multiple instances, similar to common types in schemas
```
john-h-kastner-aws, line 214: Perhaps more similar to the macros/functions proposed in #61 (I think @mwhicks1 was saying he might be looking at those again, so worth keeping in mind)

john-h-kastner-aws (approved):
I like it

lianah:
We discussed this RFC during the Cedar maintainers meeting. With three approvals, including from @mwhicks1, who provided substantial feedback, and no new comments since April, the RFC has been accepted. @benkehoe could you fix the failing DCO check so we can merge and begin implementation?

PR #113, opened 2026-04-14
RFC: accept empty principal/resource lists in human schema

olehbozhok:
Aligns the human-readable schema parser with RFC 0055 semantics, which already defines an empty principal/resource list as "action applies to no entities of that kind" and is honored by the JSON parser and the validator. The human parser alone rejects this form; this RFC proposes to remove that rejection. No semantic change.

File will be renamed to the assigned PR number after opening the PR. SUMMARY.md entry will be added in the same PR once the number is known.

[Rendered](https://github.com/olehbozhok/cedar-rfcs/blob/rfc/human-schema-empty-applies-to/text/0113-human-schema-empty-applies-to.md
)

## Checklist

- [x] Added RFC to `text/SUMMARY.md` in numerical order


olehbozhok:
We have a case when we use `partial-eval`, and we need to use it in the schema
```
principal: [NoPrincipal]
```

I think it is a bad solution.

text/0113-human-schema-empty-applies-to.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Drawbacks
+
+- **A mis-author who writes `principal: []` intending "applies to any principal" no longer gets a parse error.** The mistake surfaces at policy validation time instead, with the clear existing message "no principal types are valid for {action}." Linters and IDE tooling may additionally warn on empty lists; that is out of scope for this RFC.
+- **Formatter and test churn.** Round-trip and golden-file tests must be updated. Estimated small: single-digit test files.
+- **Schema consumers that currently rely on "empty list is a parse error" as a validation step will need to emit a warning instead.** No currently-valid schema changes meaning; no currently-valid schema becomes invalid. In that sense this is not a breaking change for schema authors — only for tools that depend on the rejection.
+
+## Alternatives
+
+1. **Leave as-is; keep the sentinel-entity workaround.** Rejected: the sentinel adds no semantic content, directly contradicts the spirit of RFC 0055's "remove dummy entities" direction, and the inconsistency with the JSON parser remains.
```
john-h-kastner-aws, line 117: This is a misunderstanding of RFC55's conclusion. We decided not to introduce a implicit dummy entity, but explicitly defining and using a dummy entity in your schema in an intended use case when the principal or resource is irrelevant. 

> This RFC proposes that users should instead explicitly create these "dummy" entities in an application-specific manner. For this example, the user should create entities of three types: FileSystem, UnauthenticatedUser, and AccountManager. The first will act as the principal for a "createFile" request, and the second and third will act as the principal and resource for a "createAccount" request.

https://github.com/cedar-policy/rfcs/blob/main/text/0055-remove-unspecified.md#basic-example

text/0113-human-schema-empty-applies-to.md
```diff
@@ -0,0 +1,129 @@
...
+# Accept empty `principal` / `resource` lists in the human-readable schema
+
+## Related issues and PRs
+
+- Prior RFCs:
+  - [RFC 0055 — Explicit Unspecified Entities](./0055-remove-unspecified.md). Accepted 2024-05-28, landed in `cedar-policy` v4.0.0. Defines the current semantics this RFC builds on: omitting an `appliesTo` field is a parse error, and **an empty list means the action applies to no entities of that kind**.
+- Reference Issues:
+  - cedar-policy/cedar#1335 — "improve error message for explicitly empty appliesTo" (closed via #1552). Improved the error wording for the human parser's rejection but did not revisit whether the rejection was correct.
+  - cedar-policy/cedar#351 — "Changing semantics of unspecified appliesTo.principalTypes/resourceTypes" (closed, `requires-RFC`). Predates RFC 0055; its concerns were largely resolved by 0055.
+- Downstream motivation: Cedarling multi-issuer / partial-authorization flows that call `Authorizer::is_authorized_partial` for actions whose principal type is intentionally absent.
```
john-h-kastner-aws, line 12: According to Cedar's semantics, it's impossible to validate any request for an action which applies to either no principal or no resource types.

Each action in a schema expands to a set of request environments, each a tuple `(PrincipalType, Action, ResourceType, Context)` computed from the product of the principal and resource types lists. If either list is empty, then the product is empty. To be valid, a request must validate in at least one environment, so a request for an action without any environments cannot be valid.

Even if using partial evaluation, the request must conform to a request environment, with some allowances for missing request data. That request environment specifies the type of the principal.

Can you explain more about how you're calling this?

text/0113-human-schema-empty-applies-to.md
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+}
+```
+
+Under this RFC, both forms parse and produce the same internal schema AST. Omitting the `principal:` clause remains a parse error, per RFC 0055.
+
+## Motivation
+
+**1. The human parser is inconsistent with RFC 0055.** RFC 0055 resolved the semantics: empty list means "applies to no entities of that kind." The JSON parser honors this. The validator honors this (`coreschema.rs` produces a clear "no principal types are valid for {action}" message for policies written against such actions). The human parser alone still rejects the syntax. This is an implementation gap relative to an accepted RFC, not a new feature.
```
john-h-kastner-aws, line 58: This attributes more to RFC55 than is accurate. RFC55 establishes the semantics for missing fields in appliesTo

> Request constructor will no longer accept optional arguments and schemas will no longer support missing appliesTo fields. The Request constructor API will be changed to require an EntityUid (instead of Option) for the principal/action/resource, requiring code changes in programs using Cedar's Rust APIs. The schema APIs will not change, but omitting an appliesTo field will result in a parse error.

It doesn't deal with the semantics of an empty list in appliesTo. Before RFC55 the JSON and human-readable syntax differed in how they handled one-of-two lists being empty. From RFC24, both lists must be non-empty if present. It initially allowed omitting one of the lists to for an unspecified entity type. With RFC55, this changed to forbid omitting either list, with no change to the behavior on empty lists.

text/0113-human-schema-empty-applies-to.md
```diff
@@ -0,0 +1,129 @@
...
+Today (rejected by the human parser, accepted by the JSON parser):
+
+```cedarschema
+entity User;
+entity Doc;
+
+action View appliesTo {
+    principal: [],          // parse error today: "'principal' is '[]', which is invalid"
+    resource:  [Doc]
+};
```
john-h-kastner-aws, line 36: According to Cedar's semantics this is equivalent to `action View;`

It is unfortunate that the JSON format accepts and empty list on only one attribute, but the human-readable format should not become more confusing just for the sake of alignment. The human readable syntax is specifically meant to be an improvement over the JSON syntax.

john-h-kastner-aws (commented):
I don't think we'd be interested in accepting this change.

The proposed syntax 

```
action View appliesTo {
    principal: [],
    resource:  [Doc]
};
```

would be semantically equivalent to `action View`, but it feels unlikely that this would be the schema author's intent. This mistake would lead to any policies or request using the action being marked invalid, so it's preferable to detect the error early and stop the author when they're writing their schema.

If I understand the motivating use case, it can be resolved by introducing an place holder entity for this case where the principal or resource is irrelevant. Rather than using `NoPrincipal`, we would recommend using an entity type name with some semantic meaning in your application.

text/0113-human-schema-empty-applies-to.md
nynymike: no principal = All Principals. In english, couldn't I say 
>Anyone can perform action A on resource B given context C conditions.

We're just electing not to use the principal as the basis for authz.

Issue #103, opened 2025-08-13
Path extension type


benkehoe:
I'm looking to gather use cases for an extension type for paths. I see value in it, and @emina suggested seeing about whether other people have a need for it as well before writing up an RFC. So, I guess I'm asking folks to comment on this about their use cases for path matching in policies.

Many filesystem and filesystem-like use cases (AWS S3, URLs, etc.) have identifiers that encode a tree hierarchy, i.e. paths. Policies for these use cases often include matching against certain components of the path but not others.

A path of a fixed length can be modeled directly as a record (as with the colon-separated parts of an AWS ARN) but not for a path of variable length (as with the last component of many ARNs).

A path of variable length can, of course, be modeled with hierarchy, but this can be significant overhead especially when parents are implicit and may not actually exist as entities themselves (e.g., in URLs).

It can also be modeled with tags, splitting the path into tags, for example a path `"foo/bar/baz"` modeled as tags `"path1": "foo"`, `"path2": "bar"`, `"path3": "baz"`. Then matching a path like `foo/*/baz` is:
```
resource.hasTag("path1") && resource.getTag("path1") == "foo"
&&
resource.hasTag("path2") && resource.getTag("path2") like "*" // can be omitted
&&
resource.hasTag("path3") && resource.getTag("path3") == "baz"
&&
! resource.hasTag("path4")
```

This certainly works, but is pretty awkward, and might not be obvious to Cedar users (I know it took me some time to work out). Packaging it up as a `Path` type seems useful. You'd have a `.getComponent(i)` function. It seems like `.getLength()` might not be workable, but boolean comparator functions for length like `.lengthEquals(n)` might work? `.hasComponent(i)` is equivalent to `.lengthGreaterThanOrEqualTo(i+1)` at least, and `! .hasComponent(i)` to `.lengthLessThan(i+1)`. Maybe they could only take constants?

So you'd have something like:
```
resource.path.lengthEquals(3)
&&
resource.path.getComponent(0) == "foo"
&&
resource.path.getComponent(1) like "*" // can be omitted
&&
resource.path.getComponent(2) == "baz"
```

emina:
Here is one proposal for a path abstraction that would be analyzable.  The question is whether it would be useful.

This is a rough sketch, and the syntax / method names are chosen just for illustration purposes. 

First, we would introduce a `Path` extension type.  Conceptually, a path is a sequence of strings.

[...]

PR #115, opened 2026-06-05
Propose a path extension type

apg:
[Rendered](https://github.com/apg/cedar-rfcs/blob/apg/path-extension-type/text/0115-path-extension-type.md)

## Checklist

- [X] Added RFC to `text/SUMMARY.md` in numerical order

### DIFF
--- a/text/0115-path-extension-type.md
+++ b/text/0115-path-extension-type.md
@@ -1,0 +1,236 @@
+# `path` extension type
+
+## Related issues and PRs
+
+- Reference Issues: [#103](https://github.com/cedar-policy/rfcs/issues/103)
+- Implementation PR(s): (leave this empty)
+
+## Timeline
+
+- Started: 2026-06-05
+- Accepted: TBD
+- Stabilized: TBD
+
+## Summary
+
+This RFC proposes a `path` extension type for Cedar. A path represents a variable-length, ordered sequence of strings — for example, the components of a filesystem path, S3 object key, URL path, MQTT topic, or reversed hostname. The `path` type exposes two matching methods, `.matches()` and `.prefixMatches()`, whose arguments must be string literals. This restriction keeps Cedar's SMT-based policy analysis decidable and efficient. String splitting and delimiter handling are intentionally out of scope; path values are constructed from pre-parsed components supplied by the application.
+
+## Basic example
+
+The following policy allows principals to read objects under the `reports/2025/` prefix of an S3-like object store:
+
+```cedar
+permit(
+  principal is User,
+  action == S3Action::"GetObject",
+  resource is S3Object
+) when {
+  resource.key.prefixMatches("reports", "2025")
+};
+```
+
+Here `resource.key` is a `path` value whose components are pre-parsed from the raw key string. If the key is `"reports/2025/annual.pdf"`, the application constructs `path("reports", "2025", "annual.pdf")` and stores it on the entity.
+
+The following policy matches exactly a three-component path with a wildcard in the middle:
+
+```cedar
+permit(
+  principal is User,
+  action == Action::"read",
+  resource is File
+) when {
+  resource.path.matches("data", "*", "config.json")
+};
+```
+
+This matches `["data", "prod", "config.json"]` but not `["data", "prod", "extra", "config.json"]` (four components) or `["data", "prod", "config.yaml"]` (last component doesn't match).
+
+## Motivation
+
+Many authorization use cases involve resources whose identifiers encode a tree hierarchy: filesystem paths, S3 object keys, URL paths, MQTT topic paths, SPIFFE SVIDs, and domain names. Cedar currently has no built-in representation for variable-length sequences of strings, forcing policy authors to choose between awkward workarounds.
+
+**Hierarchy modeling** works when parent entities exist as Cedar entities and Cedar's `in` operator applies. For implicit hierarchies — S3 keys, URL paths, MQTT topics — the parents do not exist as entities, and creating synthetic ones imposes significant overhead on the calling application.
+
+**Tag-based modeling** splits a path into numbered tags (`"key0"`, `"key1"`, etc.). This works but is verbose, non-obvious, and requires policy authors to carefully manage tag numbering and boundary conditions. The approach is difficult to discover without guidance, and the intent is obscured by the bookkeeping.
+
+**String attributes with `like`** (`resource.keyString like "reports/2025/*"`) is fragile when components can contain the delimiter, cannot enforce exact component count, and cannot distinguish "one wildcard component" from "multiple components matching `*`".
+
+The `path` type packages this common pattern into a clean, analyzable abstraction.
+
+### Use cases
+
+The following use cases motivated this RFC.
+
+**Filesystem paths and S3 object keys.** Prefix matching checks whether a resource lives under a given directory; wildcard patterns handle variable segments such as account IDs or environment names.
+
+**URL paths and SPIFFE SVIDs.** Resources identified by URL path can store path segments as a `path` value, enabling prefix-based access control without synthetic parent entities.
+
+**Hostname suffix matching.** Domains are naturally represented as reversed component sequences: `["com", "github"]` for `github.com`. Storing a pre-reversed hostname as a `path` value enables suffix matching via `.prefixMatches()`. Given a reversed hostname attribute, the following patterns map cleanly:
+
+| Pattern | Cedar expression |
+|---|---|
+| `github.com` (exact) | `context.host.matches("com", "github")` |
+| `*.github.com` (one subdomain) | `context.host.matches("com", "github", "*")` |
+| `**.github.com` (one or more subdomains) | `context.host.prefixMatches("com", "github", "*")` |
+| `foo.*.github.com` (wildcard second-level) | `context.host.matches("com", "github", "*", "foo")` |
+
+For patterns like `foo.**.github.com` — at least one subdomain, ending with `foo` — both ends can be constrained by combining a forward and reversed path attribute:
+
+```cedar
+context.host.prefixMatches("com", "github", "*") &&
+context.reversedHost.prefixMatches("foo")
+```
+
+**MQTT topic paths** and **OAuth scopes** (e.g., `"repo:status"`) follow the same pattern as filesystem paths and can be handled with `.matches()` and `.prefixMatches()`.
+
+## Detailed design
+
+### The `path` type
+
+The `path` extension type represents an ordered, variable-length sequence of strings. It supports equality: two paths are equal if and only if they have the same number of components and the same string at every position. The `path` type does not support `<`, `<=`, `>`, or `>=` comparisons.
+
+### Constructor
+
+The `path(string, ...)` function constructs a `path` value from zero or more string-typed arguments:
+
+```cedar
+path("reports", "2025", "annual.pdf")  // 3-component path
+path()                                  // empty path
+path(resource.bucket, context.tenant)  // dynamic components
+```
+
+Arguments may be any string-typed expression — they are not required to be literals. Strict validation requires each argument to have type `string`.
+
+### Matching methods
+
+The `path` type provides two methods for checking path structure. Pattern syntax follows Cedar's `like` operator: `*` matches zero or more characters within a single component; all other characters match literally. Patterns do not match across component boundaries.
+
+#### `.matches(pattern1, ..., patternN)`
+
+Returns `true` if the path has **exactly** `N` components and each matches the corresponding pattern:
+
+```cedar
+path("data", "prod", "config.json").matches("data", "*", "config.json")  // true
+path("data", "prod", "config.json").matches("data", "*")                  // false — length 3 ≠ 2
+path().matches()                                                           // true — empty matches empty
+```
+
+#### `.prefixMatches(pattern1, ..., patternN)`
+
+Returns `true` if the path has **at least** `N` components and the first `N` match the corresponding patterns:
+
+```cedar
+path("reports", "2025", "annual.pdf").prefixMatches("reports", "2025")  // true
+path("reports", "2024", "annual.pdf").prefixMatches("reports", "2025")  // false — second component
+path("reports").prefixMatches("reports", "2025")                        // false — too short
+path("anything", "goes").prefixMatches()                                // true — zero patterns
+```
+
+#### Argument restrictions
+
+**All arguments to `.matches()` and `.prefixMatches()` must be string literals**, enforced by strict validation. This is the same restriction that requires arguments to `datetime()` and `duration()` to be literals, and for the same reason: it is what makes these methods analyzable. See [Analyzability](#analyzability).
+
+### Element access
+
+`.get(n)` returns the component at zero-based index `n` as a `string`. The argument must be a non-negative integer literal; a computed index is a strict validation error. Accessing an out-of-bounds index is an evaluation error.
+
+```cedar
+path("foo", "bar", "baz").get(0)  // "foo"
+```
+
+`.get()` is useful when a specific position carries a fixed semantic meaning, such as a tenant ID in the first component of a URL path.
+
+### Suffix matching
+
+The `path` type does not provide `.suffixMatches()`. Suffix matching requires locating the start position based on the runtime path length — an unbounded computation that cannot be encoded as a decidable SMT formula. See [Drawbacks](#drawbacks).
+
+The most common suffix use case — matching the final component — can be handled by storing it as a separate attribute:
+
+```cedar
+resource.dirPath.prefixMatches("reports", "2025") &&
+resource.fileName like "*.csv"
+```
+
+For general suffix matching, users can store a pre-reversed path as an additional attribute and apply `.prefixMatches()` to it.
+
+### Pre-parsing
+
+Path values are constructed from pre-parsed components via `path()`. Applications are responsible for splitting path strings into components and handling domain-specific ambiguities (e.g., leading slashes, repeated delimiters). This follows Cedar's design philosophy of pushing parsing and normalization to the application layer.
+
+### JSON encoding
+
+In entity and context data:
+
+```json
+{ "__extn": { "fn": "path", "arg": ["reports", "2025", "annual.pdf"] } }
+```
+
+In the policy AST, the `path()` constructor and its methods follow the same convention as other variadic extension functions:
+
+```json
+{ "path": [{ "Value": "reports" }, { "Value": "2025" }, { "Value": "annual.pdf" }] }
+```
+
+### Schema
+
+```json
+{ "type": "Extension", "name": "path" }
+```
+
+### Analyzability
+
+Cedar's SMT-based analysis decides properties like "does policy A permit any request that policy B forbids?" For analysis to be decidable, all operations on a type must be expressible as finite formulas in theories SMT solvers handle.
+
+The key insight is that arguments to `.matches()` and `.prefixMatches()` are always literal patterns known at analysis time. Given `p.matches("reports", "2025", "*")`, the analyzer knows the expected length (3) and the exact pattern at each position, and encodes the call as a fixed conjunction of `like`-style string constraints — finite, quantifier-free, and decidable.
+
+If non-literal arguments were allowed, the analyzer would need to reason about pattern sequences of unknown length, requiring universal quantification over pattern values. This falls outside the decidable SMT theories Cedar targets.
+
+`.get(n)` requires a literal index for the same reason: the analyzer substitutes the access with a direct reference to the `n`-th component rather than encoding general array-indexing semantics.
+
+## Drawbacks
+
+1. **New extension type maintenance burden.** Adding `path` increases Cedar's language surface area and requires updates to the specification, Lean formalization, Rust implementation, and any other language implementations.
+
+2. **Pre-parsing requirement.** Applications that store paths as plain strings today would need to pre-parse them into components. This is a one-time migration cost for existing Cedar deployments.
+
+3. **No suffix matching.** The absence of `.suffixMatches()` means naturally suffix-oriented use cases require data model restructuring. This is a fundamental constraint: suffix matching requires a runtime reversal operation whose loop bound is not statically known, which requires universal quantifiers over path indices and is not decidable in Cedar's SMT theories. The workarounds (separate final-component attribute, or pre-reversed path attribute) are functional but less ergonomic.
+
+4. **Analysis cost scales with pattern length.** Each argument to `.matches()` or `.prefixMatches()` adds SMT constraints. Very long patterns may slow analysis, though path depths in real authorization scenarios are rarely large enough for this to be a practical concern.
+
+5. **Empty components are not restricted.** The `path` type allows empty string components. Applications that do not want them (e.g., from a leading slash) must filter during pre-parsing.
+
+## Alternatives
+
+### Alternative A: Tag-based modeling
+
+Each path component is stored as a numbered entity tag (`"key0"`, `"key1"`, etc.). This is the most naturally available workaround today, but it is verbose, non-obvious, requires manual boundary-condition management, and buries intent under bookkeeping. The `path` type replaces this pattern with a clean abstraction.
+
+### Alternative B: Records for fixed-length paths
+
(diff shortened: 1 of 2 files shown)


text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Timeline
+
+- Started: 2026-06-05
+- Accepted: TBD
+- Stabilized: TBD
+
+## Summary
+
+This RFC proposes a `path` extension type for Cedar. A path represents a variable-length, ordered sequence of strings — for example, the components of a filesystem path, S3 object key, URL path, MQTT topic, or reversed hostname. The `path` type exposes two matching methods, `.matches()` and `.prefixMatches()`, whose arguments must be string literals. This restriction keeps Cedar's SMT-based policy analysis decidable and efficient. String splitting and delimiter handling are intentionally out of scope; path values are constructed from pre-parsed components supplied by the application.
```
patjakdev, line 16: I might propose `prefixLike` instead of `prefixMatches` for symmetry with the `like` operator.

You might also consider a `prefixEquals` for situations where `*` should be treated as a literal asterisk rather than a wildcard without the need to escape it.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+Arguments may be any string-typed expression — they are not required to be literals. Strict validation requires each argument to have type `string`.
+
+### Matching methods
+
+The `path` type provides two methods for checking path structure. Pattern syntax follows Cedar's `like` operator: `*` matches zero or more characters within a single component; all other characters match literally. Patterns do not match across component boundaries.
+
+#### `.matches(pattern1, ..., patternN)`
+
+Returns `true` if the path has **exactly** `N` components and each matches the corresponding pattern:
```
patjakdev, line 110: Presumably, this method errors if a non-string is passed as an argument.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+The `path(string, ...)` function constructs a `path` value from zero or more string-typed arguments:
+
+```cedar
+path("reports", "2025", "annual.pdf")  // 3-component path
+path()                                  // empty path
+path(resource.bucket, context.tenant)  // dynamic components
+```
+
+Arguments may be any string-typed expression — they are not required to be literals. Strict validation requires each argument to have type `string`.
```
patjakdev, line 102: Interesting. This might be the first extension type constructor that doesn't require a string literal for strict validation since it's not possible to pass an invalid value.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+path("anything", "goes").prefixMatches()                                // true — zero patterns
+```
+
+#### Argument restrictions
+
+**All arguments to `.matches()` and `.prefixMatches()` must be string literals**, enforced by strict validation. This is the same restriction that requires arguments to `datetime()` and `duration()` to be literals, and for the same reason: it is what makes these methods analyzable. See [Analyzability](#analyzability).
+
+### Element access
+
+`.get(n)` returns the component at zero-based index `n` as a `string`. The argument must be a non-negative integer literal; a computed index is a strict validation error. Accessing an out-of-bounds index is an evaluation error.
```
patjakdev, line 135: I'm curious about the effects of this on validation soundness. If a policy accesses a path component on a variable like `resource.key.get(23)`, the validator can't know that index 23 exists in that path and so the policy evaluation can error at runtime, as you mention.

This would introduce a new class of runtime error which strict validation cannot prevent. Another exception would have to be added to the list in [this section](https://docs.cedarpolicy.com/policies/validation.html#validation-benefits-of-schema).

Maybe such issues could be avoided in a similar way to optional attribute access by introducing a `.has(n)` method that checks if the index exists. The validator would then confirm that `.has()` is checked before accessing the index via `.get()`.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+In the policy AST, the `path()` constructor and its methods follow the same convention as other variadic extension functions:
+
+```json
+{ "path": [{ "Value": "reports" }, { "Value": "2025" }, { "Value": "annual.pdf" }] }
+```
+
+### Schema
+
+```json
```
patjakdev, line 178: Include the Cedar format as well?

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Alternative B: Records for fixed-length paths
+
+Records with positional fields work for paths whose length is fixed by the schema. They fail for variable-length paths — S3 keys, URL paths, MQTT topics — which is exactly what this RFC targets.
+
+### Alternative C: String attributes with `like`
+
+`resource.keyString like "reports/2025/*"` is fragile when components can contain the delimiter, cannot enforce exact component count, and cannot distinguish "one wildcard component" from "multiple components". It provides no structural abstraction for hierarchical identifiers.
+
+### Alternative D: Expose `.getLength()` and `.get(n)` with variable indices
```
patjakdev: I don't think this should even be mentioned as an alternative.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Alternative A: Tag-based modeling
+
+Each path component is stored as a numbered entity tag (`"key0"`, `"key1"`, etc.). This is the most naturally available workaround today, but it is verbose, non-obvious, requires manual boundary-condition management, and buries intent under bookkeeping. The `path` type replaces this pattern with a clean abstraction.
+
+### Alternative B: Records for fixed-length paths
+
+Records with positional fields work for paths whose length is fixed by the schema. They fail for variable-length paths — S3 keys, URL paths, MQTT topics — which is exactly what this RFC targets.
+
+### Alternative C: String attributes with `like`
```
patjakdev, line 214: Isn't this the same thing as Alternative F? I suppose alternatives A and B are also the same thing as Alternative F.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Alternative F: Status quo (do nothing)
+
+Tags, entity hierarchy, records, and `like`-based string matching cover path-related use cases today. The `path` type is justified by the frequency of the pattern and the non-obviousness of the workarounds: the original issue notes that working out the tag-based approach "took some time," and several independent parties arrived at the same feature request through different use cases.
+
+## Unresolved questions
+
+- **`.get()` inclusion.** Should element access be included here or deferred to a follow-on? The primary value of `path` comes from `.matches()` and `.prefixMatches()`; `.get()` adds convenience but also implementation and specification surface area.
+
+- **Wildcard semantics.** Pattern matching follows Cedar's `like` semantics: `*` matches zero or more characters within a single component and does not cross component boundaries. This means `path("foo", "bar").matches("*")` is `false` (length 2, not 1). The interaction between zero-length matches and the hostname use case (where `*.github.com` should require a non-empty subdomain) should be confirmed.
```
patjakdev, line 226: I don't quite understand what this is getting at.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+- **Empty path behavior.** Should `path()` — a zero-component path — be a valid value? `path().matches()` and `path().prefixMatches()` would both return `true`, which seems correct, but edge cases in entity data deserve consideration.
+
+- **Feature flag.** Should the `path` type be gated behind a feature flag (as `datetime` was initially) or unconditionally available from the start?
```
patjakdev: Not really something that needs to be considered in an RFC for the language.

text/0115-path-extension-type.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Unresolved questions
+
+- **`.get()` inclusion.** Should element access be included here or deferred to a follow-on? The primary value of `path` comes from `.matches()` and `.prefixMatches()`; `.get()` adds convenience but also implementation and specification surface area.
+
+- **Wildcard semantics.** Pattern matching follows Cedar's `like` semantics: `*` matches zero or more characters within a single component and does not cross component boundaries. This means `path("foo", "bar").matches("*")` is `false` (length 2, not 1). The interaction between zero-length matches and the hostname use case (where `*.github.com` should require a non-empty subdomain) should be confirmed.
+
+- **Empty path behavior.** Should `path()` — a zero-component path — be a valid value? `path().matches()` and `path().prefixMatches()` would both return `true`, which seems correct, but edge cases in entity data deserve consideration.
```
patjakdev, line 228: Seems fine to me.

If we were to generalize `path` into `list`, though, it would be a problem because the type would be unknown.

patjakdev (commented):
I wonder about the feasibility of generalizing the proposed `path` type to an ordered `list` type, where `path` becomes just a `list[string]`.

The biggest wart that I think can think of is that `list[string]` would be the only type to allow for the special wildcard syntax when doing matching. A new `any` keyword could potentially be added to allow for wildcards in `list`s of other types.

text/0115-path-extension-type.md
apg: I'm not sure about the Like vs Matches. I think _Matches_ might work better given its relationship to structual pattern matching, though, you're right about the name symmetry--`prefixLike` might be more ergonomic. Open to other feedback here!

text/0115-path-extension-type.md
apg: Basically, `matches("*")` is strict on the number of components. In the example above `path("foo", "bar").matches("*", "bar")` is fine because it has two components. The variant `prefixMatches` supports the variable number of components.

text/0115-path-extension-type.md
apg: True!

text/0115-path-extension-type.md
apg: Yeah. That would make sense! I'll have to update this.

apg:
> I wonder about the feasibility of generalizing the proposed `path` type to an ordered `list` type, where `path` becomes just a `list[string]`.
> 
> The biggest wart that I think can think of is that `list[string]` would be the only type to allow for the special wildcard syntax when doing matching. A new `any` keyword could potentially be added to allow for wildcards in `list`s of other types.

I think this is something @emina would have to look at. I simply based this off of this comment https://github.com/cedar-policy/rfcs/issues/103#issuecomment-3192317344. It seems as though general lists would be analyzable if they worked like the proposal (only integer literals for accessing elements, no length).

text/0115-path-extension-type.md
apg: I think `.has(n)` probably makes sense. I can add something.

chaluli:
Rather than adding a new path type. I think we can enriching the regular language matching mechanism we currently using in string matching (the `like` operator) to support the glob matching this PR is trying to support.

I concretely propose the following syntax:

```
string.matches(glob("glob literal", "delimiter"))      # globs support both * and ** patterns (1 or 0+ components)
string.matches(re("regular language literal"))         # optional but is the most general matching mechanism
string.componentAt(n_index, "delimiter") == (expr)     # resource.domain.componentAt(1, '.') == principal.team
string.countComponents("delimiter") OP (Integer)       # resource.domain.countCompoents('.') <= 4
```

These all remain analyzable because like the `like` operator, they can be directly compiled into string operations that are supported in SMTLib (mostly `str.in_re(s, R)` where R is a regular language literal, which is exactly what the `like` operator already compiles to when analyzed). Note, we mean that the regular language literal is equivalent to an NFA not a regular expression (which may be non-regular due to forward/backward lookups, groups, etc.)

`my_string.matches(re("R"))` becomes `str.in_re(my_string, R)` (here we assume that R is a regular language as described by SMTLib standard).

`my_string.matches(glob("*.github.com", "."))` becomes `str.in_re(my_string, "[^\.]*\.github\.com")` Behaves just like the `like` operator but makes use of delimiter.

The `string.componentAt(n_index, "delimiter")` and `string.countComponents("delimiter")` unfortunately requires the delimiter be a single character (or at least not e.g., `..` or `//` where there are repeating segments). The first makes use of `str.substr/indexof` (formula grows linearly in the value of `n_index`, fine for n_index = 0, 1, 2, 3, 4, etc. but very bad for n_index = 10000).

emina:
> The `string.componentAt(n_index, "delimiter")` and `string.countComponents("delimiter")` unfortunately requires the delimiter be a single character (or at least not e.g., `..` or `//` where there are repeating segments). The first makes use of `str.substr/indexof` (formula grows linearly in the value of `n_index`, fine for n_index = 0, 1, 2, 3, 4, etc. but very bad for n_index = 10000).

I don’t think we should use `st.substr/indexof` or anything that brings the theory of integers into the translation.  We’ve carefully avoided this so far, because Cedar itself doesn’t have integers:  it uses 64-bit vectors. So, unless n_index is a literal, the translation has to go back and forth between the two representations, which leads to poor performance in practice.  Or if it is, then the formula grows linearly with n_index (as the path representation’s formula would). 

Taking a step back, maybe we should re-consider an MVP solution that handles most use cases: `like` matching with a delimiter.  This would introduce a variation of the `like` operator, along the lines of `<expr> like(<delim>) <pattern>`, or whatever syntax we find most palatable.  This translates directly into a `str.in_re` check, where the `<delim>` character is treated as a literal in the pattern.  While not as expressive as either paths or globing, it avoids the theory of integers and/or linear scaling of the formula with index size. It should perform comparably to `like` without delimiters.  And it should be easier to implement / formalize / maintain than the alternatives.

chaluli:
> Taking a step back, maybe we should re-consider an MVP solution that handles most use cases: `like` matching with a delimiter. This would introduce a variation of the `like` operator, along the lines of `<expr> like(<delim>) <pattern>`, or whatever syntax we find most palatable. This translates directly into a `str.in_re` check, where the `<delim>` character is treated as a literal in the pattern. While not as expressive as either paths or globing, it avoids the theory of integers and/or linear scaling of the formula with index size. It should perform comparably to `like` without delimiters. And it should be easier to implement / formalize / maintain than the alternatives.

I agree with this. The `like` matching with a delimiter I think is the biggest improvement to effort ratio without introducing any new operators into the SMT encoding for analyzability. The last two *could* be done but put us into a situation where we need to increase the kinds of SMT operators in our encoding and/or increase the size of the produced formula proportional to the value of a literal (not good).

apg:
I've been thinking about this today, and I think I can get behind it. I'm wondering about syntax though:

1. `like('/')` -- this introduces a new type, character, that we'd have to deal with.
2. `like("/")` -- this doesn't introduce a new type, maybe is only valid if it's length 1? 
3. `like("::")` -- Is this valid? How would `like("::")` match the string `a:::b` given the potential ambiguity? 

2 makes the most sense to me, but it feels rather unergonomic to use it and restrict it to a single character, when it generally isn't restricted in size.

emina:
> I've been thinking about this today, and I think I can get behind it. I'm wondering about syntax though:
> 
> 1. `like('/')` -- this introduces a new type, character, that we'd have to deal with.
> 2. `like("/")` -- this doesn't introduce a new type, maybe is only valid if it's length 1?
> 3. `like("::")` -- Is this valid? How would `like("::")` match the string `a:::b` given the potential ambiguity?
> 
> 2 makes the most sense to me, but it feels rather unergonomic to use it and restrict it to a single character, when it generally isn't restricted in size.

I’d stick with just single character separators to start with, not arbitrary strings.  If we find that we absolutely need string separators later, we can allow them while remaining backward compatible.

Since the separator string is a literal, CST -> AST or the validator can simply reject any literal with more than character.  So we wouldn’t need a new type.

We can explore other syntactic forms too (e.g., `<expr> like (<pattern>, <delim>)`) or make this a function.  My instinct was to reuse `like` because the two operators are very similar.  But we don’t have to if we feel that would cause too much confusion or parsing issues.

apg:
I like the idea of `<expr> like (<pattern>, <delim>)`, but that probably complicates the grammar. It's probably easier to introduce a builtin function `well_liked(<expr>, <pattern>, <delim>)`

chaluli:
I think this is substantially similar to the `like` operator. Perhaps we could do something like:

```
<expr> like(<delimiter>) <pattern>

"github.com" like(".") "*.github.com"                # false
"cedar.github.com" like(".") "*.github.com"          # true
"policies.cedar.github.com" like(".") "*.github.com" # false
```

We also need to nail down the multi-delimited pattern format.

I see you propose `"**.github.com"` to match `"github.com"`, `"cedar.github.com"`, `"policies.cedar.github.com"`, etc.

Do we want `"**"` to mean any number of delimited sections?
Pro: simple pattern, fairly intuitive.
Con: I believe in the current like operator `"*"` and `"**"` match the same sets of strings. In the proposed update the `*` character is playing double duty: the first `*` is match any character (except the delimiter) any number of times and the second occurrence of `*` is repeat the preceding pattern (the first wild card) 0 or more times. I think this is due to inter-mixing the glob meaning of `*` and the regular language meaning of `*` which are very similar but slightly different.

Perhaps not necessary for a V1. But we might want to also consider if we want to allow matching each delimited section against a pattern rather than just wild-card.

For example
```
"<pattern>.<pattern>.<pattern>" like(".") "(<pattern>)<repeat-delimited-section-sentinel-character>"
```

However, that is getting closer to full support for regular language patterns than the simple glob matching Cedar currently supports.

mwhicks1:
> Do we want `"**"` to mean any number of delimited sections?

I would like to confirm: Is the semantics of `like` with a delimiter that `*` components maximally match everything _except_ the delimiter? In other words, `"github.com" like('.') "*.com"` is true but `"foo.github.com" like('.') "*.com"` is false? That seems to be a pretty small change to the normal `like` handling today; is that sufficient for the use-cases the owner of this RFC has in mind?

Assuming yes, I like this idea of using `**`. It is naturally backward compatible with normal `like` because that form does not have a delimiter -- for normal `like` you interpret `**` the same as `*`. One caveat is that if users are not careful and they take an expression `principal.host like "*.github.com"` and change it to `principal.host like('.') "*.github.com"` it will start to return `false` for things that used to be `true` -- they will have to explicitly change the `*` to `**` to be as before. Probably that's a good thing; if you are explicitly updating your policy to add the delimiter, you are looking for more precise behavior and will thus change the pattern too.

apg:
> I would like to confirm: Is the semantics of like with a delimiter that * components maximally match everything except the delimiter? In other words, "github.com" like('.') "*.com" is true but "foo.github.com" like('.') "*.com" is false? That seems to be a pretty small change to the normal like handling today; is that sufficient for the use-cases the owner of this RFC has in mind?

Yeah, let's break this down so we're on the same page:

1. `"github.com" like('.') "*.com"` is true but `"foo.github.com" like('.') "*.com"` is false? -- **YES**. `*` with `"."` as the delimiter defines a segment that each `*` can match.
2. `"foo.bar.github.com like "**.github.com"` is true, because, as you say "**" is exactly like `like "*.github.com"` today. 

`**` having the semantics of "It matches any number of delimited segments" makes a lot of sense to me, and I believe matches the semantics I proposed in this RFC. To put it in the Perl Compatible Regular Expressions perspective:

1. `*` is equivalent to `.*`
2. `**` is equivalent to `.*`
3. `*` with a delim, `D`, is equivalent to `[^D]+D` (e.g. `*.github.com` only matches if there's a, e.g., `xxxx.` in front)
4. `**` with a delim, `D`, is equivalent to `([^D]+D)+` (e.g. `**.github.com` matches if there's _one or more_, e.g., `xxxx.` in front)

mwhicks1:
> `**` having the semantics of "It matches any number of delimited segments" makes a lot of sense to me, and I believe matches the semantics I proposed in this RFC.

Excellent! If you are aligned with this, would you be able to rework the RFC to make this the proposal and move the current proposal to "Alternatives" for posterity? 

I think this is a nice extension. It seems to be useful and it fits the language design well.

chaluli:
> 3. `*` with a delim, `D`, is equivalent to `[^D]+D` (e.g. `*.github.com` only matches if there's a, e.g., `xxxx.` in front)

I belive it's simply `[^D]+`, right?  Otherwise, `"domain.github.com" like(".") "*.github.com"` would be false; and `"domain..github.com" like(".") "*.github.com"` would be true (note the double `.`).


> 4. `**` with a delim, `D`, is equivalent to `([^D]+D)+` (e.g. `**.github.com` matches if there's _one or more_, e.g., `xxxx.` in front)

I believe the intended semantics is `([^D]+D)*[D^]+` for a similar reason to the above.

Otherwise, that looks like a good change.

paxan:
like      "*.github.com" # matches any subdomain 
              under "github.com", but not "github.com"
    like(".") "*.github.com" # matches "foo.github.com",
              but not "yo.foo.github.com", not "github.com"

so `**` is not even needed right?

chaluli:
> ```
> like      "*.github.com" # matches any subdomain 
>           under "github.com", but not "github.com"
> like(".") "*.github.com" # matches "foo.github.com",
>           but not "yo.foo.github.com", not "github.com"
> ```
> 
> so `**` is not even needed right?

In @apg 's suggested semantics. The delimited glob operators match non-empty sequences. Where in the non-delimited glob operators match possibly empty sequences. So,

```
".github.com" like(".") "*.github.com"      # False
".github.com" like(".") "**.github.com"     # False
"abc..github.com" like(".") "**.github.com" # False

".google.com" like "*.google.com"           # True
"abc..github.com" like "*.github.com"       # True
```

sspencer-slack:
FWIW, Slack's Secops use case also prefers the "like" mechanism over the original `path` proposal.  We believe that it is easier for users to reason about and implementation is probably simpler as well.  

We like the double star `**` semantic for  "one or more non-empty sections divided by the delimiter".  so in other words:

```
"test..github.com" like "*.github.com"  # True
"test..github.com" like "**.github.com" # True (but pointless, extra star)
"test..github.com" like(".") "**.github.com" # False (empty group)
"test.b.github.com" like(".") "**.github.com" # True
"test.b.github.com" like(".") "*.github.com" # False (only matches one group)
"test.github.com" like(".") "*.github.com" # True
"testgithub.com" like(".") "*github.com" # True ? (I think this should be True because I think the delimiter means "Match to the first delimiter, next marker, or start/end of string").
"testgithub.com" like(".") "**github.com" # True (same as previous in this case).
```

apg:
> I belive it's simply [^D]+, right? 

Hmm. You're right. It definitely seems like you'd like to write it `*.github.com`, but I wonder if that means we should be doing something like `X like "(*.)github.com"` or `X like "(**.)github.com"`?

That could also provide an avenue for `(/*)/apg` to prefix well formed (Unix style) file system paths.
 
Thinking this through, I think the main benefit of the Path type is that the delimiter is _eliminated_ entirely so these quirks don't exist.

chaluli:
If you propose the `(*.)github.com` syntax, that makes me think perhaps we could support e.g., `(**.)github.com/(**/)` with more than 1 kind of delimiter in a single string (i.e., for matching URL-like strings). Though one could perhaps split the domain from the path and check them independently rather than mixing delimiters within a single pattern.

Note that the `(*.)github.com` pattern syntax is *not* backwards compatible. Because ( and ) are are allowable literals within a pattern. We can't give them the special meaning to group a delimiter sequence.

apg:
> Note that the (*.)github.com pattern syntax is not backwards compatible. Because ( and ) are are allowable literals within a pattern. We can't give them the special meaning to group a delimiter sequence.

Right. Now that I see it, I don't think we can ignore the position of the delimiter. It's too restrictive to always assume it's _after_ especially if a segment can't be empty.

lianah:
@victornicolet played with some encoding options for the extended like operator and was able to generate some seemingly simple SMT queries that timeout with CVC5. We have cut the following ticket to the CVC5 maintainers to see if there are any optimizations on their side: https://github.com/cvc5/cvc5/issues/12815

chaluli:
@lianah @victornicolet I commented on the CVC5 issue. An alternative encoding for the wild-card which avoids using `re.comp` would probably be the better solution.

Specifically, define the segment-wildcard using the union of two ranges rather than language complement. This should keep you in the nice portion of regular languages (those that don't need to be determinized).

```
(define-fun non_slash () RegLan
  (re.union 
    (re.range " " ".") 
    (re.range "0" "~")))

;; Reusable: zero or more non-"/" characters
(define-fun seg_wildcard () RegLan
  (re.* non_slash))
```

victornicolet:
Thank you, yes it does seem like something even more broad for "non slash" also works in our example, e.g.:
```(re.union (re.range "\u{00}" "\u{2e}") (re.range "\u{30}" "\u{ffff}"))```

tm507211:
Ah, yeah the one I suggested was just ascii characters but full unicode shouldn't be a problem either.

PR #116, opened 2026-07-08
Entity validation using schema fragments with external entity types

john-h-kastner-aws:
[Rendered](https://github.com/cedar-policy/rfcs/blob/2ebe6528bf4f116730586f6cccabfa36f6a45aba/text/0116-entity-validation-schema-fragments.md)

## Checklist

- [ ] Added RFC to `text/SUMMARY.md` in numerical order


text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+
+```
+"B": { "external": true }
+```
+
+### Syntax alternatives
+
+Following the same basic design, there are some alternative syntax options to consider.
+
+* We could use a different keyword: `extern`, shorter, also used by C and Rust; `import`, perhaps wrongly gives the impression that we're loading in an entity definition from somewhere; `use`, similar to `import`.
```
victornicolet, line 119: I like the use of `external` or `extern` rather than `use` / `import`, which to me implies that we have access to the definition of the imported entity. `extern` makes it clear this is just a symbol that is declared externally.

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+
+When combining multiple fragments, we will attempt to link external entity type declarations with concrete definitions. Entity types are linked when their names match precisely (accounting for any containing namespace), but may never be linked with a builtin or common type definition even if the names are the same. An entity type may be declared as external any number of times, but it may only have one concrete definition across all schema fragments. If there is a definition, then that entity type becomes concrete in the resulting fragment, and is treated exactly the same as any other entity type. When there is no concrete definition the entity type remains external in the resulting fragment, but we do not report an error.
+
+If every external entity type in a fragment has a corresponding definition, then the fragment is fully concrete. We can construct a complete `Schema` from it and use it exactly as if there were no external entity type declarations. In this situation, external declarations act as a helpful tool for schema authors to explicitly specify the dependencies between different schema fragments, but could be omitted entirely.
```
victornicolet, line 70: nit: it says `fully concrete` but later refers to `complete`, I prefer `complete` to talk about a complete or closed schema with no unresolved external references.

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+* Add syntax for tagging the kind of type reference when it's used. E.g., `other: B::<entity>`. If `B` is defined as an entity, then it resolves to `B`, otherwise `B` is assumed to be external. This could also be useful for resolving ambiguity when a schema defines `B` as a common type and an entity type. I don't like this option because it doesn't explicitly say that `B` is external, just that it's an entity. We also wouldn't be able to detect simple typos like `Usr` instead of `User` before building the complete schema.
```
victornicolet, line 135: I didn't understand that part, is `<entity>` a tag that would indicate what kind of type `B` is?

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+```
+external action Shared::Action::"edit";
+entity User {
+  my_actions: Set<Shared::Action>
+};
+```
+
+To validate a `User` entity we would check that all elements of `my_actions` are one of the declared instances of `Shared::Action`.
+
+We could not validate a request which applies to an external action.
```
victornicolet, line 233: Beyond this limitation, are there other reasons we should not consider an extension to actions?

victornicolet (approved):

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+
+We have found that we want to store and manage different types of Cedar entities separately.
+
+Say we have an application managing documents for users. Documents and users are stored and managed separately, so we also want to validate their corresponding Cedar entities independently. In this model, each schema fragment typically defines a single entity type and its relationships. We can create a schema defining the document entity, but, for each document, there is a user who owns that document, so our document schema must reference the user entity type.
+
+While the current Cedar schema specification allows us to define the `Document` and `User` entities in separate documents (schema fragments), it does not allow us to validate either type of entity until we have composed a complete schema, even when much of the information in the schema is irrelevant. In fact, we may need to retrieve and load a schema fragment when we don't need any information from that fragment just to satisfy the requirement that all entity types referenced in the complete schema are also defined. In a distributed system, loading this information can lead to unnecessary latency when validating entity data.
```
ericox, line 60: This RFC will be extremely useful for a feature we are building in AVP.

text/0116-entity-validation-schema-fragments.md
ericox: I also like `external` rather than `use`/`import` for these same reasons. The keyword `import` would make me think of expansion of the entity definition in the fragment that has the `import` in it.

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+{
+    "uid": { "__entity": { "type": "Document", "id": "file.txt"} },
+    "attrs": { },
+    "parents": [
+        {"__entity": { "type": "Application", "id": "my_app"}}
+    ]
+}
+```
+
+Validating this entity using only the `Document` schema fragment results in a validation error, even though it validates successfully when using a complete schema formed from both fragments. This limitation is not immediately a problem: by being more restrictive, entity validation with a schema fragment remains sound. It also should not be entirely surprising that validation is less capable when given incomplete information. Nonetheless, this is likely to be a pain point for users.
```
ericox: This would be a pain point for a future requirement for information points post-launch (non P0 requirement). Customers have asked for cross-information point references. In this example I could have an information point `A` defined with  the first schema fragment in this example:

```
external entity Folder;
entity Document in Folder;
```

Then I could have an information point `B` defined with fragment:

```
external entity Application;
entity Folder in Application;
```

Writes to `A` would then fail.

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+
+If loading multiple fragments is simply not an option, users can still work around this issue by modifying the schema to make transitive relations between entity types explicit. For example, the relation between `Document` and `Application` was previously implicit based on the transitive nature of the entity hierarchy. We can modify the schema to make this explicit:
+
+```
+external entity Folder, Application;
+entity Document in Folder, Application;
+```
+
+Now that `Application` is declared and specified to be a valid parent of `Document`, the entity `Document::"file.txt"` validates as expected. This new fragment is semantically equivalent to the original fragment once it has been composed to form a complete schema. This workaround however makes customers replicate parts of their entity hierarchy in every fragment that needs it. A change to `Application` or `Folder` to allow for more parent entity types would require a corresponding change to the `Document` schema before being reflected in `Document` entity validation.
```
ericox, line 183: This could be a confusing customer experience as different orgs with separate IAM permissions may manage group memberships separately from a team that owns definitions of leaf nodes in the hierarchy. Replicating parts of the hierarchy would require cross-org/team communication.

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+
+* Validate an instance of an undefined external entity type. To validate an entity, we check that its attributes, tags, and ancestors have their expected types. We don't know what these types should be, so we must return a validation error.
+* Validate any policies. There may be a path to allow validating some restricted set of policies as a future extension, but it does not have a clear motivation, and the soundness of such validation is not obvious. We would also need to consider the downstream effects on type aware partial evaluation and policy analysis before committing to supporting this extension to policy validation.
+
+## Syntax
+
+I propose adding `external` as a new Cedar schema keyword. The keyword appears before the `entity` keyword, avoiding any ambiguity between the keyword and entity type names. In this example, we declare that `B` and `C` are entities, so we can resolve the entity type references in the definition of `A`. Note that the external entity declarations can be combined in the same way as regular entity declarations.
+
+```
+external entity B, C;
```
ericox, line 89: This interface, LGTM.

text/0116-entity-validation-schema-fragments.md
```diff
@@ -0,0 +1,233 @@
...
+Now that `Application` is declared and specified to be a valid parent of `Document`, the entity `Document::"file.txt"` validates as expected. This new fragment is semantically equivalent to the original fragment once it has been composed to form a complete schema. This workaround however makes customers replicate parts of their entity hierarchy in every fragment that needs it. A change to `Application` or `Folder` to allow for more parent entity types would require a corresponding change to the `Document` schema before being reflected in `Document` entity validation.
+
+### External entity parents
+
+Instead of declaring ancestors in-line, we could add syntactic support for (optionally) defining the parent types of external entities.
+
+```
+external entity Application;
+external entity Folder in Application;
+entity Document in Folder;
```
ericox, line 192: This is easier to understand compared to in-line version, IMO.

